PR #17, opened 2018-11-13
Move travis configuration to repository root

alvarocarvajald:
- Move travis configuration file to repository root
- Fix configuration file to avoid duplicate lines.


.travis.yml
```diff
@@ -0,0 +1,25 @@
+dist: xenial
+
+notifications:
+  email: false
+
+before_script:
+  - set -e
+  - curl -sLo /tmp/terraform.zip https://releases.hashicorp.com/terraform/0.11.10/terraform_0.11.10_linux_amd64.zip
```
hsehic: What I don't like here is adding fixed version ... quick check on https://releases.hashicorp.com/terraform/ didn't gave me link to -latest or -stable ... can we somehow make sure we have always the version we are working with?

alvarocarvajald: I guess we could try to get the latest "stable" version number from a previous curl (stable meaning that it does not contain "-alpha" or "-beta", but it could be dangerous). Let me see what I can think of.

alvarocarvajald: Figured a way to get the latest version of terraform from `www.terraform.io/downloads.html`. Changes pushed to the branch.

PR #26, opened 2019-01-07
Add libvirt provider

arbulu89:
New deployment using libvirt.

The deployment consists on using terraform to deploy the infrastructure and salt to provision the machines and manage SAP HANA software.

One thing to consider, would be to look if we could extrapolate the work here using salt to the other providers, to have the provisioning part working the same way (at the end, it should be almost the same for all environments). Let me know what do you think about that!


ldevulder:
Looks very good! I'm not so skilled in Terraform to be sure that all is running well in this code.

@hsehic, can you have a look too before we can trying to merge?

libvirt/terraform/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+  default = "192.168.106.0/24"
+}
+
+variable "name_prefix" {
+  description = "Prefix of the deployment VM, network and disks"
+  default = "hanatest"
+}
+
+variable "sap_inst_media" {
+  description = "url to the media where SAP software is stored. This media shall be mounted in /root/sap_inst"
```
diegoakechi:
```suggestion
  description = "URL of the NFS share where the SAP software installer is stored. This media shall be mounted in /root/sap_inst"
```

arbulu89: Thank you for your comments @diegoakechi . I mainly agree. I'll make the changes.

ldevulder:
@alvarocarvajald, when you have time can you take a look at this? I will then merge it.

hsehic:
> Waiting for @hsehic review, but LGTM.
@ldevulder Have you tested it?

ldevulder:
> @ldevulder Have you tested it?

No, I didn't have time, I sleep during night! xD

But as @diegoakechi tested it, and as only new files/things are added I think that we have little risk to merge it :)

salt/hana_node/files/pillar/cluster.sls
```diff
@@ -1,4 +1,18 @@
...
+  watchdog: /dev/watchdog
+  sbd:
+    device: '/dev/vdc'
+  configure:
+    method: 'update'
+    url: '/tmp/cluster.config'
+  # This next options only are used to update the cluster.j2 template file.
+  # admin_ip_x is used to avoid the creation of the RA in "cluster init"
+  # and create it when the file is imported (crm configure load)
+  admin_ip_x: '192.168.106.50'
```
diegoakechi: @arbulu89 As discussed, the correct terminology here is "virtual_ip", that is the IP managed by the cluster and migrated between nodes in case of failure.

```suggestion
  virtual_ip: '192.168.106.50'
```

salt/hana_node/files/pillar/cluster.sls
```diff
@@ -1,4 +1,18 @@
...
+  admin_ip_mask: '255.255.255.0'
```
diegoakechi:
```suggestion
  virtual_ip_mask: '255.255.255.0'
```

salt/hana_node/files/pillar/cluster.sls
arbulu89: Agree! Changes done in the last commit together with the NTP server usage addition

diegoakechi:
@ldevulder @hsehic, I have tested four times already, including take-over and it has worked very well, so I think we are good to merge it. It also has low risk, as it is not integrated with the other deployments yet, so, will not break any of the existing deployments.

pgonin:
looks good to me !
I see some SUSE internal info though (not critical but still I see internal IPs, internal URLs, repos) which should not appear on a public repo and with code meant to be shared for public use

juadk:
@arbulu89 Nice automatisation!
I had a very small issue with the NFS mount for SAP source. I think it will be better if the nfs version was not forced to 4.
In my case, the NFS server only accept version 3.
I've just replaced nfsv4 by nfs in this salt state : salt/hana_node/sap_inst.sls

salt/hana_node/sap_inst.sls
```diff
@@ -0,0 +1,18 @@
...
+sap_inst_directory:
+  file.directory:
+    - name: /root/sap_inst
+    - user: root
+    - mode: 755
+    - makedirs: True
+  mount.mounted:
+    - name: /root/sap_inst
+    - device: {{grains['sap_inst_media']}}
+    - fstype: nfs4
```
juadk:
```suggestion
    - fstype: nfs
```

arbulu89: Good point! I will check and update that. Thanks!

diegoakechi:
Now, is it good to merge?

salt/hana_node/hana_packages.sls
```diff
@@ -0,0 +1,29 @@
+python-shaptools:
```
ricardobranco777: This package exists in your `openSUSE_Leap15_15.0` repository as `python3-shaptools`.  I suggest adding packages for SLES 15SP1 for testing ;)

salt/hana_node/files/pillar/cluster.sls
```diff
@@ -0,0 +1,18 @@
...
+    method: 'update'
+    url: '/tmp/cluster.config'
+  # This next options only are used to update the cluster.j2 template file.
+  # virtual_ip is used to avoid the creation of the RA in "cluster init"
+  # and create it when the file is imported (crm configure load)
+  virtual_ip: '192.168.106.50'
+  virtual_ip_mask: '255.255.255.0'
+  platform: 'libvirt'
+  prefer_takeover: 'true'
+  auto_register: 'true'
```
ricardobranco777: I would change `auto_register` from `true` to `false` as it's done in the cloud deployments.

libvirt/terraform/modules/host/main.tf
```diff
@@ -0,0 +1,135 @@
...
+    )}"
+  ]
+
+  xml {
+    xslt = "${file("modules/host/shareable.xsl")}"
+  }
+
+  connection {
+    user = "root"
+    password = "linux"
```
ricardobranco777: I suggest not to use passwords at all when we can use SSH public keys.

salt/hana_node/files/pillar/cluster.sls
arbulu89: Of course we can change this parameter, but remember that this customization is up to the final user. This is not a fixed and immutable configuration.

libvirt/terraform/modules/host/main.tf
arbulu89: Working in a new salt action to upload some custom SSH keys to the machines (configurable via terraform). This way ssh is available always for all the machines in the cluster

salt/hana_node/hana_packages.sls
arbulu89: SLE 15 repository added. After the build process finishes it should be available in:
https://download.opensuse.org/repositories/home:xarbulu:sap-deployment/SLE_15

libvirt/terraform/variables.tf
```diff
@@ -0,0 +1,38 @@
+variable "qemu_uri" {
+	description = "URI to connect with the qemu-service."
+	default = "qemu:///system"
+}
+
+variable "base_image" {
```
hsehic: @arbulu89 @diegoakechi  If I followed it right, we still have download.suse.de here, what should be changed

PR #27, opened 2019-01-25
Add ami iscsi server

juadk:
- Improvement for using a different AMI for iSCSI server.
- README update according to this change
- Change of default AMI to use SLES 15 instead of SLES 12SP3, it's just an update for default options.


ricardobranco777:
> I don't know why Travis is failing, as all LGTM.

`terraform fmt` must be run to fix a missing blank line on `variables.tf`

alvarocarvajald:
> I don't know why Travis is failing, as all LGTM.

It's complaining about `variables.tf` in `aws/terraform`. As @ricardobranco777 says, do a `terraform fmt` in the directory and push the change, and it should be fixed.

PR #30, opened 2019-02-06
Fix and init type

ricardobranco777:
- The first commit fixes a bug that currently prevents the deployment due to changes in upstream scripts from Google.
- The second commit adds the `init_type` variable, used by other PC deployments. Valid values are `all` (installs HANA and configures cluster), `skip-hana` (does not install HANA, but configures cluster) and `skip-cluster` (installs HANA, but does not configure cluster). Defaults to `all`.

### DIFF
--- a/gcp/terraform/README.md
+++ b/gcp/terraform/README.md
@@ -95,11 +95,14 @@ These are the relevant files and what each provides:
 
 - The `sap_hana_deployment_bucket` variable must contain the name of the Google Storage bucket with the HANA installation files.
 
-- The `images_path_bucket` must contain the name of the Google Storage bucket with the SLES image.
+- The `images_path_bucket` variable must contain the name of the Google Storage bucket with the SLES image.
 
-- The `sles4sap_os_image_file` must contain the name of the SLES image.
+- The `sles4sap_os_image_file` variable must contain the name of the SLES image.
 
-- The `post_deployment_script` specifies the URL location of a script to run after the deployment is complete. This script should be hosted on a web server or in a GCS bucket.
+- The `post_deployment_script` variable specifies the URL location of a script to run after the deployment is complete. This script should be hosted on a web server or in a GCS bucket.
+
+- The `init_type` variable controls what is deployed in the cluster nodes. Valid values are `all` (installs HANA and configures cluster), `skip-hana` (does not install HANA, but configures cluster) and `skip-cluster` (installs HANA, but does not configure cluster). Defaults to `all`.
+
 
 2. Deploy:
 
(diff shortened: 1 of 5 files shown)


gcp/terraform/README.md
```diff
@@ -95,11 +95,14 @@ These are the relevant files and what each provides:
...
 
-- The `images_path_bucket` must contain the name of the Google Storage bucket with the SLES image.
+- The `images_path_bucket` variable must contain the name of the Google Storage bucket with the SLES image.
 
-- The `sles4sap_os_image_file` must contain the name of the SLES image.
+- The `sles4sap_os_image_file` variable must contain the name of the SLES image.
+
+- The `post_deployment_script` variable specifies the URL location of a script to run after the deployment is complete. This script should be hosted on a web server or in a GCS bucket.
+
+- The `init-type` variable controls what is deployed in the cluster nodes. Valid values are `all` (installs HANA and configures cluster), `skip-hana` (does not install HANA, but configures cluster) and `skip-cluster` (installs HANA, but does not configure cluster). Defaults to `all`.
```
alvarocarvajald: I understand from this explanation the differences between `skip-hana` and `skip-cluster`, but do not see in the initialization script itself how it is implemented. There is an implementation for `skip-all`, but there does not seem to be a difference between `skip-hana` and `skip-cluster`, both of which would prevent Hana to be deployed, but would also configure the cluster.

ricardobranco777: You're right.  I renamed `skip-all` to `skip-cluster`.

gcp/terraform/startup.sh
```diff
@@ -36,19 +36,26 @@ source /dev/stdin <<< "$(curl -s ${DEPLOY_URL}/lib/sap_lib_ha.sh | sed -r 's/(AU
...
 main::get_os_version
+main::get_settings
 
 if [[ -n ${VM_METADATA[suse_regcode]} ]] ; then
 	SUSEConnect -r "${VM_METADATA[suse_regcode]}"
 	( . /etc/os-release ; SUSEConnect -p sle-module-public-cloud/${VERSION_ID%-*}/x86_64 )
 fi
 
+if [[ ${VM_METADATA[init_type]} == "skip-cluster" ]] ; then
+	exit 0
```
alvarocarvajald: With this change, `skip-cluster` not only prevents the script from setting up the cluster, but also prevents the script from installing Hana. This is a different approach than in AWS or Azure in this repository, and it is also different from what it is documented in the README as part of this PR. I recommend either amending the script or the documentation so they are consistent with each other.

PR #35, opened 2019-02-08
Expand travis check to any new provider

alvarocarvajald:
This commit makes travis check any new provider added to the repo, to guarantee that it is validated and given appropiate tests on the initial push. The travis script was also updated to take into account that
`terraform init` for libvirt will not work in travis as the terraform binary from hashicorp cannot automatically download the libvirt plugin, and that there is no remote-state.tf file in libvirt.

Libvirt's main.tf was also formatted per terraform v0.11.11.


ricardobranco777:
LGTM

ricardobranco777:
Until new versions of packages arrive, for libvirt we have to use openSUSE terraform for deployments and Hashicorp's for `terraform fmt`

alvarocarvajald:
Will merge this since we have 3 LGTMs

PR #36, opened 2019-02-08
Add support for last HANA SPS03 version

ldevulder:
This PR update the HANA install script to support the last version of HANA (SPS03rev033).

This version doesn't need a libssl compat package on newer SUSE OS.

Note: The HANA package should be uploaded on each Public Cloud of course!


ricardobranco777:
LGTM

juadk:
LGTM
It's the changes I had to make during the last BV.

PR #37, opened 2019-02-08
Change 'order' constraint to not use deprecated one

ldevulder:
The HA order constraint for HANA should use the 'Mandatory' keyword instead of the current value of '1000'.

This is mandatory for newer version of Pacemaker, but also usable on previous version available on SLE-12-SP3+.


ricardobranco777:
LGTM

juadk:
LGTM
Tested during the last BV.

PR #39, opened 2019-02-12
Feature/output data

arbulu89:
Add output option to return network data.

Due current terraform limitations, we cannot access only to the network addresses as they are created dynamically and terraform cannot find them at the init. 

Terraform does not perform any lazy loading, so it checks all the elements always, and if they don't exist the execution fails.

This will be fixed in terraform v12.

Here more:
https://bradcod.es/post/terraform-conditional-outputs-in-modules/


libvirt/terraform/modules/host/main.tf
```diff
@@ -130,6 +130,12 @@ EOF
 output "configuration" {
   value {
     id       = "${join(",", libvirt_domain.domain.*.id)}"
-    hostname = "${var.name}${var.count > 1 ? "-1" : ""}.${var.base_configuration["domain"]}"
+    hostname = "${join(",", libvirt_domain.domain.*.name)}"
```
ricardobranco777: I think it's better to have spaces separating these items and not commas. Spaces are copy/paste friendly :)

Otherwise LGTM.

ldevulder: > I think it's better to have spaces separating these items and not commas. Spaces are copy/paste friendly :)

Commas are copy/paste friendly too, IMHO. But space less for me, as we can have tabs instead of space and we don't see them.

arbulu89: Thanks for your comments. We can leave with commas right now. If we find some good reason in the future, we can obviously change!

PR #41, opened 2019-02-12
Improvement for iSCSI server.

juadk:
* A different image could now be use for the iSCSI server with the new
"image_uri_iscsi_srv " variable.
* README update according to this change

Tested with sle12sp4saprc3 as iscsi server image.

### DIFF
--- a/azure/terraform/image.tf
+++ b/azure/terraform/image.tf
@@ -6,6 +6,10 @@ variable "image_uri" {
   type = "string"
 }
 
+variable "image_uri_iscsi_srv" {
+  type = "string"
+}
+
 resource "azurerm_image" "custom" {
   name                = "BVSles4SapImg"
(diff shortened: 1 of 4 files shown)


azure/terraform/image.tf
```diff
@@ -22,3 +26,20 @@ resource "azurerm_image" "custom" {
     workspace = "${terraform.workspace}"
   }
 }
+
+resource "azurerm_image" "custom_iscsi" {
+  name                = "BVSles4SapImgIscsi"
```
diegoakechi: I would suggest to include a count here, in case the user wants to use a pre-existing image, he does not need to remove or delete the image.tf file.
```sugestion
count = "${var.image_uri_iscsi_srv != "" ? 1 : 0}"
```

PR #42, opened 2019-02-13
Add install-salt-minion.sh

diegoakechi:
Install the salt-minion from the SUSE official Repository using
registration. 
This script can be used for provisioning machines or images that do not have a pre-installed salt-minion.

### DIFF
--- a/salt/install-salt-minion.sh
+++ b/salt/install-salt-minion.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+DEREGISTER="false"
+
+function print_help(){
+    cat <<-EOF
+
+Install the salt-minion provided via the official SUSE Repositories.
+
+For this, a Registration Code is required and your system will be 
+connected to the SUSE Customer Center.
+
+Supported Options:
+
+  -r [REG CODE]    Registration Code that will be used to register the system.
+  -d               Deregister the System aften the salt-minion installation
+  -h               Show this help. 
+
+EOF
+}
+
+
+
+function install_salt_minion(){
+    local regcode=$1
+
+    # Check SLE version
+    source /etc/os-release
+
+    SUSEConnect -r "$regcode"
+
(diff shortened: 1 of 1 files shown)


juadk:
LGTM, I already use it, thanks.

salt/install-salt-minion.sh
```diff
@@ -0,0 +1,67 @@
...
+        exit 0
+        ;; 
+    d) 
+        DEREGISTER="true"
+        ;;
+    r)
+        install_salt_minion "$OPTARG"
+        exit 0 
+        ;;
+    \?) echo "Invalid option -$OPTARG" >&2
```
ricardobranco777:
```suggestion
   *) echo "Invalid option -$OPTARG" >&2
```

salt/install-salt-minion.sh
```diff
@@ -0,0 +1,67 @@
...
+
+    # Check SLE version
+    source /etc/os-release
+
+    SUSEConnect -r "$regcode"
+
+    # Register the modules accordingly with the SLE version.
+    if [[ $VERSION == *"12"* ]]; then    
+      SUSEConnect -p sle-module-adv-systems-management/12/x86_64
+    elif [[ $VERSION == *"15"* ]]; then
```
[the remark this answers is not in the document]
diegoakechi: Makes sense @ricardobranco777. Thanks!

salt/install-salt-minion.sh
diegoakechi: @ricardobranco777 I am using `?` as recommended on the getopts man page. Any reason where `*` would be better here?

ricardobranco777: The `getopts` example uses `?)` but `*)` helps to catch unhandled options specified in the option string.  It's very much like using `default` in the C version of `getopt`.

PR #44, opened 2019-02-14
Feature/salt formula improvements

arbulu89:
Libvirt deployment updated to work with the new salt formulas features;

- SSH keys are automatically created during cluster creation
- Package installation is managed in the formulas, so removed from here
- NTP server is managed in cluster formula

Basically many operations are removed from this project as they are executed in the formulas, to make it easier and add extra features to the formulas


arbulu89:
I will take care of the conflicts

ldevulder:
@arbulu89: this PR contains some commit that are already merged... so if I tried to apply it as a patch I have some trouble... Could you remove the already merged commit please?

arbulu89:
> @arbulu89: this PR contains some commit that are already merged... so if I tried to apply it as a patch I have some trouble... Could you remove the already merged commit please?

Just read that! Yes I will do that!

arbulu89:
@ldevulder @diegoakechi I have just tested again (after the new packages) and everything works fine (there is an error at then end of the execution, but this is caused by a bug in crm cluster join command).
I don't see any risk on merging this branch, what feel free to review deeper if you want so!

arbulu89:
@diegoakechi @ldevulder I have done some wrong rebasing work here... I have added even more commits. I will close this PR, and create a new one with the proper changes and commits

PR #49, opened 2019-02-26
Salt rework for AWS

juadk:
New provisioning way for AWS by using SALT instead bash script.

The deployment consists on using terraform to deploy the infrastructure and salt to provision the machines and manage SAP HANA software.

Currently, only the HA cluster deployment is working. The HANA formula doesn't work due to a dependency issue in Factory repo, it will be fixed asap.
A lot of things could be improve so let me know what do you think about that!

Obviously, it's still a work in progress.


juadk:
You are right, moreover we have to keep the traditional solution because if we encounter a bug with salt-minion, It will be hard to do the build validation...
Obviously, it will not be merged until everything is validated.

aws/terraform/instances.tf
```diff
@@ -23,6 +22,53 @@ resource "aws_instance" "iscsisrv" {
...
+   }
+
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mv /tmp/salt /root",
+      "sudo mkdir /etc/salt; sudo mv /tmp/grains /etc/salt/",
+      "sudo sh /root/salt/install-salt-minion.sh -r ${var.regcode}",
+      "sudo sh /root/salt/deployment.sh"
+    ]
+  }
```
alvarocarvajald: Do we want to use the `remote-exec` provisioner, or can this be also be run from `user_data`. The issue I have encountered with `remote-exec` is that the state returned to terraform of the instances/VMs deployment is affected by the return code from the script run (in this case `/root/salt/deployment.sh`), while via `user_data` this runs in the background, so you can get an earlier success deployment return code regarding the infrastructure, while the installation is ongoing. I see scenarios where it would be desirable to have the return code of the provisioning script affecting the state of the terraform deployment, but in cases of failure, sometimes it is not easy to destroy the partially deployed infrastructure with `terraform destroy` so I have opted from leaving the provisioning processes running in the backgroung.

aws/terraform/instances.tf
```diff
@@ -23,6 +22,53 @@ resource "aws_instance" "iscsisrv" {
...
+  5:
+    start: 4097
+    end: 5120
+ EOF
+
+    destination = "/tmp/grains"
+   }
+
+  provisioner "remote-exec" {
+    inline = [
```
alvarocarvajald: I would add a `set -e` as the first line of the inline script.

aws/terraform/instances.tf
```diff
@@ -63,10 +108,39 @@ resource "aws_instance" "clusternodes" {
...
+    destination = "/tmp/grains"
+  }
+
+  provisioner "remote-exec" {
+    inline = [
+      "sudo mv /tmp/salt /root",
+      "sudo mkdir /etc/salt; sudo mv /tmp/grains /etc/salt/",
+      "sudo sh /root/salt/install-salt-minion.sh -d -r ${var.regcode}",
+      "sudo sh /root/salt/deployment.sh",
+      "sudo sh /root/salt/formula.sh"
```
alvarocarvajald: Same comment as above.

aws/terraform/variables.tf
```diff
@@ -20,23 +20,24 @@ variable "sles4sap" {
   }
 }
 
-# Map used for suse-sles-sap-12-sp4-byos-v20181212-hvm-ssd-x86_64
-# SLES4SAP 12SP4 in eu-central-1: ami-027ab92ac76584bfd
+# Map used for suse-sles-sap-15-byos-v20180816-hvm-ssd-x86_64
+# SLES4SAP 15 in eu-central-1: ami-024f50fdc1f2f5603
```
alvarocarvajald: Why the change from 12-SP4 to 15-GM? Shouldn't we leave both options?

salt/hana_node/files/test/pillar/hana.sls
```diff
@@ -0,0 +1,46 @@
+hana:
+  nodes:
+    - host: 'ip-10-0-1-0'
+      sid: 'prd'
+      instance: '"00"'
+      password: 'Qwerty1234'
```
alvarocarvajald: In the script-based deployment this SAP password was stored in the configuration file that is in the installation master parent directory. I think we should get the password in a similar manner to avoid having password in the clear. One idea is to have a `%HANAPWD%` tag instead, and add a line in the provisioning script to replace it with a password provided by the customer.

salt/hana_node/files/test/pillar/hana.sls
```diff
@@ -0,0 +1,46 @@
...
+      install:
+        software_path: '/root/sap_inst/'
+        root_user: 'root'
+        root_password: ''
+        system_user_password: 'Qwerty1234'
+        sapadm_password: 'Qwerty1234'
+        extra_parameters:
+          hostname: 'ip-10-0-1-0'
+      primary:
+        name: NUREMBERG
```
alvarocarvajald: Same comment regarding the password applies to the `NUREMBERG` and `PRAGUE` names. In the script based configuration these names were `MST` (master) and `SLV` (slave) to allow the configuration to be more general. A better solution would be to allow the customer to set up their own names in a configuration file.

aws/terraform/variables.tf
juadk: I change the OS version for the iSCSI server because the salt formula use targetcli. 
So I thought it was a good idea switch to SLES15. Do you prefer to stay with 12SP4 for the iSCSI server?
The functionality for using a different AMI for the iSCSI server stills exist.

salt/hana_node/files/test/pillar/hana.sls
juadk: Yes I understand but I included a test folder for our QA needs, Salt use this file if you set the variable test_mode to true. It was to avoid us generating the files for build validation before having  something more automated.
The customer needs to generate a pillar file according to the example to do his proper configuration. I guess that I need to clarify this point.

aws/terraform/instances.tf
juadk: Good point!, yes the terraform deployment fails if you have a bad return code with Salt.
According to your remark, I will try to add deployment.sh and formula.sh via user_data.

aws/terraform/variables.tf
alvarocarvajald: No, just curious as 15 is older than 12-SP4

salt/hana_node/files/test/pillar/hana.sls
hsehic: full ack

salt/hana_node/files/pillar/hana.sls
```diff
@@ -3,41 +3,42 @@ hana:
...
-        system_user_password: 'Qwerty1234'
-        sapadm_password: 'Qwerty1234'
+        root_password: ''
+        system_user_password: 'ENTER_YOUR_PASSWORD'
+        extra_parameters:
+          hostname: 'hana01'
+       sapadm_password: 'ENTER_YOUR_PASSWORD'
       primary:
-        name: NUREMBERG
+        name: MASTER
```
diegoakechi: I would suggest Primary/Secondary terminology here as it is more common in most of the SAP documentation.

salt/hana_node/QA_templates/config/cluster.j2
```diff
@@ -0,0 +1,72 @@
...
+op_defaults \
+    timeout="600"
+
+#
+# production HANA
+#
+
+primitive rsc_ip_{{ sid }}_HDB{{ instance }} ocf:heartbeat:IPaddr2 \
+    params \
+    ip="10.0.0.250" cidr_netmask="255.255.0.0" \
```
diegoakechi: I think we don't need this file copied into the templates. As soon as the package provides this, it should work with the virtual_ip informed by the cluster pillar file.

salt/hana_node/files/pillar/hana.sls
ricardobranco777: But the primary may fail and the secondary would be promoted to master and the name would be misleading.

Something generic like LOCATION_1 & LOCATION_2 would be better, IMHO.

diegoakechi: @ricardobranco777 Humm, to be more precise, it should be then `PRIMARY_SITE_NAME` and `SECONDARY_SITE_NAME`. I don't think primary and secondary is needed to be avoided here, as this is seen from the POV of the deployment, and HANA being a Primary/Secondary system, is very common to all references to guide customers this way.

ricardobranco777: @diegoakechi Agreed.

salt/hana_node/QA_templates/config/cluster.j2
juadk: @diegoakechi Totally agree with you, I missed this point, thanks.

salt/hana_node/files/pillar/hana.sls
juadk: @diegoakechi @ricardobranco777 I agree for `PRIMARY_SITE_NAME` and `SECONDARY_SITE_NAME`. Pushed

arbulu89:
@juadk @diegoakechi We have finally finished with this PR, so we can remove the WIP tag, right?

Let's try to have a look and test it properly. It will help us to continue working in other providers.

salt/hana_node/files/test/pillar/hana.sls
arbulu89: Hello guys! We have changed the initial approach, and instead of having pillar files stored in place, we have created a pillar_examples folder in root and store there some available examples. Now, the user is forced to copy them to the correct folder to work, so they have to manipulate the files content in order to set the SAP HANA and Cluster as the desired.

hsehic:
> @juadk @diegoakechi We have finally finished with this PR, so we can remove the WIP tag, right?
> 
> Let's try to have a look and test it properly. It will help us to continue working in other providers.

FMPOV we can now remove the WIP tag but testing it would take time as we are expecting BV image later today ... therefore today and tomorrow are probably gone ...

salt/hana_node/files/test/pillar/hana.sls
hsehic: @arbulu89 sounds reasonable

juadk:
> @juadk @diegoakechi We have finally finished with this PR, so we can remove the WIP tag, right?
> 
> Let's try to have a look and test it properly. It will help us to continue working in other providers.

@arbulu89, I'm wondering is the "default" password is needed in /pillar_examples/automatic/hana.sls

arbulu89:
> > @juadk @diegoakechi We have finally finished with this PR, so we can remove the WIP tag, right?
> > Let's try to have a look and test it properly. It will help us to continue working in other providers.
> 
> @arbulu89, I'm wondering is the "default" password is needed in /pillar_examples/automatic/hana.sls

No, I totally forgot that, we can remove it and set "ENTRY YOUR PASSWORD" or similar

ricardobranco777:
Before merging, please tag our current work as `v1.0`. @ldevulder @diegoakechi @alvarocarvajald

pillar_examples/README.md
```diff
@@ -0,0 +1,17 @@
...
+In order to use any of them, copy the content of each provider to:
+salt/hana_node/files/pillar with the same file names.
+
+All the information about how to tune the deployment is available in:
+- https://github.com/SUSE/saphanabootstrap-formula (to manipulate the hana.sls file)
+- https://github.com/SUSE/habootstrap-formula (to manipulate the cluster.sls file)
+
+Finally, if instead of deploying SAP HANA and the cluster together, to only
+deploy one of them update the salt/hana_node/files/salt/top.sls file only using
+the desired componente and removing/commenting the other.
```
alvarocarvajald: Should be "component"

alvarocarvajald:
@juadk Since we are going to have 2 sub-directories within `aws` for deployments with and without salt, I think it would be good to add a README to the `aws` folder indicating the differences between the 2. Perhaps even merging common documentation. What do you think? However, if we plan to phase out the directory with deployment without salt, leave as is. Tagging @hsehic and @diegoakechi to get their thoughts on this.

aws/terraform_salt/README.md
```diff
@@ -0,0 +1,597 @@
...
+    - `<module_name>/<product_version>/<architecture>`
+    - *Example:* Suggested modules for SLES for SAP 15
+
+          sle-module-basesystem/15/x86_64
+          sle-module-desktop-applications/15/x86_64
+          sle-module-server-applications/15/x86_64
+          sle-ha/15/x86_64 (use the same regcode as SLES for SAP)
+          sle-module-sap-applications/15/x86_64
+
+ For more information about registration, check the ["Registering SUSE Linux Enterprise and Managing    Modules/Extensions"](https://www.suse.com/documentation/sles-15/book_sle_deployment/data/              cha_register_sle.html) guide.
```
alvarocarvajald: There seems to be an space after `https://www.suse.com/documentation/sles-15/book_sle_deployment/data/`.

aws/terraform_salt/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+* **additional_repos**: Additional repos to add to the guest machines.
+ * **additional_packages**: Additional packages to add to the guest machines.
+ * **hosts_ips**: Each cluster nodes IP address (sequential order). Mandatory to have a generic `/etc/hosts` file.
+
+ Specific QA variable
+ * **qa_mode**: If set to true, it disables all extra packages that do not come from the image (for example, we use `true`to perform the Build Validation of a new AWS Public Cloud image).
```
alvarocarvajald: Suggestion:

**qa_mode**: If set to true, it disables extra packages not already present in the image. For example, set this value to true if performing the validation of a new AWS Public Cloud image.

aws/terraform_salt/terraform.tfvars
```diff
@@ -0,0 +1,80 @@
...
+
+# SSH Public key to configure access to the remote instances
+public_key = "ADD_YOUR_SSH_PUBLIC_KEY_HERE"
+
+# Private SSH Key location
+private_key_location = "/path/to/your/private/ssh/key"
+
+# Custom AMI for iscsi_srv
+#iscsi_srv = {
+#    "eu-central-1" = "ami-xxxxxxxxxxxxxxxxx"
```
alvarocarvajald: IMO, if due to the iscsi salt formula we require the iSCSI server to be 15-GM or higher, then allowing the modification of this variable in `terraform.tfvars` is confusing.

diegoakechi:
@alvarocarvajald IMHO we should maintain only the Salt version. @juadk only did this intermediate step to maintain retro-compatibility with the old version. If it is working for you and you guys are comfortable with that, I would merge this PR as it is, and remove the duplicated directories and do doctos adjustments in separated and more specifics PRs.

alvarocarvajald:
Tested this today, both to deploy an HA cluster without HANA, as well as a HanaSR.

Works as expected.

I've left some comments in the documentation of things that were either not clear enough, or that are no longer required.

In general, expected modification to the configuration after cloning the repository is similar to the pure shell version of the configuration, which is excellent as no extra complexity is introduced for the end user.

From a QA perspective, having the configuration install `salt-minion` from SCC-injected repositories on the iscsi server and cluster nodes is not desired, so for the time being I vote to keep both versions of the configuration (shell script based and salt based) available.

diegoakechi:
> From a QA perspective, having the configuration install `salt-minion` from SCC-injected repositories on the iscsi server and cluster nodes is not desired, so for the time being I vote to keep both versions of the configuration (shell script based and salt based) available.

Fine for me.

PR #50, opened 2019-02-26
Update libvirt deployment to use las formulas

arbulu89:
- NTP connection managed in formula
- SSH keys created dynamically in formula
- HA factory repository usage added
- Salt pillar information adapted to new formulas

*Info*: This is the same PR than the previous, only cleaning old commits


juadk:
The major part makes sense for me except for the SSH keys:
This works for libvirt provider but not on the public clouds because root login and password authentication for SSH are disabled.
It means we have to keep cluster_ssh_pub and cluster_ssh_key or do a difference between libvirt and PC.

arbulu89:
@juadk I totally understand your point. But as this folder contains the deployment for libvirt, I think it makes sense. We can still use the same approach (share ssh keys before the formulas) for Public clouds, the unique difference would be that we need to remove the **cluster.sshkeys** option for the cluster.sls pillar file. In this case, the SSH part would be avoided and original keys would be used.

What we would need to do is just instead of having this information fixed in cluster.sls, somehow make this configurable.

Anyway, I think that it makes sense to create this changes when we have to merge the PC and libvirt salt folders.

What do you think?

juadk:
@arbulu89 ok for me

libvirt/terraform/modules/hana_node/variables.tf
```diff
@@ -29,6 +29,11 @@ variable "additional_repos" {
   default     = {}
 }
 
+variable "ha_factory_repo" {
```
diegoakechi: I would name this variable to something like "ha_sap_deployment_repo" or "ha_sap_formulas_repo", as the ha_factory is very SUSE specific, and the deployment packages could be from other sources like other repos, distros, PackageHub etc.

arbulu89: Changes pushed. Let's try to merge this PR in order to proceed with the AWS changes too, as we will need to merge both changes as well

PR #51, opened 2019-03-04
Add google scripts

ricardobranco777:
Currently, the deployment depends on Google scripts and resource agents that are downloaded.  This PR adds these scripts to avoid that any upstream changes break our deployment.  We should instead merge those upstream changes when needed.

This PR also:
   - Fixes a bug in the Google scripts for SLES 15.
   - Adds support for 15SP1
   - Uses smaller disks when not installing HANA
   - Parameterizes SAP HANA backup size

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ script:
       fi ;
       /tmp/terraform init ;
       if [[ "$provider" == "gcp" ]]; then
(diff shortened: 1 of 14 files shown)


ricardobranco777:
> LGTM, but is there a way to validate/include a validation that we have compatible versions of the google scripts always in the repo?

The only validation done so far is checking the changeset and testing.

ldevulder:
@ricardobranco777: do you need this PR for the Build Validation?

ricardobranco777:
@ldevulder Yes.

PR #53, opened 2019-03-20
Fix the comparison when the reg email is not informed

diegoakechi:
The grains function returns None and the comparison fails generating a
string "-e None" when the email is not informed.

### DIFF
--- a/salt/default/registration.sls
+++ b/salt/default/registration.sls
@@ -2,14 +2,20 @@
 {% if grains['reg_code'] %}
 register_system:
   cmd.run:
-    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }}  {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] is not None else "" }}
+    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }} {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] else "" }}
+    - retry:
+        attempts: 3
+        interval: 15
 {% endif %}
 
 {% if grains['reg_additional_modules'] %}
 {% for module, mod_reg_code in grains['reg_additional_modules'].items() %}
 {{ module }}_registration:
   cmd.run:
-    - name: /usr/bin/SUSEConnect -p {{ module }}  {{ ("-r " ~ mod_reg_code) if mod_reg_code is not None else "" }}
+    - name: /usr/bin/SUSEConnect -p {{ module }} {{ ("-r " ~ mod_reg_code) if mod_reg_code else "" }}
+    - retry:
+        attempts: 3
+        interval: 15
 {% endfor %}
 {% endif %}
 {% endif %}
--- a/salt/default/registration.sls
+++ b/salt/default/registration.sls
@@ -2,14 +2,14 @@
 {% if grains['reg_code'] %}
 register_system:
   cmd.run:
-    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }}  {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] != "" else "" }}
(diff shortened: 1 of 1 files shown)


salt/default/registration.sls
```diff
@@ -2,14 +2,14 @@
...
-    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }}  {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] != "" else "" }}
+    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }}  {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] != None else "" }}
 {% endif %}
 
 {% if grains['reg_additional_modules'] %}
 {% for module, mod_reg_code in grains['reg_additional_modules'].items() %}
 {{ module }}_registration:
   cmd.run:
-    - name: /usr/bin/SUSEConnect -p {{ module }}  {{ ("-r " ~ mod_reg_code) if mod_reg_code != "" else "" }}
+    - name: /usr/bin/SUSEConnect -p {{ module }}  {{ ("-r " ~ mod_reg_code) if mod_reg_code != None else "" }}
```
ricardobranco777:
```suggestion
    - name: /usr/bin/SUSEConnect -p {{ module }}  {{ ("-r " ~ mod_reg_code) if mod_reg_code is not None else "" }}
```
The Python way to compare to `None`.

diegoakechi: Agreed. I pushed this change.

salt/default/registration.sls
```diff
@@ -2,14 +2,14 @@
 {% if grains['reg_code'] %}
 register_system:
   cmd.run:
-    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }}  {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] != "" else "" }}
+    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }}  {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] is not None else "" }}
```
arbulu89: So far this will work if we don't set the reg_code, but maybe would be more secure to use:
```
if not grains['reg_email']
```
If it's the same as in Python this would cover the None and "" scenarios.
But I guess that as we have the next way to declare the example should be enough
```
#reg_email = "<<your email>>"
```

diegoakechi: @arbulu89 I like the short version, but in this case, shouldn't be:

```suggestion
    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }}  {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] else "" }}
```
as it will return true in case the reg_email exists?

arbulu89: ups, you are right, without the exclamation mark.
Anyway, we have to test that, as I don't know what will happen if the key (reg_email) in the grains dict does not exit.
Have you tested that @diegoakechi ?

diegoakechi: I didn't, as it is always created by the project. When not informed, it creates `reg_email:` so it is empty but still exists in the grains. I can give it a try later, but AFAIK it will raise an exception... lets see...

ldevulder:
Can this PR be merged? It's opened for quite a while now :)

arbulu89:
> Can this PR be merged? It's opened for quite a while now :)

From my side it can be merged!

MalloZup:
@diegoakechi @arbulu89  @alvarocarvajald if you agree on this pr feel free to merge . Just look at it again and decide :rocket:

arbulu89:
Changing `None` by `none` fixes the issue. I will commit the changes before the merge

arbulu89:
Changes done and tested.
Now everything works fine.
I will merge the PR

PR #60, opened 2019-03-25
Feature/public ssh key

arbulu89:
Currently we need to copy the public key string. This is quite annoying.

I have updated how this parameter is used to use the path instead the content (this is already done this way in GCP)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ script:
       fi ;
       /tmp/terraform init ;
       if [[ "$provider" == "gcp" ]]; then
-        /tmp/terraform validate -var-file=terraform.tfvars.example -var sap_hana_sidadm_password="NOT_SECRET" -var sap_hana_system_password="NOT_SECRET" -var gcp_credentials_file=/dev/null -var ssh_key_file=/dev/null -var ssh_pub_key_file=/dev/null ;
+        /tmp/terraform validate -var-file=terraform.tfvars.example -var sap_hana_sidadm_password="NOT_SECRET" -var sap_hana_system_password="NOT_SECRET" -var gcp_credentials_file=/dev/null -var private_key_location=/dev/null -var public_key_location=/dev/null ;
       else
         /tmp/terraform validate -var-file=terraform.tfvars.example -var private_key_location=/dev/null -var public_key_location=/dev/null ;
       fi ;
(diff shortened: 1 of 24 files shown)


juadk:
Good point, tested today with your branch feature/azure_salt and it works.

arbulu89:
> LGTM.
> 
> It would be nice to use the same variables (for SSH at least) across _all_ providers. GCP uses `ssh_key_file` & `ssh_pub_key_file`.

Agree. I didn't touch this because I have not used GCP yet. But if you agree I can change all of them. In fact `libvirt` uses other name too

ricardobranco777:
> > LGTM.
> > It would be nice to use the same variables (for SSH at least) across _all_ providers. GCP uses `ssh_key_file` & `ssh_pub_key_file`.
> 
> Agree. I didn't touch this because I have not used GCP yet. But if you agree I can change all of them. In fact `libvirt` uses other name too

Of course.  Better now than never... :)

arbulu89:
> > > LGTM.
> > > It would be nice to use the same variables (for SSH at least) across _all_ providers. GCP uses `ssh_key_file` & `ssh_pub_key_file`.
> > 
> > 
> > Agree. I didn't touch this because I have not used GCP yet. But if you agree I can change all of them. In fact `libvirt` uses other name too
> 
> Of course. Better now than never... :)

@ricardobranco777 Names updated in all providers. Now all of them are using the same naming (even libvirt)

ricardobranco777:
> > > > LGTM.
> > > > It would be nice to use the same variables (for SSH at least) across _all_ providers. GCP uses `ssh_key_file` & `ssh_pub_key_file`.
> > > 
> > > 
> > > Agree. I didn't touch this because I have not used GCP yet. But if you agree I can change all of them. In fact `libvirt` uses other name too
> > 
> > 
> > Of course. Better now than never... :)
> 
> @ricardobranco777 Names updated in all providers. Now all of them are using the same naming (even libvirt)

Now fix the Travis check :)

juadk:
LGTM, thanks @arbulu89

arbulu89:
@alvarocarvajald @hsehic I don't want to create any pressure, but could you have a look on this PR? This PR is blocking some of my other tasks, and the changes are quite trivial, so it shouldn't take long to check it.

Many thanks in advance!

hsehic:
LGTM

PR #62, opened 2019-04-08
Update repository selection conditions

arbulu89:
Grains information is different in old SLES12 version (sp1, sp2, sp3).

To solve that we need to use osmajorrelease instead of osfinger.

Anyway, this approach must be improved in the future to use the exact repository version.
By now, as the packages are not ready for old versions we can workaround with the current solution.


juadk:
LGTM, I agree with @diegoakechi .

PR #63, opened 2019-04-08
Update aws cli usage adding the region option

arbulu89:
In SLES12SP3 (and older) version the aws-cli version doesn't work if region is not provided.

To fix that we will use the region in all the deployments as it works fine.


juadk:
LGTM

PR #74, opened 2019-05-03
Add option to disable salt execution

arbulu89:
A potential user has requested the option to disable the salt provisioning in order to apply custom options. 
This PR adds the option to disable salt execution.
Besides that, I have structured the code to add easily other provisioning methods (ansible for example).

By now, I have only added the option to Libvirt. If you agree with the changes I will extend the change to the other providers following a similar structure.

Actually, in the future, we could think about unifying all the provisioning for all the customers in a unique module, but this requests more changes.


arbulu89:
One small improvement would be to replace the `enable_salt` option to `provisioner` (or similar) this being a string with available options like `salt`, `ansible`, etc. It will reduce the used variables amount

diegoakechi:
> One small improvement would be to replace the `enable_salt` option to `provisioner` (or similar) this being a string with available options like `salt`, `ansible`, etc. It will reduce the used variables amount.
> 
> What do you think?

I like this more. Fewer variables to manage. It also allows to have custom provision, as the user could set this variable to their own customized provisioning tf file.

arbulu89:
Azure and AWS updated too.
Now playing with `provisioner` and `background` options in the tfvars file we can modify the execution

juadk:
@arbulu89 @diegoakechi, I'll do a review asap but please, don't merged it this week since we have the GMC build validation.

arbulu89:
Return code issues fixed. Now when any salt deployment fails the terraform execution will fail (but only the provisioner resource, so the VMs will be correctly created)

aws/terraform_salt/init-server.tpl
```diff
@@ -1,27 +0,0 @@
...
-# Node servers are deregistered just after
-# If you don't use qa_mode, a complete registration is done by salt later
-
-if grep -q 'role: "iscsi_srv"' /tmp/grains; then
-  sh /root/salt/install-salt-minion.sh -r ${regcode}
-elif [[ ! -e /usr/bin/salt-minion ]]; then
-  sh /root/salt/install-salt-minion.sh -d -r ${regcode}
-fi
- 
-mkdir -p /etc/salt; mv /tmp/grains /etc/salt/
```
juadk: I tested AWS and nothing move /tmp/grains to /etc/salt anymore so the salt deployment fails.

arbulu89:
> AWS, Azure and libvirtd have been modified to support the new variables, but GCP is missing...

GCP updated

aws/terraform_salt/init-server.tpl
arbulu89: Issue fixed in the last commit

ldevulder:
I tested GCP with success and @juadk  also tested on AWS, Azure and GCP.

We didn't check on libvirt, but not first target, so we could modify later if needed.

PR #75, opened 2019-05-06
GCP: Support public images

ricardobranco777:
This PR:

- Adds support for public images in GCP using `use_custom_image` variable.
- Fixes the name for iscsi-data.  It should contain the workspace as prefix.

### DIFF
--- a/gcp/terraform/disks.tf
+++ b/gcp/terraform/disks.tf
@@ -1,5 +1,5 @@
 resource "google_compute_disk" "iscsi_data" {
-  name  = "iscsi-data"
+  name  = "${terraform.workspace}-${var.name}-iscsi-data"
   type  = "pd-standard"
   size  = "10"
   count = "${var.use_gcp_stonith == "true" ? 0 : 1}"
@@ -23,7 +23,8 @@ resource "google_compute_disk" "backup" {
 }
 
 resource "google_compute_image" "sles_bootable_image" {
-  name = "${terraform.workspace}-${var.name}-sles"
+  count = "${var.use_custom_image == "true" ? 1 : 0}"
+  name  = "${terraform.workspace}-${var.name}-sles"
 
   raw_disk {
     source = "${var.storage_url}/${var.images_path_bucket}/${var.sles_os_image_file}"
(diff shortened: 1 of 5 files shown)


gcp/terraform/disks.tf
```diff
@@ -23,15 +23,17 @@ resource "google_compute_disk" "backup" {
 }
 
 resource "google_compute_image" "sles_bootable_image" {
-  name = "${terraform.workspace}-${var.name}-sles"
+  count = "${var.use_custom_image == "true" ? 1 : 0}"
```
arbulu89: This comparison is not really needed.
```
 count = "${var.use_custom_image ? 1 : 0}"
``` 
works

ricardobranco777: The variable is a string, not a boolean.  I don't trust Terraform's booleans as they're really string-based.

alvarocarvajald:
Tested this for the GMC BV, and the changes themselves are working, so will merge unless someone is against it. But there are issues:

- Configuration requests Hana system and sidadm passwords even when `init_type = "skip-hana"` is set.
- SCC registration code is mandatory, which is not all that helpful for BV. Without registration in the cluster nodes, the cluster is not started. This is different to the configurations for AWS and Azure.
- iSCSI server leaves logs in `/root/init-iscsi.log`. Cluster nodes do not. Had to go to `/var/log/messages`  to see what was happening.
- Why are these improvements only on the non-salt configuration? I am starting to worry of having different approaches not only between public cloud providers, but between salt or no salt.

ricardobranco777:
> * Configuration requests Hana system and sidadm passwords even when `init_type = "skip-hana"` is set.

This a defect of Terraform.  Maybe setting a default value when `init_type = "skip-hana"` would fix it.

> * SCC registration code is mandatory, which is not all that helpful for BV. Without registration in the cluster nodes, the cluster is not started. This is different to the configurations for AWS and Azure.

Will check this.

> * iSCSI server leaves logs in `/root/init-iscsi.log`. Cluster nodes do not. Had to go to `/var/log/messages`  to see what was happening.

This is a feature. 

> * Why are these improvements only on the non-salt configuration? I am starting to worry of having different approaches not only between public cloud providers, but between salt or no salt.

We have to create Terraform modules for this.  I would wait until 0.12 is packaged by SUSE because of the syntax changes.

alvarocarvajald:
> > * SCC registration code is mandatory, which is not all that helpful for BV. Without registration in the cluster nodes, the cluster is not started. This is different to the configurations for AWS and Azure.
> 
> We need registration to install the google-cloud-sdk package that is needed for the rest. See bsc#1125903

So, in GCP we need to register when not using salt in order to install `google-cloud-sdk`, and when using salt we also need to register in order to install `salt-minion`. Basically the only reason I was against removing the salt-less configurations for AWS and Azure was because in those registration is not necessary and it's better for build validation.

Issue #43, opened 2019-02-13
Reduce the number of configurations needed to use a non-custom image


diegoakechi:
Currently, many steps and file deletions are needed in order to use a non-custom image. Ideally, it should be done dynamically or based on a small set of centralized configuration.

PR #77, opened 2019-05-07
Azure: Simplify procedure to deploy custom/public images

ricardobranco777:
This PR:
- Adds a `use_custom_image` variable to simplify deployment of custom/public images.  Fixes https://github.com/SUSE/ha-sap-terraform-deployments/issues/43


ricardobranco777:
> Hi @ricardobranco777 ,
> 
> I like the changes. Anyway, I wonder why are you using some hardcoded values there? You could request that info to the user having some default options too.
> 
> This is what we do in the `terraform_salt` option

You're right.  I copied the map variable to the one we're using.  Eventually we have to create Terraform modules for this to avoid duplication of code.

azure/terraform/terraform.tfvars.example
```diff
@@ -30,3 +30,22 @@ az_region = "westeurope"
 
 # Variable for init-nodes.sh script
 init-type = "all"
+
+# Set to true to use custom image
+use_custom_image = "false"
+
+# Public sles4sap image (ignored if use_custom_image is false)
```
alvarocarvajald: Just noticed that `use_custom_image = "false"` means "use a custom image uploaded by yourself instead of a public image". Shouldn't it be set to true to use a custom image (ie, not a public one) and false to use one of the images available in the GCP?

alvarocarvajald: After testing, it seems the configuration itself is OK, i.e., `use_custom_image = "false"` makes the configuration use the public images configured in `terraform.tfvars`, so I guess the only fix needed is in the comment, i.e.: "ignored is use_custom_image is true".

ricardobranco777: Good catch.

alvarocarvajald:
Saw this error in the iSCSI server:

```
+ sudo /bin/bash -c 'tcm_dump --s > /etc/target/tcm_setup.sh'
Traceback (most recent call last):
  File "/usr/sbin/tcm_dump", line 374, in <module>
    main()
  File "/usr/sbin/tcm_dump", line 365, in main
    (options, args) = parser.parse_args()
  File "/usr/lib64/python2.7/optparse.py", line 1400, in parse_args
    stop = self._process_args(largs, rargs, values)
  File "/usr/lib64/python2.7/optparse.py", line 1440, in _process_args
    self._process_long_opt(rargs, values)
  File "/usr/lib64/python2.7/optparse.py", line 1515, in _process_long_opt
    option.process(opt, value, values, self)
  File "/usr/lib64/python2.7/optparse.py", line 789, in process
    self.action, self.dest, opt, value, values, parser)
  File "/usr/lib64/python2.7/optparse.py", line 809, in take_action
    self.callback(self, opt, value, parser, *args, **kwargs)
  File "/usr/sbin/tcm_dump", line 217, in tcm_dump_configfs
    p = os.open(attrib_file, 0)
OSError: [Errno 13] Permission denied: '/sys/kernel/config/target/core/iblock_0/dev_sdc10/attrib/pi_prot_format'
```

Fixed it by changing the public image of the iSCSI server to the previous OS image:

```
iscsi_srv_public = {
  "publisher" = "SUSE"
  "offer"     = "SLES-SAP-BYOS"
  "sku"       = "12-SP3"
  "version"   = "2018.08.17"
}
```

PR #80, opened 2019-05-22
Add pillar example file for cost optimized

Simranpal:
Created a separate example file for cost-optimized scenario in pillar_examples/libvirt directory


Simranpal:
> Hi @Simranpal ,
> 
> We need to add a pillar example for the cluster.sls pillar file too, as it has different values (check the best practices guide).

Hi @arbulu89 this cluster.sls file remained same for both scenarios so far.. but now that we need to add the RAs for QA db and reading corresponding entries from cost-optimized related new cluster pillar data as per other PR.. Do you mean those values will be different?

arbulu89:
> > Hi @Simranpal ,
> > We need to add a pillar example for the cluster.sls pillar file too, as it has different values (check the best practices guide).
> 
> Hi @arbulu89 this cluster.sls file remained same for both scenarios so far.. but now that we need to add the RAs for QA db and reading corresponding entries from cost-optimized related new cluster pillar data as per other PR.. Do you mean those values will be different?

The structure is the same bu the values are different:
PREFER_SITE_TAKEOVER value must be set to false in the cost optimized scenario

PR #84, opened 2019-05-29
Feature/libvirt iscsi

arbulu89:
Add iscsi support to libvirt.
Now, using the `shared_storage_type` variable in `tfvars` we can choose between `iscsi` and `shared-disk`


salt/pillar/iscsi_initiator.sls
```diff
@@ -1,12 +0,0 @@
-iscsi:
```
diegoakechi: Is this file being dropped intentionally?

arbulu89: Yes. The truth is that we were not using it. 
Currently, we don't use the iscsi formula to setup the initiator as we found issues there.
In the future, we should check if it worth to use the formula or not.

libvirt/terraform/terraform.tfvars.example
```diff
@@ -6,6 +6,12 @@ iprange = "192.168.XXX.Y/24"
 host_ips = ["192.168.XXX.Y", "192.168.XXX.Y+1"]
 additional_repos = {}
 
+# Shared storage type information
+shared_storage_type = "iscsi"
+iscsi_srv_ip = "192.168.XXX.Y+2"
+iscsi_image = "url-to-your-sles4sap-image" # sles15 or above
+
+
```
ldevulder: One blank line should be sufficient :-)

salt/default/minimal.sls
```diff
@@ -1,6 +1,5 @@
 include:
   - default.setup
-  - default.ip_workaround
```
ldevulder: Why is it not needed anymore?

arbulu89: It still needed. Only moved to other folder to make it more generic for libvirt, as all of the machines created in this provider has the same issue (ip address not assigned to eth1).

ldevulder:
@arbulu89 I had no issue on my tests, but please next time if we said that we want to test, first wait for our go to merge :-) (we can revert in case of issue, but it's sometimes not so easy).

arbulu89:
> @arbulu89 I had no issue on my tests, but please next time if we said that we want to test, first wait for our go to merge :-) (we can revert in case of issue, but it's sometimes not so easy).

I'm really sorry @ldevulder . I just misunderstood your last sentence...
Tell me if you find something that doesn't work properly. I will restore back the master branch in that case

ldevulder:
@arbulu89 no issue, that fine!

And yes I re-read my sentence and it was not clear enough... My English was bad indeed... sorry!

PR #85, opened 2019-06-04
Fix iscsi-data disk name on GCE

ldevulder:
iSCSI data disk should be prefixed with the workspace name like other resources.


ldevulder:
Tested on GCE!

PR #89, opened 2019-06-13
Fix VM name on all Cloud providers

ldevulder:
Fix a regression that appears during the last GCE refactoring...

EDIT: fixed for all provider

### DIFF
--- a/aws/terraform_salt/salt_provisioner.tf
+++ b/aws/terraform_salt/salt_provisioner.tf
@@ -117,9 +117,9 @@ provider: aws
 region: ${var.aws_region}
 role: hana_node
 scenario_type: ${var.scenario_type}
-name_prefix: ${var.name}
+name_prefix: ${terraform.workspace}-${var.name}
 host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]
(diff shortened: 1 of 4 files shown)


ldevulder:
> Anyway, is this not causing the same issue in Azure and AWS?

We could add the same kind of name to Azure and AWS. It's done at that time. It's needed on GCE because the `hostname` variable in `salt_provisioner.tf` is used by GCE to setup the hostname inside the OS, but it's not used like this in Azure and AWS.

But if you think it's more consistent to have the same for all provider I'm OK for this also!

ldevulder:
@arbulu89 I will do the modifications also for Azure and AWS (and test  it!).

arbulu89:
> @arbulu89 I will do the modifications also for Azure and AWS (and test it!).

Thanks @ldevulder It might solve some potential issues in the future

ldevulder:
Tested with success on AWS, I'm updating the PR.

ldevulder:
This PR should be OK.

ldevulder:
@arbulu89 ok, I will merge this PR and will check what you suggest for another PR.

PR #96, opened 2019-07-04
Use public OS image on GCE

ldevulder:
We currently use the same image provided for SAP nodes for the iSCSI server, which is not really needed.

This commit modify this to use a public OS image, by default the SLES4SAP-15 image.

An improvement could be to use a non-SAP image but we will need to add/use another regcode for this image.

These changes have been successfully tested.

### DIFF
--- a/gcp/terraform_salt/README.md
+++ b/gcp/terraform_salt/README.md
@@ -52,7 +52,15 @@ gsutil mb gs://sles-images
 
 7. Upload the image you want to use with:
 
-`gsutil cp OS-Image-File-for-SLES4SAP-for-GCP.tar.gz gs://sles-images`
+```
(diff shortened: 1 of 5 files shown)


ldevulder:
> Just as a comment,
> Could we somehow give the option to the user to use a public image to create the Hana nodes too?
> Not onfy for ISCSI as this PR is doing.
> Having custom images is OK for QA, but for general purposes the public image would be more suitable

Ok, I will do this.

ldevulder:
> Could we somehow give the option to the user to use a public image to create the Hana nodes too?
> Not onfy for ISCSI as this PR is doing.

Done and tested with SLE-15-SAP BYOS image available in GCE.

PR #98, opened 2019-07-04
Fix missing signed-keys in zypper

ldevulder:
In some cases we may need to force refresh the repos and the keys.
Keys may change during lifetime of this OS/image.

### DIFF
--- a/salt/install-salt-minion.sh
+++ b/salt/install-salt-minion.sh
@@ -12,7 +12,7 @@ connected to the SUSE Customer Center.
 Supported Options:
 
   -r [REG CODE]    Registration Code that will be used to register the system.
-  -d               Deregister the System aften the salt-minion installation
+  -d               Deregister the System after the salt-minion installation
   -h               Show this help. 
 
 EOF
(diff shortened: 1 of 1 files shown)


diegoakechi:
> What wouldn't be better to run this action using salt?
@arbulu89 this change is on the script that we use to install the `minion` that is why we cannot do it using salt.  Right now, it is not very efficient, as it register and deregister the system, and then the salt installation register it again. An improvement could be to make the repos.sls to check if the system is registered and skip the step, so we could stop deregistering on the minion script.

ldevulder:
> > What wouldn't be better to run this action using salt?
> > @arbulu89 this change is on the script that we use to install the `minion` that is why we cannot do it using salt.

Yes, that's the reason why I'm not using salt for this: installation of salt-minion fails without this workaround :-)

> Right now, it is not very efficient, as it register and deregister the system, and then the salt
> installation register it again. An improvement could be to make the repos.sls to check if the system is > registered and skip the step, so we could stop deregistering on the minion script.

Could be yes, in another PR maybe? (I could try to do the change, this could help to improve my salt knowledge :)).

arbulu89:
> > What wouldn't be better to run this action using salt?
> > @arbulu89 this change is on the script that we use to install the `minion` that is why we cannot do it using salt.  Right now, it is not very efficient, as it register and deregister the system, and then the salt installation register it again. An improvement could be to make the repos.sls to check if the system is registered and skip the step, so we could stop deregistering on the minion script.

Ok, I was just asking, because this only affect in the salt minion installation, in nothing else. If this is the purpose, go ahead, i see OK the changes

diegoakechi:
> Could be yes, in another PR maybe? (I could try to do the change, this could help to improve my salt knowledge :)).

Yes, another PR is fine. Great if you could take it. Let us know if you need help.

PR #99, opened 2019-07-04
Refresh zypper repos not only on libvirt

ldevulder:
Zypper refresh in Salt code is only executed on libvirt but it can be useful to do this on all cloud provider.

### DIFF
--- a/salt/default/init.sls
+++ b/salt/default/init.sls
@@ -11,10 +11,4 @@ include:
 {% if grains['provider'] == 'libvirt' %}
   - default.timezone
   - default.auth_keys
-refresh_repos:
-  cmd.run:
-    - name: zypper --non-interactive --gpg-auto-import-keys refresh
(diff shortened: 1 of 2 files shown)


ldevulder:
> @ldevulder Yes, you are right. This should be done in all provider.
> I think it would make sense to move this state to [repos](https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/salt/default/repos.sls)

Ok, will do this.

> Anyway, are this PR and [98](https://github.com/SUSE/ha-sap-terraform-deployments/pull/98) doing the same action?

No, but it's during the check of this issue that I found that this refresh was only for libvirt.

ldevulder:
> I think it would make sense to move this state to [repos](https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/salt/default/repos.sls)

done!

PR #102, opened 2019-07-18
WIP: For our HA envinroment version 0.11 of terrafom is required.

varkoly:
Save time for tester!

### DIFF
--- a/libvirt/terraform/README.md
+++ b/libvirt/terraform/README.md
@@ -10,8 +10,8 @@
 
 # Requirements
 
-1. You need to have Terraform and the the Libvirt provider for Terraform. You may download packages from the
-   [openSUSE Build Service](http://download.opensuse.org/repositories/systemsmanagement:/terraform/) or
+1. You need to have Terraform and the the Libvirt provider for Terraform. You need the version v0.11.NN from this packages. You may download packages from the
+   [openSUSE Build Service](http://download.opensuse.org/repositories/systemsmanagement:/sumaform/) or
    [build from source](https://github.com/dmacvicar/terraform-provider-libvirt)
 
    You will need to have a working libvirt/kvm setup for using the libvirt-provider. (refer to upstream doc of [libvirt provider](https://github.com/dmacvicar/terraform-provider-libvirt))


MalloZup:
@varkoly thx: we have already somehow this:

https://github.com/SUSE/ha-sap-terraform-deployments#terraform-version

libvirt/terraform/README.md
```diff
@@ -10,8 +10,8 @@
 
 # Requirements
 
-1. You need to have Terraform and the the Libvirt provider for Terraform. You may download packages from the
-   [openSUSE Build Service](http://download.opensuse.org/repositories/systemsmanagement:/terraform/) or
+1. You need to have Terraform and the the Libvirt provider for Terraform. You need the version v0.11.NN from this packages. You may download packages from the
+   [openSUSE Build Service](http://download.opensuse.org/repositories/systemsmanagement:/sumaform/) or
```
MalloZup: why `sumaform repo here`? I don't think we want to depend on sub-packages

MalloZup:
@varkoly thx: we have already somehow this:

https://github.com/SUSE/ha-sap-terraform-deployments#terraform-version

libvirt/terraform/README.md
varkoly: Diego told me I can get v0.11.NN from this repo. If you know a better one feel free to modify it.

diegoakechi: @MalloZup on this repo (http://download.opensuse.org/repositories/systemsmanagement:/terraform/) we have only terraform 0.12 that does not work with our project. I think we don't want to change the link here, but maybe complement the text saying to get the package 0.11 from some repo until we migrate the project to terraform 0.12.

MalloZup: I'm not really a fun of depending on custom internal pkgs. If people would need to download an old version, we could say to downolad this from here: https://releases.hashicorp.com/terraform/

The switch to TF12 imho will happen really soonish, so dunno, to me is just a link that we will need to remove..

MalloZup: @diegoakechi  imho even if I'm not really a fun of this, I think it is ok. We will remove it later.

Can we put the link in the main.readme.md? In this way we avoid that we forget nested documentation. I think we should update only the central readme.md with this kind of info regarding terraform and not libvirt specific..  tia

diegoakechi: Yes, I am fine with that. The most important thing is that users get the information easily. I am no against of pointing also the link of the official Hashicorp page.

MalloZup: as you prefer. Imho the hashicorp would be safer imho

MalloZup:
autogenerated with https://github.com/MalloZup/doghub:  pr older then 6days. Please update the PR or close it

MalloZup:
closing this since we will migrate soonish to the tf12

PR #103, opened 2019-07-18
Add monitoring module for libvirt and allow automatic-deployment for monitoring

MalloZup:
THIS PR IS on WIP don't review it at moment

https://trello.com/c/F5iPWHzU/795-creating-a-separate-terraform-module-for-monitoring

# Note for reviewers:

The module will add only the grafana/prometheus host.

I intentionally left over the automatically IP injection method. I think for a 1st Pr iteration is ok

## Reminders for later:

add alerts automatically via salt see this  https://github.com/uyuni-project/sumaform/pull/539/files as example


salt/pre_installation/packages.sls
```diff
@@ -5,12 +5,12 @@ iscsi-formula:
         attempts: 3
         interval: 15
 
+{% if grains['role'] == 'iscsi_srv' %}
```
MalloZup: @arbulu89  I moved this up because otherwise this salt where failing for grafana host.

Imho the move iscs-folder anyways should not beloong to pre-installation generic and make sense to restrict only to the rolse iscsi server

salt/monitoring/init.sls
```diff
@@ -0,0 +1,131 @@
...
+  file.managed:
+    - name: /etc/prometheus/prometheus.yml
+    - makedirs: True
+    - contents: |
+        scrape_configs:
+          - job_name: 'sumaform'
+            scrape_interval: 5s
+            static_configs:
+         ## TODO: grains server need to be passed to monitoring module , change localhost
+              - targets: [localhost:8001'] # hanadb_exporter
```
MalloZup: note for reviewers: this is so intentionally. We can inject the IPSs later on. I think we can do later . For moment the module will just add grafana/proemetheus working.

salt/hana_node/sap_inst.sls
```diff
@@ -4,12 +4,14 @@ nfs-client:
      attempts: 3
      interval: 15
 
-sap_inst_directory:
+sap-inst-directory:
```
arbulu89: Why have you changed this? I have found many other state names with underscore. I'm in favour for unifying, but in that case change all of them

salt/hana_node/sap_inst.sls
```diff
@@ -4,12 +4,14 @@ nfs-client:
...
   file.directory:
     - name: {{grains['hana_inst_folder']}}
     - user: root
     - mode: 755
     - makedirs: True
+
+mount-sap-inst-directory:
```
arbulu89: I don't see necessary this change, but I guess is not harmful

salt/hana_node/sap_inst.sls
```diff
@@ -28,6 +31,7 @@ sap_inst_directory:
     - mkmnt: True
     - persist: True
     - opts: vers=3.0,username={{grains['storage_account_name']}},password={{grains['storage_account_key']}},dir_mode=0777,file_mode=0777,sec=ntlmssp
-    - required:
-      - nfs-client
-  {% endif %}
+    - require:
+      - pkg: nfs-client
+      - file: sap-inst-directory
+{% endif %}
```
arbulu89: Is this `endif` correctly indented. I doens't look like here at least

salt/monitoring/init.sls
```diff
@@ -0,0 +1,131 @@
+# TODO: this repo should detect the os itself and chooose right repo depending the os
```
arbulu89: I was wondering if we can split the logic of this file in several subfiles... Just to improve readability...
I would split in: repos, prometheus, grafana...

salt/monitoring/init.sls
```diff
@@ -0,0 +1,131 @@
...
+
+prometheus:
+  pkg.installed:
+    - name: golang-github-prometheus-prometheus
+    - require:
+      - pkgrepo: suse-manager-head-repo
+      - pkgrepo: sle-15-update
+      - pkgrepo: sle-15-pool
+
+prometheus_configuration:
```
arbulu89: For example here we are using underscore instead of dash

salt/monitoring/init.sls
```diff
@@ -0,0 +1,131 @@
...
+  file.managed:
+    - name: /etc/prometheus/prometheus.yml
+    - makedirs: True
+    - contents: |
+        scrape_configs:
+          - job_name: 'sumaform'
```
arbulu89: sumaform?

salt/monitoring/init.sls
```diff
@@ -0,0 +1,131 @@
...
+            static_configs:
+         ## TODO: grains server need to be passed to monitoring module , change localhost
+              - targets: [localhost:8001'] # hanadb_exporter
+          - job_name: 'hanadb'
+            scrape_interval: 5s
+            static_configs:
+         ## TODO: change localhost  with grain
+              - targets: ['localhost:9800']
+
+prometheus_service:
```
arbulu89: Just curiosity, installing prometheus doesn't install the service files?

libvirt/terraform/modules/monitoring/variables.tf
```diff
@@ -0,0 +1,99 @@
...
+}
+
+variable "public_key_location" {
+  description = "path of additional pub ssh key you want to use to access VMs"
+  default     = "/dev/null"
+
+  # HACK: "" cannot be used as a default because of https://github.com/hashicorp/hil/issues/50
+}
+
+variable "scenario_type" {
```
arbulu89: Is this parameter needed?

salt/monitoring/init.sls
arbulu89: Is the file parsing working? I would say that a `'` is missing in the targets entry

MalloZup: Yep is working

salt/hana_node/sap_inst.sls
MalloZup: Right I will redo'. I agree we should use a std format

MalloZup:
Thx for review I will adapt

salt/hana_node/sap_inst.sls
MalloZup: https://github.com/SUSE/ha-sap-terraform-deployments/issues/105

salt/hana_node/sap_inst.sls
MalloZup: removed

salt/hana_node/sap_inst.sls
MalloZup: i just removed this change for moment they are not necessary

salt/monitoring/init.sls
MalloZup: yes, in theory we should aim in future what sumaform has, like we have a dir of `repos` https://github.com/uyuni-project/sumaform/tree/master/salt/repos/repos.d.

MalloZup: https://github.com/SUSE/ha-sap-terraform-deployments/issues/106

salt/monitoring/init.sls
MalloZup: saltstack use underscores as convention, i just find it out now so ok with this. We need to downcase the others

salt/monitoring/init.sls
MalloZup: thx done

libvirt/terraform/modules/monitoring/variables.tf
MalloZup: good catch not

salt/monitoring/init.sls
MalloZup: good catch. Modified and removed. We enable only the service now

MalloZup:
foo

PR #104, opened 2019-07-22
Update the pre configuration states to support sle15sp1 (and bugfixing)

arbulu89:
Salt formula sates are installed in different path beyond sle15sp1.

Besides that, some hardcoded repository names has been removed, as the packages might come from packageHub now the salt formulas are released there.

### DIFF
--- a/libvirt/terraform/README.md
+++ b/libvirt/terraform/README.md
@@ -21,7 +21,7 @@
 
 # Howto
 
-To deploy the cluster only the parameters of three files should be changed: 
+To deploy the cluster only the parameters of three files should be changed:
 
 * [main.tf](main.tf)
 
@@ -34,14 +34,14 @@ You can between following profiles:  performance optimized, cost optimized
 
 ___
 Performance optimized:
-   * [hana.sls](../../pillar_examples/libvirt/performance_optimized/hana.sls) 
+   * [hana.sls](../../pillar_examples/libvirt/performance_optimized/hana.sls)
    * [cluster.sls](../../pillar_examples/libvirt/performance_optimized/cluster.sls).
 
 ___
 
 Cost optimized:
 
-   * [hana.sls](../../pillar_examples/libvirt/cost_optimized/hana.sls) 
+   * [hana.sls](../../pillar_examples/libvirt/cost_optimized/hana.sls)
    * [cluster.sls](../../pillar_examples/libvirt/cost_optimized/cluster.sls).
 
 
@@ -66,6 +66,9 @@ additional_repos = {}
 shared_storage_type = "iscsi"
 iscsi_srv_ip = "192.168.XXX.Y+2"
 iscsi_image = "url-to-your-sles4sap-image" # sles15 or above
+
+# Monitoring system data
+monitoring_srv_ip = "192.168.XXX.Y+3"
 
 
 # Repository url used to install install HA/SAP deployment packages"
@@ -106,7 +109,7 @@ Have a look at  [specifications](#specifications) for more details.
 
 This project is mainly based in [sumaform](https://github.com/uyuni-project/sumaform/)
 
-Components: 
+Components:
 
 - **modules**: Terraform modules to deploy a basic two nodes SAP HANA environment.
 - **salt**: Salt provisioning states to configure the deployed machines with the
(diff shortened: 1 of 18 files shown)


arbulu89:
Tested using libvirt and AWS with SLE15 and SLE15SP1. Everything working as expected.

pillar_examples/automatic/hana.sls
```diff
@@ -29,6 +29,10 @@ hana:
           user_name: SYSTEM
           user_password: YourPassword1234
           database: SYSTEMDB
+      exporter:
```
MalloZup: we need to add documentation for this in monitoring

MalloZup: there is a monitoring section markdown file where we can add information about this.

MalloZup: https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/libvirt/terraform/doc/monitoring.md

salt/deployment.sh
```diff
@@ -1,13 +1,13 @@
...
 
-salt-call --local --file-root=/root/salt/ \
+salt-call --local --file-root=/root/salt \
     --log-level=info \
     --log-file=/tmp/salt-pre-installation.log \
     --log-file-level=all \
     --retcode-passthrough \
     --force-color state.apply pre_installation || exit 1
 
-salt-call --local --file-root=/root/salt/ \
```
MalloZup: why are we removing this?

arbulu89: Because the PR changes the salt minion configuration (`/etc/salt/minion`) and this path is added to the `file_roots` (during the `pre_installation` salt state execution). After doing that, we are not required to have the path explicitly there)

pillar_examples/automatic/hana.sls
arbulu89: There is already some documentation in the `saphanabootstrap-formula` project:
https://github.com/SUSE/saphanabootstrap-formula/blob/master/pillar.example#L33

Anyway, we can add it there in the monitoring docs too. Let me add it there

arbulu89: Docs updated

arbulu89:
@MalloZup I have added more changes fixing the IP addresses collision. Now the VMs are deployed with different addresses.
For that, I have included an extra parameter to define the monitoring server ip address.

**INFO: This doesn't fix the targets selection in prometheus, this is done incorrectly yet. This will be fixed in future PR**

libvirt/terraform/README.md
```diff
@@ -67,6 +67,9 @@ shared_storage_type = "iscsi"
 iscsi_srv_ip = "192.168.XXX.Y+2"
 iscsi_image = "url-to-your-sles4sap-image" # sles15 or above
 
+# Monitoring system data
```
MalloZup: imho regarding documentation, we should start to isolated to much places for documenting and just put the documentation for monitoring in monitoring.md

libvirt/terraform/README.md
```diff
@@ -148,6 +152,7 @@ data.
 - **shared_storage_type**: Shared storage type between iscsi and KVM raw file shared disk. Available options: `iscsi` and `shared-disk`.
 - **iscsi_srv_ip**: IP address of the machine that will host the iscsi target (only used if `iscsi` is used as a shared storage for fencing)
 - **iscsi_image**: Source image of the machine hosting the iscsi target (sles15 or above) (only used if `iscsi` is used as a shared storage for fencing)
+- **monitoring_srv_ip**: IP address of the machine that will host the monitoring stack
```
MalloZup: I'm thinking that we should perhaps move it only to monitoring part. There is so much info imho

arbulu89:
> great JOB! my only concern is that imho we should document everything about monitoring in the monitoring.md
> the actual documentations is to much information
> In theory our documentation should be modularized too, like data for monitoring module in monitoring.md
> and the main.readme should imho actually contains only links to various modules.
> ( this is a tech-debt in the IA information architecture)
> But for this pr, I think we should perhaps move all the monitoring doc to md. what do you think?

Agree! The aim of this last commit was just to fix the bug and let the project working. We will work on that next

MalloZup:
feel free to ertge

PR #110, opened 2019-07-25
WIP: update doc with main.tf.example

MalloZup:
# description:

this pr introduce a way to select modules.

The entrypoint is the main.tf.

An user should only copy and paste `main.tf.example` modules to its `main.tf` for enabling and disabling.

This is also used in sumaform and terraform upstream modules

see the various examples.. https://github.com/uyuni-project/sumaform/blob/master/main.tf.libvirt-testsuite.example
etc

# fix

https://github.com/uyuni-project/sumaform/blob/master/main.tf.libvirt-testsuite.example


MalloZup:
there are some downsides of this approach, since one need to update the `output.tf` file.

I will explore also the `enabled option`

nick-wang:
> there are some downsides of this approach, since one need to update the `output.tf` file.
> 
Just a thought, Maybe also create files `output.tf.xxx`, corresponding to its `main.tf.xxx`. Could use symbol link it different `main.tf.xxx` can share the same `output.tf.xxx`.

MalloZup:
@nick-wang  thx for sharing.

I think my idea would be to remove the `output.xxx` and encapsulate it directly to main.tf. this is possible

so we don't have all this manual operation.

MalloZup:
see my last example https://github.com/SUSE/ha-sap-terraform-deployments/pull/110/commits/7527c681e755a40620985fde9a75277f4883e825

MalloZup:
@arbulu89  thx for comment. I agree  with your question: who are the users of this project?

If we don't answer this question explicitly IN CODE we will stay in the middle and bringing a project without shape.

From My point, we need to be developer/engineer first oriented, since we are the first user.

Although we can then build from the simplicity we build for dev, and we provide a way for users to use it. But imho we shouldn't target both at same level.

On the other side, having an inventory of supported examples for users is much more easier then having to much configuration in entrypoint.

We use the same approach in the formula with `scenarios`,  we can reuse the concept of  scenarios.

E.g currently we support 2 scenarios: `monitoring` and `minimal-cluster`(2 nodes).

In future this can be added. If we give fixed possibility to users, it is more easy to understand and we can also drive it.

nick-wang:
With the implementing of more modules and scenarios(use case of the modules combination), we may need more and more variables, many of then just to *enable/disable* the modules. 

And conditions many become awkward in one `main.tf` with complex usage, though it is fine with the current scenarios.

For users, it may easier to pickup one `main.tf.example` most suitable to it needs, then slightly polish as the need. Compare to modify a bigger file with more modules disabled.
The drawbacks of using multiple `main.tf.example` is it much include many duplicate modules exist  in many files. Modify one minor place may need to modify others as well, really easy to miss and require more testing... Effort will multiple with more providers

MalloZup:
I think we need to go in the direction of templates because having variables which control the behaviours is really tricky thing. In lonterm we should decrease the number of behavioural variables at the minimal size.

Having a template of module.tf, I don't think we will have to much templated. Maximial 5 in a  really really longterm imho.

@ldevulder @alvarocarvajald thoughts on this? this will be the direction also from cloud provider.

arbulu89:
@MalloZup @nick-wang I agree that if we manage correctly the `main.tf` example files the outcome can be really good. Anyway, we are always only focusing in `libvirt`... We need to extrapolate all the changes to the other providers to have a unified deployment. The cloud providers are quite far away of this solution.

And of course, if we are going to follow the examples file approach, we need to remove the flags used to create the modules, update docs, etc

I think this PR has a lot of work to do.

But, you have my approval for following this approach instead of using flags!

nick-wang:
Just a immature idea and may slightly digress:
Maintain some modules file, each one only define one module sourced from `"./modules/xxx`.
Then define one rule files, including i want 1 `base` , 1 `sbd_disk`, 2 `hana_node`, 0 `iscsi_server`, 0 `monitoring`. Then script will generate(concat) the modules to a `main.tf` based on the pre-defined rules. (It actually works like *flags* and *counts*, but not show in `variables.tf` and `main.tf`)
But we need to write a *scenarios markdown* to users for some recommend/support combinations.
It can reduce the duplicate module/code exist in different `main.tf.example`, but may increase the testing for multiple/flexible combinations. 
Potential issue may come with incompatible modules in future. Like actually 1 `monitoring` module make no sense with 0 `hana_node`. (but in this case, monitoring may more suitable to be a sub module of hana_node)

MalloZup:
we can adress this later on for moment closing

PR #112, opened 2019-07-26
Enable hanadb_exporter only beyond sle15 in pillar

arbulu89:
As python-prometheus_client package is only available beyond SLE15 by now, the best is to disable the deployment of the exporter for older versions.

### DIFF
--- a/pillar_examples/automatic/hana.sls
+++ b/pillar_examples/automatic/hana.sls
@@ -29,10 +29,12 @@ hana:
           user_name: SYSTEM
           user_password: YourPassword1234
           database: SYSTEMDB
+      {% if grains['osrelease_info'][0] == 15 %}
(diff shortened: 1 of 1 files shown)


pillar_examples/automatic/hana.sls
```diff
@@ -29,10 +29,12 @@ hana:
           user_name: SYSTEM
           user_password: YourPassword1234
           database: SYSTEMDB
+      {% if grains['osrelease_info'][0] == 15 %}
```
MalloZup: @arbulu89  perhaps we can add a comment here, saying why we are disabling it so we explain the rationale. I think this would fit in the codebase

arbulu89:
Closing PR as packages for older SLE12 version has been created

PR #114, opened 2019-07-26
add node_exporter for nodes and sles15 only

MalloZup:
# Description:

- add `node_exporter service` to each host. This can be disabled with `monitoring_enabled option`, by default is set to True (enabled). The rationale behind this is that we will need to have other exporter, so a monitoring_enabled will just install them. It is not fine_grained but I think it should be ok, also in future we might move the node_exporter etc to formulas when the pkg repository are morestable. So I think for moment is pretty ok.

- add a mechanism to monitor multiple service, monitored_services. We have a list of HOST:PORT.
  see https://github.com/SUSE/ha-sap-terraform-deployments/pull/114/files#diff-20a93b26850139589fffe838def26807R41

- add repos pkg needed, also this `https://download.opensuse.org/repositories/server:/monitoring/SLE_15/` for monitoring

- add documentation needed.

### DIFF
--- a/libvirt/terraform/modules/host/salt_provisioner.tf
+++ b/libvirt/terraform/modules/host/salt_provisioner.tf
@@ -53,7 +53,6 @@ additional_packages: [${join(", ", formatlist("'%s'", var.additional_packages))}
 authorized_keys: [${trimspace(file(var.base_configuration["public_key_location"]))},${trimspace(file(var.public_key_location))}]
 host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]
 host_ip: ${element(var.host_ips, count.index)}
-monitoring_enabled: ${var.monitoring_enabled}
 ${var.grains}
 
 EOF
--- a/libvirt/terraform/modules/host/salt_provisioner.tf
+++ b/libvirt/terraform/modules/host/salt_provisioner.tf
@@ -53,7 +53,7 @@ additional_packages: [${join(", ", formatlist("'%s'", var.additional_packages))}
 authorized_keys: [${trimspace(file(var.base_configuration["public_key_location"]))},${trimspace(file(var.public_key_location))}]
 host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]
 host_ip: ${element(var.host_ips, count.index)}
-
+monitoring_enabled: ${var.monitoring_enabled}
 ${var.grains}
 
 EOF
--- a/libvirt/terraform/modules/host/variables.tf
+++ b/libvirt/terraform/modules/host/variables.tf
@@ -97,3 +97,8 @@ variable "additional_disk" {
   description = "disk block definition(s) to be added to this host"
   default     = []
 }
+
+variable "monitoring_enabled" {
+  description = "enable the host to be monitored by exporters, e.g node_exporter"
+  default     = true
+}
--- a/libvirt/terraform/modules/host/variables.tf
+++ b/libvirt/terraform/modules/host/variables.tf
@@ -97,8 +97,3 @@ variable "additional_disk" {
   description = "disk block definition(s) to be added to this host"
   default     = []
 }
-
-variable "monitoring_enabled" {
-  description = "enable the host to be monitored by exporters, e.g node_exporter"
-  default     = true
-}
--- a/salt/hana_node/monitoring.sls
+++ b/salt/hana_node/monitoring.sls
@@ -1,0 +1,40 @@
+server_monitoring_repo:
+ pkgrepo.managed:
+    - humanname: Server:SLE15:Monitoring
+    - baseurl: https://download.opensuse.org/repositories/server:/monitoring/SLE_15/
+    - refresh: True
+    - gpgautoimport: True
+
+sle_15_update:
+ pkgrepo.managed:
+    - humanname: SLE15:Update
+    - baseurl: http://download.suse.de/ibs/SUSE/Updates/SLE-Module-Basesystem/15/x86_64/update/
+    - refresh: True
+    - gpgautoimport: True
+
+sle_15_pool:
+ pkgrepo.managed:
+    - humanname: SLE15:Pool
+    - baseurl: http://download.suse.de/ibs/SUSE/Products/SLE-Module-Basesystem/15/x86_64/product/
+    - refresh: True
+    - gpgautoimport: True
+
+prometheus_node_exporter:
+  pkg.installed:
+    - name: golang-github-prometheus-node_exporter
+    - require:
+      - pkgrepo: server_monitoring_repo
+      - pkgrepo: sle_15_update
+      - pkgrepo: sle_15_pool
+
+node_exporter_service:
+  service.running:
+    - name: prometheus-node_exporter
+    - enable: True
+    - restart: True
+    - require:
+      - pkg: prometheus_node_exporter
+      - pkgrepo: server_monitoring_repo
+      - pkgrepo: sle_15_update
+      - pkgrepo: sle_15_pool
+
--- a/salt/hana_node/monitoring.sls
+++ b/salt/hana_node/monitoring.sls
@@ -5,27 +5,11 @@ server_monitoring_repo:
     - refresh: True
     - gpgautoimport: True
 
-sle_15_update:
- pkgrepo.managed:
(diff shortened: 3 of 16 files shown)


salt/monitoring/prometheus/prometheus.yml
```diff
@@ -7,6 +7,7 @@ scrape_configs:
   - job_name: 'hanadb'
     scrape_interval: 5s
     static_configs:
-      {% for ip in grains['host_ips'] %}
-      - targets: ['{{ ip }}:8001'] # hanadb_exporter port TODO: we can improve this 
+      # servive is an entry like : HOST_IP:PORT 192.XX.XX.1:8001
+      {% for service in grains['monitored_services'] %}
```
arbulu89: What happens here if the monitored services list is empty? It would fail, right? We should cover this use case too (where we don't deploy any exporter)

arbulu89: Besides that, with this implementation, you are creating a lot of targets entries. Is this valid? I doesn't look like.
Now you would create (supposing we have more than 1 target):

```
static_configs:
    - targets: ['localhost:9000']
    - targets: ['localhost:9001']
```
Shouldn't we have the next:
```
static_configs:
    - targets: ['localhost:9000', 'localhost:9001']
```

libvirt/terraform/modules/host/salt_provisioner.tf
```diff
@@ -53,7 +53,7 @@ additional_packages: [${join(", ", formatlist("'%s'", var.additional_packages))}
 authorized_keys: [${trimspace(file(var.base_configuration["public_key_location"]))},${trimspace(file(var.public_key_location))}]
 host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]
 host_ip: ${element(var.host_ips, count.index)}
-
+monitoring_enabled: ${var.monitoring_enabled}
```
arbulu89: I think it would be better to add this in the `hana_node` module. Eventually, only hana nodes (and netweaver in the future) are the hosts we want to only monitor, not all the hosts (monitoring machine for example).

Actually, right now, we cannot change this value (the default value would be used always) as we are not bypassing the value through `hana_node` module

libvirt/terraform/modules/host/variables.tf
```diff
@@ -97,3 +97,8 @@ variable "additional_disk" {
   description = "disk block definition(s) to be added to this host"
   default     = []
 }
+
+variable "monitoring_enabled" {
```
arbulu89: Same here. IMO it should be moved to `hana_node`

salt/hana_node/monitoring.sls
```diff
@@ -0,0 +1,40 @@
+server_monitoring_repo:
```
arbulu89: Maybe we could take advantage of the already create salt prometheus formula to start the node exporter:
https://github.com/salt-formulas/salt-formula-prometheus/blob/master/prometheus/exporters.sls

Just as a comment XD

salt/hana_node/monitoring.sls
```diff
@@ -0,0 +1,40 @@
...
+ pkgrepo.managed:
+    - humanname: Server:SLE15:Monitoring
+    - baseurl: https://download.opensuse.org/repositories/server:/monitoring/SLE_15/
+    - refresh: True
+    - gpgautoimport: True
+
+sle_15_update:
+ pkgrepo.managed:
+    - humanname: SLE15:Update
+    - baseurl: http://download.suse.de/ibs/SUSE/Updates/SLE-Module-Basesystem/15/x86_64/update/
```
arbulu89: Are thes url-s links publicly available? Remember that this project must work with any network configuration
(Actually, as internal url, we shouldn't publish them here...)

salt/monitoring/init.sls
```diff
@@ -1,20 +1,20 @@
...
+    - humanname: Server:SLE15:Monitoring
+    - baseurl: https://download.opensuse.org/repositories/server:/monitoring/SLE_15/
     - refresh: True
     - gpgautoimport: True
 
-sle-15-update:
+sle_15_update:
  pkgrepo.managed:
     - humanname: SLE15:Update
     - baseurl: http://download.suse.de/ibs/SUSE/Updates/SLE-Module-Basesystem/15/x86_64/update/
```
arbulu89: The same comment than before... I passed that in the last PR

salt/monitoring/prometheus/prometheus.yml
MalloZup: I think having multiple targes entry is more readable then an enourmous list. I will fix the emptyness.thx

libvirt/terraform/modules/host/salt_provisioner.tf
MalloZup: I think in beginning I was thinking the same on `hana_node` but I thought to make more generic.

however I agree that doing on `hana_node` make more sense.

changing it

MalloZup: done

libvirt/terraform/modules/host/variables.tf
MalloZup: done

salt/hana_node/monitoring.sls
MalloZup: imho relying on upstream formulas is more tricky and bug prones. I think we can for moment skip this imho but thx for sharing

salt/hana_node/monitoring.sls
MalloZup: It is ok to have the repo links outside , e.g take look at suma (https://github.com/uyuni-project/sumaform/blob/master/salt/repos/repos.d/Devel_Galaxy_Manager_Head_ToSLE.repo)
and they have plenty of other repos in GitHub and that is ok also from sec pov.

Regarding the internal vs public thing,  I think that If we are assuming the system is a SLES15 there is not such a way we can provide the repos. I might just add a check of version if SLE15 install that repos.

What do you think?

diegoakechi: @MalloZup as this is a SLES instance, for the Pool and Update repos, I would prefer to have the repos configured via registration, using `SUSEConnect` for example as it is the official way. The current way will not work outside SUSE internal network.

MalloZup: ok make sense thx! 
perhaps the suseConnect way add already things?  I will do that way

diegoakechi: @MalloZup I assume that the default salt provisioning step is being executed on the monitoring machine:
https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/salt/default/registration.sls
If it is the case, the system will be already registered and you can remove this step and request salt only to install the packages directly as the repositories will be already there.

MalloZup:
ok thx. I have adressed most of the points. Testing it

MalloZup:
@arbulu89 and all this works now. adreessed all comments. Please re-review
:dog2:

salt/monitoring/prometheus/prometheus.yml
arbulu89: Is this change fixing the issue?
If I'm not wrong this implementation would do exactly the same than before.
I'm not totally sure how the yml file is read here, but if wi don't have any monitored service here we will have something finishing in:

```
static_config:
```
Is this end of file valid? I mean, as we are not setting any value to this entry.

salt/hana_node/monitoring.sls
arbulu89: We sill have non public urls here...

MalloZup: Forgot to push changes somehow

salt/monitoring/prometheus/prometheus.yml
MalloZup: great adressed. testing

MalloZup: https://github.com/SUSE/ha-sap-terraform-deployments/pull/114/files#diff-44184760a4384d0057c0ba11b51c0fdfR6

MalloZup:
feel free to re-review @ayoub-belarbi @arbulu89 tia

MalloZup:
I habe testéd this 🐩🐩🐩🎊🇧🇪🇧🇪🇧🇪🇧🇪🇧🇪

PR #115, opened 2019-07-30
Minor fixup for iscsi partitioning and default repo requirement.

nick-wang:
Two minor fixes for:
+ fix the `repos` requirement in `pkgs.sls` with folder name *default*

+ Change start from `0` to `1` to make partition aligned, also use percentage instead of fixed position.

### DIFF
--- a/aws/terraform_salt/salt_provisioner.tf
+++ b/aws/terraform_salt/salt_provisioner.tf
@@ -55,20 +55,20 @@ ha_sap_deployment_repo: ${var.ha_sap_deployment_repo}
 
 partitions:
   1:
-    start: 0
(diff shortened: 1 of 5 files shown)


arbulu89:
@nick-wang @juadk In this case, I would extrapolate the change to the rest of the providers too (aws, azure and gcp). I don't want to let behind the changes in the cloud providers.
@nick-wang You can find the same structure in each of the folder (in `terraform_salt`)

nick-wang:
@juadk @arbulu89 @MalloZup Thanks for review. Add the change to all cloud providers.
Will change cloud providers together if (likely) no conflict in the later pr.

PR #118, opened 2019-08-05
Handle default for grain monitoring

### LINKED ISSUE #117: Jinja variable 'dict object' has no attribute 'monitoring_enabled'
Applying the GCP salted configuration with `init_type = "skip-hana"`, the deployment fails with the following error:

```
local:
    Data failed to compile:
----------
    Rendering SLS 'base:hana_node' failed: Jinja variable 'dict object' has no attribute 'monitoring_enabled'
```

Checked also with `init_type = "all"`, and the same issue is observed.

MalloZup:
fix #117

### DIFF
--- a/salt/hana_node/init.sls
+++ b/salt/hana_node/init.sls
@@ -18,6 +18,6 @@ include:
   - hana_node.hana_packages
   - hana_node.cluster_packages
   - hana_node.formula
-  {% if grains['monitoring_enabled'] %}
+  {% if grains.get('monitoring_enabled') %}
   - hana_node.monitoring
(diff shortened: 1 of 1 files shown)


juadk:
> The changes look good, but we need to test it carefully.
> By now, the `qa_mode` variable is a string instead of a boolean (we should change that...). And the default value is `false` as string. I don't know how the new comparison manages this (even how the `sameas false` was working...)

I agree it could be better with a boolean instead of a string, let me fix that.
It will be clearer.

MalloZup:
ok we can merge this and improve later thx

PR #119, opened 2019-08-05
remove Running variable

### LINKED ISSUE #113: remove variable running
```
variable "running" {
  description = "whether this host should be turned on or off"
  default     = true
}
```

this variable is cosmetical to me. When we deploy a VM we want it always running.

We should keep our codebase KISS and remove it.

@nick-wang @ldevulder @alvarocarvajald any thouhgts on this? do you want to keep it? I would be in favor to remove it, since nobody use it.

MalloZup:
fix #113 . Discussed above. the running variable is not needed for our usa-case so a variable less is less behaviour to maintain

### DIFF
--- a/libvirt/terraform/modules/hana_node/main.tf
+++ b/libvirt/terraform/modules/hana_node/main.tf
@@ -33,7 +33,6 @@ EOF
   // Provider-specific variables
   memory         = "${var.memory}"
   vcpu           = "${var.vcpu}"
-  running        = "${var.running}"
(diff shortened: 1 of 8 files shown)


MalloZup:
(https://github.com/dmacvicar/terraform-provider-libvirt/blob/19c23424ccfdecf6dfcbdd87783bf835ba5f585e/website/docs/r/domain.html.markdown#L34)

MalloZup:
yes tested it against libv

PR #120, opened 2019-08-05
Change qa_mode as a boolean

juadk:
Using "true" and "false" as a string could be misunderstand.
This PR changes qa_mode variable from a string to a boolean type.

**Tests:**
AWS: OK
Azure: OK
GCP: OK


juadk:
> It looks fine!
> The unique thing would be to add the changes to libvirt.
> This variable is only available in the `iscsi_srv` module, so the changes are quite straightforward

Hey @arbulu89 & @MalloZup 
Changes added to libvirt.

PR #122, opened 2019-08-06
Improve the registration process to be optional

juadk:
Now we (QA) have public cloud images with salt-minion pre-installed, so the registration process could be optional to let the user choose the proper way.
That's why this PR add the variable `install_salt_minion`.

To be the most clear as possible, a file is created in /tmp to know if the system was registered or not:
`register_to_install_salt` or `not_register_to_install_salt`
You could easily know in which scenario you are.

From a QA perspective, we prefer not to register the system.

**Tests:**
AWS: **OK with and without salt-minion pre-installed with SLE12SP4**
GCP: **OK with and without salt-minion pre-installed with SLE15**
Azure: **OK with and without salt-minion pre-installed with SLE15SP1**

### DIFF
--- a/aws/terraform_salt/terraform.tfvars.example
+++ b/aws/terraform_salt/terraform.tfvars.example
@@ -77,6 +77,10 @@ ha_sap_deployment_repo = ""
 # Run provisioner execution in background
 #background = true
 
+# Enable salt-minion installation from SCC. Some cloud images don't bring this package pre-installed.
+# WARNING: This flag requires the usage of `reg_code`
+# install_salt_minion = true
+
 # QA variables
 
 # Define if the deployment is using for testing purpose
--- a/salt/salt_provisioner_script.tpl
+++ b/salt/salt_provisioner_script.tpl
@@ -4,9 +4,13 @@ mv /tmp/salt /root || true
 
 # SCC Registration to install salt-minion
 
+# The iSCSI server won't be de-registered as it needs to install some additional packages.
 if grep -q 'role: iscsi_srv' /tmp/grains; then
   sh /root/salt/install-salt-minion.sh -r ${regcode}
-elif [[ ! -e /usr/bin/salt-minion ]]; then
+
+# System is registered to install salt-minion and de-registered afterwards
+# if the variable install_salt_minion is true
+elif grep -q 'install_salt_minion: 1' /tmp/grains; then
   sh /root/salt/install-salt-minion.sh -d -r ${regcode}
 fi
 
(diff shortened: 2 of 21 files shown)


aws/terraform_salt/terraform.tfvars.example
```diff
@@ -77,6 +77,11 @@ ha_sap_deployment_repo = ""
 # Run provisioner execution in background
 #background = true
 
+# Define if you need to install the salt-minion because
```
arbulu89: I would update the description to;
```
Enable salt-minion installation from SCC. Some cloud images don't bring this package pre-installed.
WARNING: This flag requires the usage of `reg_code`
install_salt_minion = true
```

salt/salt_provisioner_script.tpl
```diff
@@ -4,10 +4,18 @@ mv /tmp/salt /root || true
...
+# and we keep it registered
 if grep -q 'role: iscsi_srv' /tmp/grains; then
   sh /root/salt/install-salt-minion.sh -r ${regcode}
-elif [[ ! -e /usr/bin/salt-minion ]]; then
+
+# We register the system to install salt-minion and then de-register it
+# if the variable install_salt_minion is true
+elif grep -q 'install_salt_minion: 1' /tmp/grains; then
   sh /root/salt/install-salt-minion.sh -d -r ${regcode}
+  touch /tmp/register_to_install_salt
```
arbulu89: As we commented, this file might be removed and check if the registration was done with
```
cat /etc/salt/grains | grep install_salt_minion
```
Anyway, it's not harmful, so up to you XD

salt/salt_provisioner_script.tpl
```diff
@@ -4,10 +4,18 @@ mv /tmp/salt /root || true
 
 # SCC Registration to install salt-minion
 
+# The iSCSI server have to be register to install additional packages
```
arbulu89: Some typos here. I would change to:
```
The iSCSI server won't be de-registered as it needs to install some additional packages.

```

salt/salt_provisioner_script.tpl
```diff
@@ -4,10 +4,18 @@ mv /tmp/salt /root || true
...
+# and we keep it registered
 if grep -q 'role: iscsi_srv' /tmp/grains; then
   sh /root/salt/install-salt-minion.sh -r ${regcode}
-elif [[ ! -e /usr/bin/salt-minion ]]; then
+
+# We register the system to install salt-minion and then de-register it
```
arbulu89: I would change to:
```
System is registered to install salt-minion and de-registered afterwards.
```

aws/terraform_salt/terraform.tfvars.example
juadk: I agree, thanks to point this WARNING.

salt/salt_provisioner_script.tpl
juadk: I decided not to use a file in /tmp.

juadk:
The code has changed a lot, I will another pr later.

PR #125, opened 2019-08-07
Add terraform and salt modules to deploy SAP NW

arbulu89:
The PR is still in a really early stage. Only available as a reference, if anyone else want to try SAP NW deployments.

It depends in the latest code of:
python-shaptools
salt-shaptools
sapnwbootsrap-formula


arbulu89:
Closing this PR as I will create a new one already using terraform 0.12 version

PR #133, opened 2019-08-21
Hawk apiserver monitoring

ayoub-belarbi:


libvirt/terraform/README.md
```diff
@@ -154,6 +154,7 @@ data.
 - **iscsi_image**: Source image of the machine hosting the iscsi target (sles15 or above) (only used if `iscsi` is used as a shared storage for fencing)
 - **monitoring_srv_ip**: IP address of the machine that will host the monitoring stack
 - **ha_sap_deployment_repo**: Repository with HA and Salt formula packages. The latest RPM packages can be found at [https://download.opensuse.org/repositories/network:/ha-clustering:/Factory/{YOUR OS VERSION}](https://download.opensuse.org/repositories/network:/ha-clustering:/Factory/)
+- **install_from_ha_sap_deployment_repo**: Whether or not to install HA/SAP packages from ha_sap_deployment_repo
```
MalloZup: looks good now. I'm thinking only perhaps we could change the variable to be more end-goal oriented.

`monitoring_from_ha_repo` or something more usegfull.. :grin:

MalloZup: or `use_pkg_from_ha_repo`? dunno...

ayoub-belarbi: Thanks for the review, what that variable is doing is controlling if the HA/SAP packages are installed from there the `ha_sap_deployment_repo` and can be used for testing the packages and not for monitoring, but I will think right now for a better name! Thanks!

salt/hana_node/cluster_packages.sls
```diff
@@ -3,3 +3,24 @@ habootstrap-formula:
     - retry:
         attempts: 3
         interval: 15
+
+{% if grains['install_from_ha_sap_deployment_repo'] is defined and grains['install_from_ha_sap_deployment_repo'] == true %}   # TODO: Very long line
+install_cluster_packages: # Change the name
+  pkg.installed:
+    - fromrepo: {{ grains['ha_sap_deployment_repo'] }}
```
diegoakechi: @ayoub-belarbi I don't think it is necessary since you are changing the repo priority, the packages would come from there anyway, so I would suggest to drop this change and have the parameter only to set the priority you want for the ha-factory repo.

If you still think it makes sense, I think it is cleaner to have the param receiving the repo name, and them you install from it, this way it is not limited to this specific repo.

salt/pre_installation/ha_repos.sls
```diff
@@ -3,7 +3,12 @@ ha-factory-repo:
     - name: ha-factory
     - baseurl: {{ grains['ha_sap_deployment_repo'] }}
     - gpgautoimport: True
+    {% if grains['install_from_ha_sap_deployment_repo'] is defined and grains['install_from_ha_sap_deployment_repo'] == true %}  # TODO: Very long line
+    - priority: 98
```
diegoakechi: @ayoub-belarbi Instead of this IF, I think would be better to have it set from a parameter and in case of this parameter is `None`, we assume the default `110`.

salt/hana_node/cluster_packages.sls
MalloZup: I agree with Diego that if we can spare us an IF with more implicit mechanism is better. good catch

salt/monitoring/prometheus/prometheus.yml
```diff
@@ -1,15 +1,34 @@
...
-    scrape_interval: 5s
+  - job_name: hanadb
     static_configs:
-      # servive is an entry like : HOST_IP:PORT 192.XX.XX.1:8001
-      {% for service in grains['monitored_services'] %}
-      - targets: ['{{ service }}'] # service to be monitored
-      {% endfor %}
+      - targets:
+        {% for ip in grains['monitored_services'] %}
+        - "{{ ip }}:8001"
```
MalloZup: can we add here a comment that describe which service correspond to port
like
`
- "{{ ip }}:8001" # node_exporter
`

salt/monitoring/prometheus/prometheus.yml
```diff
@@ -1,15 +1,34 @@
...
+        {% endfor %}
+
+  - job_name: node
+    static_configs:
+      - targets:
+        {% for ip in grains['monitored_services'] %}
+        - "{{ ip }}:9100"
```
MalloZup: ditto

libvirt/doc/monitoring.md
```diff
@@ -43,6 +43,33 @@ hana:
...
 
+# Enable the HA exporter
+
+The HA Prometheus metrics are exported using the hawk-apiserver [hawk-apiserver](https://github.com/ClusterLabs/hawk-apiserver).
+In order to enable the exporter for each cluster node `cluster` pillar entries must be modified.
+
+Here an example:
+
+```
+cluster:
```
MalloZup: Perhaps we could create an example file

salt/monitoring/prometheus/prometheus.yml
```diff
@@ -1,15 +1,34 @@
...
-    scrape_interval: 5s
+  - job_name: hanadb
     static_configs:
-      # servive is an entry like : HOST_IP:PORT 192.XX.XX.1:8001
-      {% for service in grains['monitored_services'] %}
-      - targets: ['{{ service }}'] # service to be monitored
-      {% endfor %}
+      - targets:
+        {% for ip in grains['monitored_services'] %}
+        - "{{ ip }}:8001" # 8801: hanadb exporter port
```
arbulu89: I don't like having fixed ports. Actually, the hana instance deployed in the second VM is not using this port. 

More than that, the name wouldn't make any sense anymore. We are refering to machines here, not services...

I think that we need to consider many things here

libvirt/doc/monitoring.md
```diff
@@ -8,11 +8,11 @@ The terraform module follows the same conventions as other modules
 
 * mandatory:
 
-`monitored_services` this is a list containing the services to be monitored. Format: `HOST_IP:PORT`. Under the hood this var tell prometheus the IP and port where to scrape.
+`monitored_services` this is a list containing the IP addresses of hosts to be monitored. Under the hood this var tell prometheus the IP where to scrape.
 
 See tfvars.example
 ```
-monitored_services = ["192.168.110.X:8001", "192.168.110.X+1:8001", "192.168.110.X:9100", "192.168.110.X+1:9100"]
+monitored_services = ["192.168.110.X", "192.168.110.Y"]
```
arbulu89: Why are you changing this? This forces us to use static ports.
I don't know if this is the best solution

libvirt/doc/monitoring.md
```diff
@@ -43,6 +43,33 @@ hana:
...
+    device: /dev/watchdog
+  sbd:
+    device: '/dev/vdc'
+  ntp: pool.ntp.org
+  sshkeys:
+    overwrite: true
+    password: linux
+  resource_agents:
+    - SAPHanaSR
+  ha_exporter:
```
arbulu89: I know that the name come from the salt bootstrap formula, but is it not a bit redundant? I mean, we could call it `exporter` directly.
Finally, maybe only adding the exporter part example would be enough, but it's OK anyway

.gitignore
```diff
@@ -1,4 +1,5 @@
 **/.terraform
+**/terraform.tfstate.d
```
arbulu89: Could be just use something like `**/terraform.tfstate**` ?

libvirt/README.md
```diff
@@ -155,6 +155,7 @@ data.
 - **iscsi_image**: Source image of the machine hosting the iscsi target (sles15 or above) (only used if `iscsi` is used as a shared storage for fencing)
 - **monitoring_srv_ip**: IP address of the machine that will host the monitoring stack
 - **ha_sap_deployment_repo**: Repository with HA and Salt formula packages. The latest RPM packages can be found at [https://download.opensuse.org/repositories/network:/ha-clustering:/Factory/{YOUR OS VERSION}](https://download.opensuse.org/repositories/network:/ha-clustering:/Factory/)
+- **ha_pkgs_from_factory**: Whether or not to install HA/SAP packages from ha_sap_deployment_repo
```
arbulu89: Why exactly we need this new variable? I understand what is doing, but I don't see any need for that.

To run a development version? If this is the case, maybe I would choose other naming, like devel_mode.

MalloZup: I was asking me same question somehow the pkg are avail' I think perhaps devel mode make sense or the question is why the pkg aren't avail

libvirt/doc/monitoring.md
MalloZup: We had this in the past. I prefer also the port fine grained level of control.

salt/monitoring/prometheus/prometheus.yml
MalloZup: I think for moment we can keep the old part of service with Port number.

.gitignore
ayoub-belarbi: you mean `**/terraform.tfstate*` and not `**/terraform.tfstate**` right ? because `**/terraform.tfstate**` won't work

libvirt/doc/monitoring.md
ayoub-belarbi: it's a personal point of view here and I don't think using ha prefix problematic, it's only for clarification. We need to change it in the formula before anyway.

libvirt/doc/monitoring.md
ayoub-belarbi: The best option would be figuring out the port number from the pillar file we will be in a separate PR.

salt/monitoring/prometheus/prometheus.yml
ayoub-belarbi: I think for now, it's best if we enable the monitoring in both nodes by default and avoid specifying the IP of the hosts.

salt/monitoring/prometheus/prometheus.yml
```diff
@@ -1,15 +1,34 @@
...
+        {% endfor %}
+
+  - job_name: node
```
MalloZup: use `hana node`

salt/monitoring/prometheus/prometheus.yml
MalloZup: if we add this, we should change `monitored_services` with `monitored_ips`

I think in future this solution is really less flexible, because imagine we will have 4 ips,

for all for IPS we will monitor the same serivces.

Before it was more fine grained. I think we have more duplicata but more control which imho better..

MalloZup: btw there is a typo on the comment .. :grin:  `8801 != 8001 port`

libvirt/modules/monitoring/variables.tf
```diff
@@ -118,7 +118,7 @@ variable "pool" {
 }
 
 
-variable "monitored_services" {
+variable "monitored_hosts" {
   description = "HOST:PORT of service you want to monitor, it can contain same host with different ports number (diff services)"
```
arbulu89: We would need to update this doc

libvirt/terraform.tfvars.example
```diff
@@ -36,8 +41,8 @@ ha_sap_deployment_repo = ""
 #background = true
 
 # Monitoring:
-# add here HOST:PORT. In this example we monitor on same hosts different services
-monitored_services = ["192.168.110.X:8001", "192.168.110.X+1:8001", "192.168.110.X:9100", "192.168.110.X+1:9100"]
+# Add here IP addresses of the hosts to be monitored:
```
arbulu89: The last colom maybe it's not needed here

pillar_examples/automatic/cluster.sls
```diff
@@ -26,6 +26,10 @@ cluster:
   {% endif %}
   resource_agents:
     - SAPHanaSR
+  {% if grains.get('devel_mode') %}
```
arbulu89: As a temporary solution it might work, but eventually we will need to update this

libvirt/variables.tf
```diff
@@ -97,6 +97,11 @@ variable "ha_sap_deployment_repo" {
   type        = "string"
 }
 
+variable "devel_mode" {
```
arbulu89: And now we have the `devel_mode` and `qa_mode` variables.
Could we just have `mode` or something similar.
Are they auto exclusive?

salt/hana_node/cluster_packages.sls
```diff
@@ -1,3 +1,12 @@
+{% if grains.get('devel_mode') %}
```
arbulu89: Couldn't we just add this operation in the `pre_installation` states?
I mean, they do the same...
Just add the if condition there and depending the path set one priority or other

salt/pre_installation/ha_repos.sls
```diff
@@ -4,6 +4,7 @@ ha-factory-repo:
     - baseurl: {{ grains['ha_sap_deployment_repo'] }}
     - gpgautoimport: True
     - priority: 110
```
arbulu89: As mentioned in other comment, wouldn't be better to manage the priority (and the new `devel_mode` variable) here?

libvirt/variables.tf
MalloZup: I think that `devel` and `qa` mode are incompatible.. :(

libvirt/doc/monitoring.md
MalloZup: I personally kind to like `ha_exporter`, It might be also to generic somehow.. we can improve later

libvirt/doc/monitoring.md
MalloZup: I'm not sure if the best option is the pillar one. To me the pillar one tangled us to the pillar, so it create a dependency where when we specify IP:PORT we dont' have one

salt/pre_installation/ha_repos.sls
ayoub-belarbi: that caused some problems during the sap pattern installation if I recall correctly since it happens before the HA one.

salt/hana_node/cluster_packages.sls
ayoub-belarbi: that caused some problems during the sap pattern installation if I recall correctly since it happens before the HA one.

libvirt/variables.tf
ayoub-belarbi: yes now we have both, but I'm not sure about the qa mode scope to be honest.

libvirt/terraform.tfvars.example
ayoub-belarbi: what do you mean exactly?

libvirt/modules/monitoring/variables.tf
ayoub-belarbi: Nice catch!

libvirt/terraform.tfvars.example
arbulu89: Just that we might remove the last `:`. Never mind... XD

libvirt/variables.tf
arbulu89: Ok, this is something we will need to manage... Maybe unify them to a single variable.

We can work on it in a next PR

salt/hana_node/cluster_packages.sls
arbulu89: mmm, Ok, it can be.
Anyway, the `devel_mode` is kind of strange in that case as it only affects to a single package.

MalloZup:
@arbulu89 
@ayoub-belarbi 

please open issue in github and reference to this PR so we can remember and keep track of them.

PR #137, opened 2019-08-23
Make monitoring module indipendent from host

### LINKED ISSUE #123: research how to remove hcl weirdness [later]
# description:
This issue is a follow-up.

We should perhaps research how to improve hcl syntax. I will try to search if I can do it, but I have quite already digged and the current syntax is the ugly cleanest one :grin

so it can be tricky to find a better one

# details

What would happen here if we have 2 disk in the additional_disk list?
The solution looks quite strange (maybe there is not any better).
Anyway, we could improve it in the future as well, I don't really mind if it works

_Originally posted by @arbulu89 in https://github.com/SUSE/ha-sap-terraform-deployments/pull/116_

MalloZup:
# Checklist:

- [x] This pr was tested by a deployment. Everything OK. 

# Info:

This PR is part of the RFC i wrote. (https://confluence.suse.com/pages/viewpage.action?pageId=302186655)

This PR will make the monitoring module indipendent from HANA/and host disks, so we don't have any dependency from other modules.

I will test this shortly and removing the WIP. Should work already but I will deploy for safety
fix #123

### DIFF
--- a/libvirt/main.tf
+++ b/libvirt/main.tf
@@ -49,11 +49,11 @@ module "hana_node" {
   hana_count             = 2
   vcpu                   = 4
   memory                 = 32678
+  host_ips               = var.host_ips
   hana_inst_folder       = var.hana_inst_folder
   sap_inst_media         = var.sap_inst_media
   hana_disk_size         = "68719476736"
   hana_fstype            = var.hana_fstype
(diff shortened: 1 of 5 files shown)


MalloZup:
works merging thx

PR #138, opened 2019-08-23
 Remove SBD module and move it to hana_module for clarity

MalloZup:
PART OF RFC  : https://confluence.suse.com/pages/viewpage.action?pageId=302186655

# what does it?

- move SBD disk creation to hana_node and remove module tricky mechanism.
- we have less variable propagation.

## tested?
- [x] deployed to gollum OK

PR #139, opened 2019-08-26
remove host-module concept and unify it to hana module (from 2 to 1 module)

MalloZup:
# Desc:
This PR remove the host module, follow-up from RFC.

# status:

this is builded on top of pr #138

still in wip for moment.


MalloZup:
feel free to review. So the sbd pb is fixed

I have already tested it, I will retest it few times otherwise we can merge it

PR #143, opened 2019-08-29
azure monitoring (non module)

MalloZup:
## TODOS:

- [x] fix sapahana pattern install @arbulu89
- [x] the monitoring instance in azure cannot contact the hana node via network fix this @MalloZup

### DIFF
--- a/azure/README.md
+++ b/azure/README.md
@@ -188,7 +188,6 @@ In the file [terraform.tfvars.example](terraform.tfvars.example) there are a num
 * **instancetype**: SKU to use for the cluster nodes; basically the "size" (number of vCPUS and memory) of the VM.
 * **ninstances**: number of cluster nodes to deploy. Defaults to 2.
 * **az_region**: Azure region where to deploy the configuration.
-* **init-type**: initilization script parameter that controls what is deployed in the cluster nodes. Valid values are `all` (installs Hana and configures cluster), `skip-hana` (does not install Hana, but configures cluster) and `skip-cluster` (installs hana, but does not configure cluster). Defaults to `all`.
 * **cluster_ssh_pub**: SSH public key name (must match with the key copied in sshkeys folder)
 * **cluster_ssh_key**: SSH private key name (must match with the key copied in sshkeys folder)
 * **ha_sap_deployment_repo**: Repository with HA and Salt formula packages. The latest RPM packages can be found at [https://download.opensuse.org/repositories/network:/ha-clustering:/Factory/{YOUR OS VERSION}](https://download.opensuse.org/repositories/network:/ha-clustering:/Factory/)
--- a/azure/README.md
+++ b/azure/README.md
@@ -1,5 +1,16 @@
-
 # Azure Public Cloud deployment with terraform and Salt
+
+- [quickstart](#quickstart)
+- [highlevel description](#highlevel-description)
+- [advanced usage](#advanced-usage)
+- [specification](#specification)
+
+# Quickstart
+
+* Install [azure commandline](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-zypper?view=azure-cli-latest)
+
+
+# Highlevel description
 
 The terraform configuration files in this directory can be used to create the infrastructure required to install a SAP HanaSR cluster on Suse Linux Enterprise Server for SAP Applications in **Azure**.
 
@@ -21,9 +32,7 @@ Once the infrastructure is created by Terraform, the servers are provisioned wit
  The cluster and HANA installation is done using Salt Formulas.
  To customize this provisioning, you have to create the pillar files (cluster.sls and hana.sls) according to the examples in the [pillar_examples](https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/pillar_examples) folder (more information in the dedicated [README](https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/pillar_examples/README.md))
 
-
-# Please enter the commit message for your changes. Lines starting
-## Relevant files
+## Specification
 
 These are the relevant files and what each provides:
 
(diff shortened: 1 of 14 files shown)


MalloZup:
this PR isn't perfect but I prefer to split up the work and doing smaller prs and make it easier to review.
I will add and improve this later. 

I would not do this if it would be a DESIGN change , but this pr isn't changing the design so we can adopt this strategy to make a not big pr and easy review.

If I can describe this PR:

-  it add documentation, 
- it fix some issues
- it add the monitoring instance in experimental way, I will perfection this later.

azure/README.md
```diff
@@ -1,6 +1,74 @@
...
+- [specification](#specification)
+
+# Quickstart
+
+## 1) Install the azure client 
+
+* [azure commandline](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-zypper?view=azure-cli-latest)
+
+
+### 2) Configuration of terraform:
```
arbulu89: I would remove the `:` of the end

azure/README.md
```diff
@@ -1,6 +1,74 @@
...
+
+1) Rename terraform.tfvars `mv terraform.tfvars.example terraform.tfvars`
+
+2) Generate private and public keys for the cluster nodes with:
+
+```
+mkdir ../salt/hana_node/files/sshkeys; ssh-keygen -t rsa -f ../salt/hana_node/files/sshkeys/cluster.id_rsa
+```
+The key files need to be named as you defined it in `terraform.tfvars` file.
```
arbulu89: Maybe better as:
```The key files must be named as you define them in the `terraform.tfvars` file```

File: azure/README.md [hunk 3]
```diff
@@ -83,7 +136,6 @@ In the file [terraform.tfvars.example](terraform.tfvars.example) there are a num
 * **instancetype**: SKU to use for the cluster nodes; basically the "size" (number of vCPUS and memory) of the VM.
 * **ninstances**: number of cluster nodes to deploy. Defaults to 2.
 * **az_region**: Azure region where to deploy the configuration.
-* **init-type**: initilization script parameter that controls what is deployed in the cluster nodes. Valid values are `all` (installs Hana and configures cluster), `skip-hana` (does not install Hana, but configures cluster) and `skip-cluster` (installs hana, but does not configure cluster). Defaults to `all`.
```
arbulu89: Why have you removed that? Is this variable not existing any longer?

azure/image.tf
```diff
@@ -9,31 +9,49 @@ variable "sles4sap_uri" {
...
-variable "sles4sap_public" {
-  type = map(string)
-
-  default = {
-    "publisher" = "SUSE"
-    "offer"     = "SLES-SAP-BYOS"
-    "sku"       = "12-sp4"
-    "version"   = "2019.03.06"
-  }
+variable "hana_public_publisher" {
```
arbulu89: Honestly, I preferred the `map` kind of data for this variable. But I don't know how it behaves with terraform 12.
Anyway, It's just a comment

azure/image.tf
```diff
@@ -9,31 +9,49 @@ variable "sles4sap_uri" {
...
-    "offer"     = "SLES-SAP-BYOS"
-    "sku"       = "15"
-    "version"   = "2018.08.20"
-  }
+variable "hana_public_version" {
+  type    = string
+  default = "2019.03.06"
+}
+
+variable "iscsi_publisher" {
```
arbulu89: shouldn't be `iscsi_public_publisher?

azure/image.tf
```diff
@@ -9,31 +9,49 @@ variable "sles4sap_uri" {
   default = ""
```
arbulu89: I guess we would need to update the `sles4sap_uri` to `hana_image_uri` or similar. What do you think?

azure/monitoring.tf
```diff
@@ -0,0 +1,29 @@
+variable "monitoring_count" {
+  description = "number of hosts like this one"
```
arbulu89: Could be improve the description?
`Number of hosts with the monitoring stack installed` or similar maybe?

azure/terraform.tfvars.example
```diff
@@ -55,9 +55,6 @@ hana_inst_folder = "/root/sap_inst/"
 # Device used by node where HANA will be installed
 hana_disk_device = "/dev/sdc"
 
-# Variable for init-nodes.sh script
-init-type = "all"
```
arbulu89: I don't understand why you are removing this variable yet. As far as I know the QA team was using it. Have you talked with them?

salt/pre_installation/ha_repos.sls
```diff
@@ -8,3 +8,13 @@ ha-factory-repo:
     - retry:
```
arbulu89: I think we could improve this state file, just changing the `priority` number depending on if `devel_mode` is set or not. I mean, having only one state, exactly as it was before, and depending on the `devel_mode` use one priority or another

azure/image.tf
MalloZup: I prefer also the map, but I think terraform has an issue with this. TF12 :( . I will open an issue as follow-up if we can investiagate on this

MalloZup: issue opened as follow-up  https://github.com/SUSE/ha-sap-terraform-deployments/issues/157

azure/image.tf
MalloZup: ah good catch yes

azure/terraform.tfvars.example
MalloZup: I will preserve it. 
thx my cleaning syndrmome :)

salt/pre_installation/ha_repos.sls
MalloZup: good thing yes make sense

azure/image.tf
MalloZup: I think the rename we can do later. I had look it is quite tangled everywhere..
Perhaps is good name the sles4sap_uri for moment. (e.g we might reuse this image for other things)

MalloZup:
@xabi good point I will need to test this (using the sles4sap_uri) etc.

azure/image.tf
arbulu89: thanks!

MalloZup:
I'm testing currenlty with slessap_uri just in case

MalloZup:
thx! I'm openining follow-up issue for azure so we don't forget about things. I can't fix all the stuff

MalloZup:
https://github.com/SUSE/ha-sap-terraform-deployments/issues/159

MalloZup:
I'm merging this. We will need to figure out how to deal with hawk-api-server

MalloZup:
test with custom image worked

PR #153, opened 2019-09-03
FIX: inject the develmode grain in salt-provisioner

MalloZup:
# current behaviour:

currently the variable develmode is set but is not injected via a grain (salt-provisioner.tf)

This pr fix this.

Also, I think that we don't need a global variable. We just make a variable belong to `hana_node` module and we set it to `main.tf`. 

In future we will need to remove this  but for moment let's get it working :)


MalloZup:
I"m trying to test this

MalloZup:
I think before expanding it to other platform we  might think to remove it the variable.

I'm hardcoding to `main.tf` becaause basically if you want to use `ha-monitoring` you need to set to true and most of the time you will need to use to true for having it deployed.

MalloZup:
I'm ok to just inject the grain for this PR. please take look thx

PR #155, opened 2019-09-03
remove additional_repos variable

MalloZup:
we never used this variable and it can be tricky to manage.
An user can add always later additional repos after the deployment

I will test this PR just in case of but it should be already ok


MalloZup:
@Simranpal yep I know. It is a redundant option. I know historically this was present in sumaform but even there this feature was more a `nice to have` secondary feature, which was never used.

We should remove it for this, preserving our minimalism.

If an user want to add repo later, he can do his scripts. Our goal with this project is to deploy our solution not to give a way for doing later on things.

For this  I will remove this grain

MalloZup:
tested works fine

MalloZup:
I agree. I take responsibility on this PR. If it will be needed I will revert it.

THere are also other zoombie feature from sumaform which I will adress the removal once I have time.

Thx

PR #164, opened 2019-09-09
fix devel_mode for deployment

MalloZup:
# what does this PR:

we had issues with devel_mode espcially with azure. but was a coincidence.

This pr will allow to safe update pkgs.

When we update the system in pre-install phase, this fix issues we have
encountered in Azure cloud, where the image had already some pkgs
installed like hawk-api-server.

If we now do a allow-vendor changes and zypper up in devel mode, if
there are already this pkgs we will update them, and fix the issues.

If the pkgs aren't installed, we will install them to latest since we allow the vendor changes.

### DIFF
--- a/azure/instances.tf
+++ b/azure/instances.tf
@@ -3,10 +3,11 @@
 # Availability set for the VMs
 
 resource "azurerm_availability_set" "myas" {
-  name                = "myas"
-  location            = var.az_region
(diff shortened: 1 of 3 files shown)


MalloZup:
Ok I have tested with libvirt as well. works! thx merging

PR #167, opened 2019-09-11
Fix sshkey documentation for AWS and GCP

juadk:
As the provision symlink was removed, we need to update documentations about where to generate sshkeys.


MalloZup:
Great catch

PR #173, opened 2019-09-16
Add option to set disk type and caching for HANA

juadk:
When I was doing benchmark tests in Azure, I noticed that we don't use SSD as DATA disks in Azure like we are doing in other cloud providers.

This PR changes:

- Default DATA disks from `Standard_LRS` to `StandardSSD_LRS`.
Disk type could be set with `hana_data_disk_type` variable.

- New variable `hana_data_disk_caching` added to manage data disk caching.
Could be None, ReadOnly or ReadWrite (ReadWrite by default).

### DIFF
--- a/azure/README.md
+++ b/azure/README.md
@@ -134,6 +134,8 @@ In the file [terraform.tfvars.example](terraform.tfvars.example) there are a num
 * **hana_inst_master**: path to the storage account where SAP HANA installation files are stored.
 * **hana_fstype**: filesystem type used for HANA installation (xfs by default).
 * **instancetype**: SKU to use for the cluster nodes; basically the "size" (number of vCPUS and memory) of the VM.
+* **hana_data_disk_type**: disk type to use for HANA (Standard_LRS by default).
+* **hana_data_disk_caching**: caching mode for HANA disk, could be None, ReadOnly or ReadWrite (ReadWrite by default).
 * **ninstances**: number of cluster nodes to deploy. Defaults to 2.
 * **az_region**: Azure region where to deploy the configuration.
 * **init-type**: initilization script parameter that controls what is deployed in the cluster nodes. Valid values are `all` (installs Hana and configures cluster), `skip-hana` (does not install Hana, but configures cluster) and `skip-cluster` (installs hana, but does not configure cluster). Defaults to `all`.
--- a/azure/variables.tf
+++ b/azure/variables.tf
@@ -20,6 +20,16 @@ variable "ninstances" {
   default = "2"
 }
 
+variable "hana_data_disk_type" {
+  type    = string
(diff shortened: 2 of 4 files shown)


azure/variables.tf
```diff
@@ -20,6 +20,15 @@ variable "ninstances" {
   default = "2"
 }
 
+variable "hana_data_disk_type" {
+  type    = string
+}
```
ldevulder: No default type for this?

ldevulder:
> THX I'm not sure we need a variable, I mean in sake of minimalism perhaps we can just hardcode the values.

Hardcore doesn't let choice for customers, some people may want to use more cheaper HDD maybe.

> I think that adding more variable we will loose control.

I don't agree with this, at least I agree that thousand of variables is not so good, but we don't have so much here and this one is clearly important IMHO. And we should add the same for the other providers I think (to be consistent).

juadk:
> retrospectively even if I contradict myself, I think it make more sense to just hardcode this values Because having variable add complexity and different behaviour from provider to provider.
> 
> In theory what we want in future would be to keep the difference between providers really low

I understand your point of view but maybe this variable is more important than the others because it has a cost impact for the customer.

azure/README.md
```diff
@@ -134,6 +134,8 @@ In the file [terraform.tfvars.example](terraform.tfvars.example) there are a num
 * **hana_inst_master**: path to the storage account where SAP HANA installation files are stored.
 * **hana_fstype**: filesystem type used for HANA installation (xfs by default).
 * **instancetype**: SKU to use for the cluster nodes; basically the "size" (number of vCPUS and memory) of the VM.
+* **hana_data_disk_type**: disk type to use for HANA.
```
MalloZup: add default type

azure/variables.tf
MalloZup: yep

juadk: I hesitated for this one, I would not set SSD for default and surprise customer with extra costs.
As we specify SSD in terraform.tfvars.example file, we could set standard HDD as default.

MalloZup:
OK I can see the value of this variable but just add to all providers right now, so we don't forget it. thx!

azure/variables.tf
MalloZup: `hdd` default imho is ok and make sense yep

azure/README.md
juadk: done thx

azure/variables.tf
juadk: done thx

juadk:
> feel free either to merge this and do follow-up PRs or just do everything in 1 pr as you wish
I will do another one PR to standardize disks variables usage.

PR #174, opened 2019-09-16
WIP: Add alerts for ha-metrics

MalloZup:
# Desc:

This PR rely on the new ha metrics exporter defined here: 

`https://github.com/MalloZup/ha_cluster_exporter`
This export cluster metrics. From this we create alerts and dashboards.

- Add a 1st useful alert when, in our case when the master resource is down we will have an alert.  This is 

- Add needed configuration for enabling alerts in prometheus.

# done 

- This alert rely on the new `ha_metrics_exporter` which we need pkg first

### DIFF
--- a/salt/monitoring/prometheus/prometheus.yml
+++ b/salt/monitoring/prometheus/prometheus.yml
@@ -36,6 +36,6 @@ scrape_configs:
     static_configs:
       - targets:
         {% for ip in grains['monitored_hosts'] %}
-        - "{{ ip }}:9001" # 9001: hawk-apiserver exporter port
+        - "{{ ip }}:9002" # 9002: ha_cluster_exporter metrics
         {% endfor %}
 {% endif %}
--- a/salt/monitoring/prometheus/prometheus.yml
+++ b/salt/monitoring/prometheus/prometheus.yml
@@ -4,6 +4,15 @@ global:
   scrape_timeout: 5s
   evaluation_interval: 5s
 
+alerting:
+  alertmanagers:
+  - static_configs:
+    - targets:
+      - localhost:9093
+
+rule_files:
+    - /etc/prometheus/rules.yml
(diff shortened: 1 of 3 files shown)


salt/monitoring/prometheus/rules.yml
```diff
@@ -1,13 +1,9 @@
...
-
-# this is only a boostrap/initial alert to setup things
-- name: hana-node01-network-latency
+- name: primary-resource-saphana-down
   rules:
-  - alert: HighRequestLatency
-    expr: job:request_latency_seconds:mean5m{job=hanadb"} > 0.5
-    for: 10m
+  - alert: SAPHANAMasterResourcesDown
+    expr: absent(cluster_node_resources{resource_name="rsc_saphana_prd_hdb00",role="master"})
```
MalloZup: we might add here like regex I didn't investigate

salt/monitoring/prometheus/prometheus.yml
```diff
@@ -20,10 +29,8 @@ scrape_configs:
         - "{{ ip }}:9100" # 9100: node exporter port
         {% endfor %}
 
-
   - job_name: cluster
     metrics_path: /metrics
-    scheme: https
     tls_config:
       insecure_skip_verify: true
```
ayoub-belarbi: Since we are not using https (so no self signed certificates) no need to have the option of tls_config.

salt/monitoring/prometheus/rules.yml
ayoub-belarbi: The resource name here is hard coded... I don't think this will work ....

salt/monitoring/prometheus/prometheus.yml
MalloZup: ah good catch

PR #175, opened 2019-09-16
Add hwcct benchmark

juadk:
This PR adds a possibility to execute HWCCT (Hardware Configuration Check Tool) in public cloud providers.
It's mainly a new feature for QA testing. The benchmark usually lasts about 2 hours with SSD drive.
The purpose is to execute it when we are doing build validation.
Results will be both in /root/hwcct_out and in the global log file /tmp/provisioning.log

Tested with an without hwcct set to true:
- [x] AWS 
- [x] GCP
- [x] Azure

### DIFF
--- a/aws/README.md
+++ b/aws/README.md
@@ -157,8 +157,9 @@ In [terraform.tfvars](terraform.tfvars.example) there are a number of variables 
 * **additional_packages**: Additional packages to add to the guest machines.
 * **hosts_ips**: Each cluster nodes IP address (sequential order). Mandatory to have a generic `/etc/hosts` file.
 
- Specific QA variable
+ Specific QA variables
 * **qa_mode**: If set to true, it disables extra packages not already present in the image. For example, set this value to true if performing the validation of a new AWS Public Cloud image.
+* **hwcct**: If set to true, it executes HANA Hardware Configuration Check Tool to bench filesystems. It's a very long test (about 2 hours), results will be both in /root/hwcct_out and in the global log file /tmp/provisioning.log.
 
 
 ### Relevant Details
--- a/gcp/terraform.tfvars.example
+++ b/gcp/terraform.tfvars.example
@@ -86,3 +86,8 @@ ha_sap_deployment_repo = ""
 # Except salt-minion (for the moment) and salt formulas
 # true or false (default)
 #qa_mode = false
+
+# Execute HANA Hardware Configuration Check Tool to bench filesystems
+# qa_mode must be set to true for executing hwcct
+# true or false (default)
+#hwcct = false
--- a/salt/qa_mode/files/hwcct/hwcct_bench.jinja
+++ b/salt/qa_mode/files/hwcct/hwcct_bench.jinja
@@ -1,0 +1,16 @@
+{% set data = pillar.cluster.configure.template.parameters %}
+{% set sid = data.sid.upper() %}
+{% set instance = '{:0>2}'.format(data.instance) %}
+
+#!/bin/sh
+
+# Extract and execute HWCCT
+
+typeset -r HWCCTCONFIG=/root/salt/qa_mode/files/hwcct/hwcct_config.json
+
+# Extracting HWCCT tool
+/usr/sap/{{ sid }}/HDB{{ instance }}/exe/SAPCAR -xf /root/sap_inst/DATA_UNITS/SAP_HANA_HWCCT_LINUX_X86_64/HWCCT.SAR
+
+# Executing HWCCT tool
+cd hardwareConfigurationCheckTool
+source ./envprofile.sh && ./hwcct -f ${HWCCTCONFIG}
--- a/salt/qa_mode/init.sls
+++ b/salt/qa_mode/init.sls
@@ -1,0 +1,23 @@
+{% if grains.get('hwcct') == true and 'hana01' in grains['hostname'] %}
+
+hwcct-config-file:
+ file.managed:
+   - template: jinja
+   - names:
+     - /root/salt/qa_mode/files/hwcct/hwcct_config.json:
+       - source: salt://qa_mode/files/hwcct/hwcct_config.json.jinja
+
+hwcct-bench-file:
+ file.managed:
+   - template: jinja
+   - names:
+     - /root/salt/qa_mode/files/hwcct/hwcct_bench.sh:
+       - source: salt://qa_mode/files/hwcct/hwcct_bench.jinja
+
+hwcct:
+  cmd.run:
+    - name: sh /root/salt/qa_mode/files/hwcct/hwcct_bench.sh
+    - require:
+      - file: hwcct-config-file
+      - file: hwcct-bench-file
+{% endif %}
--- a/salt/qa_mode/run_qa_mode.sh
+++ b/salt/qa_mode/run_qa_mode.sh
@@ -1,0 +1,13 @@
+#!/bin/bash -xe
+
+# We need to export HOST with the new hostname set by Salt
+# Otherwise, hwcct will error out.
+export HOST=$(hostname)
+
+# Execute qa state file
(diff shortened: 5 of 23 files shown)


salt/qa/qa.sls
```diff
@@ -0,0 +1,13 @@
...
+hwcct-config-file:
+ file.managed:
+   - template: jinja
+   - names:
+     - /root/salt/qa/files/hwcct/hwcct_config.json:
+       - source: salt://files/hwcct/hwcct_config.json.jinja
+
+hwcct:
+  cmd.run:
+    - name: sh /root/salt/qa/files/hwcct/hwcct_bench.sh
```
MalloZup: shouldn't this cmd `require` `hwcct-config-file:`?

aws/README.md
```diff
@@ -157,8 +157,9 @@ In [terraform.tfvars](terraform.tfvars.example) there are a number of variables
 * **additional_packages**: Additional packages to add to the guest machines.
 * **hosts_ips**: Each cluster nodes IP address (sequential order). Mandatory to have a generic `/etc/hosts` file.
 
- Specific QA variable
+ Specific QA variables
 * **qa_mode**: If set to true, it disables extra packages not already present in the image. For example, set this value to true if performing the validation of a new AWS Public Cloud image.
+* **hwcct**: If set to true, it executes HANA Hardware Configuration Check Tool to bench filesystems. It's a very long test (about 2 hours), results will be both in /root/hwcct_out and in the global log file /tmp/provisioning.log.
```
MalloZup: is there any upstream link to this tool? is this opensource? would be helpful to have such link

salt/qa/qa.sls
juadk: > shouldn't this cmd `require` `hwcct-config-file:`? via salt construct?

Oh yes, good catch, thx

gcp/terraform.tfvars.example
```diff
@@ -86,3 +86,7 @@ ha_sap_deployment_repo = ""
 # Except salt-minion (for the moment) and salt formulas
 # true or false (default)
 #qa_mode = false
+
```
MalloZup: for consistency you need to add this to `libvirt`

aws/README.md
juadk: AFAIK, this is not opensource.
I have this [link](https://help.sap.com/viewer/6b94445c94ae495c83a19646e7c3fd56/2.0.00/en-US/44226b888ae64f40a6e7906eeaa0a44c.html).

Other resources exist but you need to have SAP identifiers (which I do not have).

salt/qa/qa.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -xe
```
MalloZup: rename the file from `qa.sh`  to `run_qa_mode.sh` something more usefull

salt/qa/qa.sls
```diff
@@ -0,0 +1,15 @@
+{% if grains['hwcct'] == true and 'hana01' in grains['hostname'] %}
+
```
MalloZup: - rename file from `qa.sls` to `init.sls` to be inline with others.
- rename folder qa to `qa_mode`. Imho make more sense

salt/qa/files/hwcct/hwcct_bench.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# Small bash script to extract and execute HWCCT
+
+typeset -r HWCCTCONFIG=/root/salt/qa/files/hwcct/hwcct_config.json
+
+# Extracting HWCCT tool
+/usr/sap/PRD/HDB00/exe/SAPCAR -xf /root/sap_inst/DATA_UNITS/SAP_HANA_HWCCT_LINUX_X86_64/HWCCT.SAR
```
arbulu89: Could be change this to not use a predefined sid and instance numbers?
Otherwise, It will only work for PRD and 00.
I guess we can have a jinja template for that quite easily

salt/qa/qa.sh
arbulu89: I agree. I would put other more meaningful name.

salt/qa/files/hwcct/hwcct_bench.sh
juadk: > Could be change this to not use a predefined sid and instance numbers?
> Otherwise, It will only work for PRD and 00.
> I guess we can have a jinja template for that quite easily

To be honest, as I am a beginner in SAP I did not pay attention to that... good point thx

salt/qa/qa.sh
juadk: Done thx

salt/qa/qa.sls
juadk: Done thx

salt/qa/files/hwcct/hwcct_bench.sh
juadk: jinja template done

gcp/terraform.tfvars.example
juadk: > for consistency you need to add this to `libvirt`

I'm adding the code to libvirt for consistency but I hope you understand that you need a special image to use `qa_mode`.

For example you can not use a JeOS image because `qa_mode` disables all registration process.
`qa_mode` allows us to be sure that we test packages from the image. 
As a reminder, when we are doing a build validation, up-to-date repos are not available yet in SCC. 
That's why we must have all the necessary packages in the image.

`hwcct` is called by `run_qa_mode.sh` and this one is called if `qa_mode: true`

Does it make sense to execute `hwcct` without `qa_mode` set to true?
As it's a request from our QA team, it doesn't make sense to me.

WDYT? @arbulu89 @MalloZup

arbulu89: No, it doesn't in my opinion

salt/qa_mode/init.sls
```diff
@@ -0,0 +1,23 @@
+{% if grains.get('hwcct') == true and 'hana01' in grains['hostname'] %}
+
+hwcct-config-file:
+ file.managed:
+   - template: jinja
+   - names:
+     - /root/salt/qa_mode/files/hwcct/hwcct_config.json:
+       - source: salt://qa_mode/files/hwcct/hwcct_config.json.jinja
+
+hwcct-bench-file:
```
MalloZup: to keep in mind : we will need to find in future a standard naming.

We have states like `f00-bar` and `foo_bar`. Imho the latter is better. anyways unrelated for momemnt

salt/qa_mode/run_qa_mode.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -xe
+
+# We need to export HOST with the new hostname set by Salt
+# Otherwise, hwcct will complain about it
```
MalloZup: @juadk  `complain` isn't software is more IRC chat but not comment :grin: .

Write hwcct will error out or require it by default something more conventional

salt/qa_mode/files/hwcct/hwcct_bench.jinja
```diff
@@ -0,0 +1,16 @@
+{% set data = pillar.cluster.configure.template.parameters %}
+{% set sid = data.sid.upper() %}
+{% set instance = '{:0>2}'.format(data.instance) %}
+
+#!/bin/sh
+
+# Small bash script to extract and execute HWCCT
```
MalloZup: remove `small bash script` I think this is obvious

salt/qa_mode/run_qa_mode.sh
juadk: Done thx

salt/qa_mode/files/hwcct/hwcct_bench.jinja
juadk: Done thx

PR #176, opened 2019-09-16
Fix CLOUD_NETCONFIG for IPaddr2

juadk:
This PR fixes bsc#1144282 for AWS and Azure.

- [x] AWS
- [x] Azure

No impact for GCP and libvirt.


juadk:
> If you have tested an it works in all the providers I don't see any inconvenience

Yes it's done.

PR #178, opened 2019-09-17
Standardize data disk variables

juadk:
This PR standardizes data disk usage for HANA:
For all public cloud providers:
- same variable `hana_data_disk_type`
- README file
- tfvars.example file

### DIFF
--- a/gcp/disks.tf
+++ b/gcp/disks.tf
@@ -8,7 +8,7 @@ resource "google_compute_disk" "iscsi_data" {
 resource "google_compute_disk" "node_data" {
   count = "2"
   name  = "${terraform.workspace}-${var.name}-data-${count.index}"
-  type  = "pd-ssd"
+  type  = "${var.hana_data_disk_type}"
   size  = var.init_type == "all" ? 60 : 30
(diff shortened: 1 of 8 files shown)


gcp/disks.tf
```diff
@@ -8,15 +8,15 @@ resource "google_compute_disk" "iscsi_data" {
...
 }
 
 resource "google_compute_disk" "node_data2" {
   count = "2"
   name  = "${terraform.workspace}-${var.name}-backup-${count.index}"
-  type  = "pd-ssd"
+  type  = "${var.hana_data_disk_type}"
   size  = "20"
   zone  = element(data.google_compute_zones.available.names, count.index)
 }
```
juadk: I have a doubt about this 20 GB disk, I don't know if we still need it anymore.

juadk:
> lgtm thx . DO we have this in libvirt?

No, if you want to do the same thing in libvirt, I believe you can play with pool and have for example a pool with HDD and another one with SSD.

MalloZup:
ah yeah right thx for clarification I think is fine

:+1:

PR #179, opened 2019-09-17
Fix HA dashboard to match ha_cluster_exporter

ayoub-belarbi:


ayoub-belarbi:
> is this the actual one? I guess so just tell me if it the last one and we can merged
it need to be adjusted again since we changed the format again.

ayoub-belarbi:
@MalloZup Done from my side! If you want you can review it again, otherwise it's tested and works well!

PR #180, opened 2019-09-18
Implement some alerts for resources/clusters

MalloZup:
# desc

- add 2 new alerts for cluster
- remove service start exporter, this need to be done in formula

in combination with : https://github.com/SUSE/habootstrap-formula/pull/28/files


salt/monitoring/prometheus/rules.yml
```diff
@@ -7,3 +7,22 @@ groups:
       severity: page
     annotations:
       summary: Primary SAP-HANA resource down
+
+- name: a-resource-failed
+  rules:
+  - alert: aResourceFailed
+    expr: cluster_node_resources{status="failed"} > 1
```
ayoub-belarbi: Shouldn't be count(cluster_node_resources{status="failed"}), if it's not the case and no value has been retrieved, the return value will be null and I don't know what's the result of the comparison in that case. I had the same issue with Grafana

salt/monitoring/prometheus/rules.yml
```diff
@@ -21,7 +21,7 @@ groups:
 - name: secondary-resource-saphana-slave-role-changed
   rules:
   - alert: SAPHANAMasterResourcesDown
-    expr: absent(cluster_node_resources{node="dma-dog-hana02",resource_name="rsc_saphana_prd_hdb00",role="slave"})
+    expr: absent(cluster_node_resources{resource_name="rsc_saphana_prd_hdb00",role="slave"})
```
ayoub-belarbi: This is much better! shouldn't we add status=active here ?

PR #184, opened 2019-09-19
Adapt pillars to disable ha_exporter by default

### LINKED ISSUE #181: ha_exporter installed by default
Right now, the ha_exporter is installed by default, without considering if the monitoring stack is created or not.
In order to align all the component, the ha_exporter should be only installed together with the whole monitoring stack.
What do you think?
For that, we should just set this value to False in the cloud_providers (aws and gcp don't even have monitoring stack) and in the automatic pillar example depending on the `monitoring_enabled` parameter.

Examples:
https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/pillar_examples/aws/cluster.sls#L13
https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/pillar_examples/automatic/cluster.sls#L29

In fact, we should take advantage, and do the same with the hanadb_exporter (only enable according this `monitoring_enabled` parameter.

MalloZup:
For automatic pillar enable if monitoring_enabled

I'm testing this now

(@stefanotorresi  added in CC just as a turist for moment but feel free to look even :) )

fix #181

### DIFF
--- a/pillar_examples/automatic/cluster.sls
+++ b/pillar_examples/automatic/cluster.sls
@@ -26,7 +26,7 @@ cluster:
   {% endif %}
   resource_agents:
     - SAPHanaSR
-  {% if grains.get('monitoring_enabled') %}
+  {% if grains.get('monitoring_enabled', False) %}
   ha_exporter: true
   {% else %}
   ha_exporter: false
--- a/pillar_examples/automatic/cluster.sls
+++ b/pillar_examples/automatic/cluster.sls
@@ -26,8 +26,11 @@ cluster:
   {% endif %}
   resource_agents:
     - SAPHanaSR
+  {% if grains.get('monitoring_enabled') %}
   ha_exporter: true
-
+  {% else %}
+  ha_exporter: false
+  {% endif %}
   {% if grains['init_type']|default('all') != 'skip-hana' %}
   configure:
     method: update
--- a/pillar_examples/automatic/hana.sls
+++ b/pillar_examples/automatic/hana.sls
@@ -29,10 +29,12 @@ hana:
           user_name: SYSTEM
           user_password: YourPassword1234
           database: SYSTEMDB
+    {% if grains.get('monitoring_enabled', False) %}
       exporter:
         exposition_port: 8001
         user: SYSTEM
         password: YourPassword1234
+    {% endif %}
 
     - host: {{ grains['name_prefix'] }}02
       sid: prd
@@ -80,6 +82,7 @@ hana:
         {% endif %}
         system_user_password: YourPassword1234
         sapadm_password: YourPassword1234
+    {% if grains.get('monitoring_enabled', False) %}
       exporter:
         exposition_port: 8002
         user: SYSTEM
--- a/libvirt/terraform.tfvars.example
+++ b/libvirt/terraform.tfvars.example
@@ -43,8 +43,8 @@ devel_mode = false
 # Add here IP addresses of the hosts to be monitored:
 monitored_hosts = ["192.168.110.X", "192.168.110.Y"]
 
-# by default monitoring is enabled, if you want to disable it on your node set to false
-# monitoring_enabled = false
+# by default monitoring is disabled.
+monitoring_enabled = true
 
 # libvirt storage pool, select the libvirt storage pool where the volume will stored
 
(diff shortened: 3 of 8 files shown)


MalloZup:
tested with libvirt works fine

pillar_examples/automatic/cluster.sls
```diff
@@ -26,8 +26,11 @@ cluster:
   {% endif %}
   resource_agents:
     - SAPHanaSR
+  {% if grains.get('monitoring_enabled') %}
   ha_exporter: true
-
+  {% else %}
+  ha_exporter: false
+  {% endif %}
```
stefanotorresi: could this be shortened to `ha_exporter: {% grains.get('monitoring_enabled') %}` ?

arbulu89: Valid point! Yes, it is. If the key doesn't exist it will return `None`. This should work as in the `habootstrap-formula` we just check if this values is same as true to install or not the exporter.
Anyway,  we could just add `false` as default value to add more meaning to the parameter.

MalloZup: ok thx!

arbulu89:
@MalloZup Just one last comment.
Maybe we should change the default value of this parameter here too: [variables.tf](https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/libvirt/variables.tf#L122).

I would say that the monitoring stack is an additional feature that shouldn't be deployed by default.
To need to change the `terraform.tfvars` file and set the value to True makes you understand better what you are doing.
What do you think?

pillar_examples/automatic/cluster.sls
MalloZup: regarding this point,  think that if it that exporession would be shorter, is less readable in term of logic.

MalloZup:
done thx. Please remember to update then your main.tf and variables then. DOn't ping me if monitoring doesn't work ! since it will be disabled :grin:

ayoub-belarbi:
> @MalloZup Just one last comment.
> Maybe we should change the default value of this parameter here too: [variables.tf](https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/libvirt/variables.tf#L122).
> 
> I would say that the monitoring stack is an additional feature that shouldn't be deployed by default.
> To need to change the `terraform.tfvars` file and set the value to True makes you understand better what you are doing.
> What do you think?

I agree, and I said that from the beginning... the nice thing is if the customer don't need monitoring, it will save him time and resources especially on cloud providers, because spinning out another instance costs money. Other than this, LGTM!

MalloZup:
ok I'm testing it with libvirt thx. I will merge once tests are ok

pillar_examples/automatic/cluster.sls
stefanotorresi: Well, it wouldn't probably be called `monitoring_enabled` anymore then ;)

I personally find the shorter version more readable, but I understand this comes down to personal preference, so carry on. :)

juadk:
> > @MalloZup Just one last comment.
> > Maybe we should change the default value of this parameter here too: [variables.tf](https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/libvirt/variables.tf#L122).
> > I would say that the monitoring stack is an additional feature that shouldn't be deployed by default.
> > To need to change the `terraform.tfvars` file and set the value to True makes you understand better what you are doing.
> > What do you think?
> 
> I agree, and I said that from the beginning... the nice thing is if the customer don't need monitoring, it will save him time and resources especially on cloud providers, because spinning out another instance costs money. Other than this, LGTM!

Same for QA, It will save costs money for SUSE and time to deploy. Thx

pillar_examples/automatic/hana.sls
```diff
@@ -80,6 +82,7 @@ hana:
         {% endif %}
         system_user_password: YourPassword1234
         sapadm_password: YourPassword1234
+    {% if grains.get('monitoring_enabled', False) %}
```
juadk: A new if condition is introduced but looks like you forgot to add endif. deployment is currently broken.

MalloZup: https://github.com/SUSE/ha-sap-terraform-deployments/pull/185

PR #185, opened 2019-09-19
FIX: add endif missing

MalloZup:
there was a missing endif


MalloZup:
@juadk you are faster then me.. :+1: I will test in libvirt and merging.. Sorry for :hankey:  But i copied the old pillars thinking it was new.. :face_with_head_bandage:

juadk:
> @juadk you are faster then me..  I will test in libvirt and merging.. Sorry for  But i copied the old pillars thinking it was new..

I already had a stack deployed in AWS eh eh. No worries

PR #187, opened 2019-09-20
Fix HWCCT bench if 'hwcct=false'

ldevulder:
if `hwcct` variable is set to `false`, then `qa_mode/init.sls` is executed as an empty file, and so Salt exits with a `RC != 0`.

This commit fix it by adding a `test.nop` if `hwcct=false`.

Successfully manually tested on AWS.

### DIFF
--- a/salt/qa_mode/init.sls
+++ b/salt/qa_mode/init.sls
@@ -1,18 +1,17 @@
 {% if grains.get('hwcct') == true and 'hana01' in grains['hostname'] %}
-
 hwcct-config-file:
- file.managed:
-   - template: jinja
-   - names:
-     - /root/salt/qa_mode/files/hwcct/hwcct_config.json:
-       - source: salt://qa_mode/files/hwcct/hwcct_config.json.jinja
+  file.managed:
+    - template: jinja
+    - names:
+      - /root/salt/qa_mode/files/hwcct/hwcct_config.json:
+        - source: salt://qa_mode/files/hwcct/hwcct_config.json.jinja
 
 hwcct-bench-file:
- file.managed:
-   - template: jinja
-   - names:
-     - /root/salt/qa_mode/files/hwcct/hwcct_bench.sh:
-       - source: salt://qa_mode/files/hwcct/hwcct_bench.jinja
+  file.managed:
+    - template: jinja
+    - names:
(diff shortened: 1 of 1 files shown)


salt/qa_mode/init.sls
```diff
@@ -1,23 +1,25 @@
...
+      - /root/salt/qa_mode/files/hwcct/hwcct_config.json:
+        - source: salt://qa_mode/files/hwcct/hwcct_config.json.jinja
 
 hwcct-bench-file:
- file.managed:
-   - template: jinja
-   - names:
-     - /root/salt/qa_mode/files/hwcct/hwcct_bench.sh:
-       - source: salt://qa_mode/files/hwcct/hwcct_bench.jinja
+  file.managed:
```
MalloZup: why this space?

salt/qa_mode/init.sls
```diff
@@ -1,23 +1,25 @@
...
+        - source: salt://qa_mode/files/hwcct/hwcct_bench.jinja
 
 hwcct:
   cmd.run:
     - name: sh /root/salt/qa_mode/files/hwcct/hwcct_bench.sh
     - require:
       - file: hwcct-config-file
       - file: hwcct-bench-file
+{% else %}
+no-error-in-that-case:
```
MalloZup: in which case?

I would just call this state `no-operation` or something similar.

Additionally one could add a reference issue in a comment.
https://github.com/uyuni-project/sumaform/blob/8c5a2fa1707ccea52a08c92925cf06f926f75241/salt/repos/minion.sls#L27

salt/qa_mode/init.sls
ldevulder: The space on `source:`? Because there was one previously and as the previous line has a `:` at the end I though it was intentionally intended. But I can remove it if needed.

salt/qa_mode/init.sls
ldevulder: Done!

I also was not sure to add this in a `{% else %}` statement, but I think it's better to not always do this, WDYT?

MalloZup:
thx!

MalloZup:
@ldevulder @juadk  feel free to review or merge thx! :dog2:

Issue #190, opened 2019-09-20
RFC: remove `monitoring_hosts` from root tfvars


stefanotorresi:
from the followin discussion on RC:

> Stefano Torresi @storresi
> 11:34 AM
> hey folks, I have a question: why are we not just using var.hosts_ip as an input for monitored_hosts in the monitoring tf module? is there any use case at all for wanting to monitor different ips than all the nodes?
> having both hosts_ip and monitored_hosts in the root tfvars, with both always the same values, looks redundant, at first glance
> 
> 
> Xabier Arbulu @xarbulu
> 11:39 AM
> Stefano Torresi Yes, this might be an option. This variable has an history behind. Previously, we added there the address+port to be monitored, by finally It was just changed to address. I guess your point was not considered after that change. But yes, maybe we could remove that parameter now. Dario Maiocchi would be glad to do that XD
> Actually, now I see it. It makes total sense... I guess we want to monitor all the hosts if we enable the monitoring stack. In fact, if some node doesn't have an exporter, it's not harmful anyway

pros:
- one less variable to document
- simpler user-facing api (i.e. the t fvars)

cons:
- less flexibility, I guess?

What's the use case for having these values to be independent?

arbulu89:
I vote to remove that argument. I totally doubt that we will scrape metrics from nodes outside our cluster (does it have any sense)?
In sake of simplicity, it would make things easier.
Opinions @MalloZup , @ayoub-belarbi ?

MalloZup:
Ok for removal. THX

We should add the host ip adress.

stefanotorresi:
apparently the word `address` doesn't trigger automatic issue closure on GitHub, as opposed to GitLab.

MalloZup:
`fix` #number

PR #191, opened 2019-09-20
remove `monitored_hosts` var from the outer layer of the tf config API

stefanotorresi:
Addresses #190.

The `monitored_hosts` variable is now just an internal detail of the `monitoring` module (in libvirt).

For the time being, it has also been removed entirely from the Azure provisioning (except in the salt grain), but it will probably be reintroduced as a module input variable later.

### DIFF
--- a/azure/monitoring.tf
+++ b/azure/monitoring.tf
@@ -3,11 +3,6 @@ variable "timezone" {
   default     = "Europe/Berlin"
 }
 
-variable "monitored_hosts" {
-  description = "IPs of hosts you want to monitor"
-  type        = list(string)
-}
-
 variable "monitoring_srv_ip" {
   description = "monitoring server address"
   type        = string
--- a/azure/salt_provisioner.tf
+++ b/azure/salt_provisioner.tf
@@ -64,7 +64,7 @@ partitions:
   5:
     start: 80%
     end: 100%
- 
+
 EOF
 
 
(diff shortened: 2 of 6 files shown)


MalloZup:
f

MalloZup:
So the original was monitored services thiw was different from now on

MalloZup:
testing the azure part so we can merge this

stefanotorresi:
@arbulu89 I'm not sure about removing the internal variable too.

After all, the monitoring service still needs to know what to monitor. I just found to be redundant having _both_ `hosts_ip` and `monitoring_hosts` in the root tf vars, but having some kind of input within the module makes sense.

MalloZup:
In theory we could remove this `if` since it will be alwasy defined `host_ips`
https://github.com/SUSE/ha-sap-terraform-deployments/blob/502065aea3a936c2c654b1fa147c3419e8ed0807/salt/monitoring/prometheus/prometheus.yml#L16

MalloZup:
To me this PR is ok. We might open a follow-up issue if we might need to remove more things. But is already a big improvement in term of removal :)

Regarding the `monitored_host` grains I think is ok to have it since is more readable and is not really an exposed var

arbulu89:
> In theory we could remove this `if` since it will be alwasy defined `host_ips` or `monitored_hosts`
> 
> https://github.com/SUSE/ha-sap-terraform-deployments/blob/502065aea3a936c2c654b1fa147c3419e8ed0807/salt/monitoring/prometheus/prometheus.yml#L16

Yes, I guess we could remove this condition as everytime `monitoring_enabled` is set to True, this data will come.

PR #194, opened 2019-09-23
The port of ha-exporter has changed to 9002 and remove superficial if

MalloZup:
previously it was the hawk-exporter (deprecated)
now the ha-cluster exporter is running on port 9002

### DIFF
--- a/azure/network.tf
+++ b/azure/network.tf
@@ -395,13 +395,13 @@ resource "azurerm_network_security_group" "mysecgroup" {
     destination_address_prefix = "*"
   }
   security_rule {
-    name                       = "hawkExporter"
+    name                       = "ha-exporter"
     priority                   = 1007
     direction                  = "Inbound"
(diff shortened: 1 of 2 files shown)


stefanotorresi:
@MalloZup is there any way for salt to ensure that the `monitored_hosts` grain is always present?

MalloZup:
@stefanotorresi we are setting the `monitored_hosts` every deployment so we assume that it is 100% there each time.
https://github.com/SUSE/ha-sap-terraform-deployments/pull/191/files#diff-d4e4274c023136268670bd8e01283dc2R95

stefanotorresi:
so if users try to use salt outside terraform, their on their own, right? if that's so, LGTM :)

MalloZup:
(https://github.com/SUSE/ha-sap-terraform-deployments/issues/195) follow-up :robot:

PR #196, opened 2019-09-23
Add monitoring doc in tfvars.example

juadk:
As `monitoring_srv_ip` is a mandatory variable and there is no default value. We must add it in `terraform.tfvars.example`.

Otherwise, terraform will ask for value:
```
var.monitoring_srv_ip
  monitoring server address
  Enter a value
```

I could do the same for libvirt but I found something incompatible:
[Monitoring](https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/doc/monitoring.md) doc says `monitoring_enabled default True.`
In libvirt, `monitoring_enabled `default is false.

@arbulu89 @MalloZup WDYT?

### DIFF
--- a/azure/terraform.tfvars.example
+++ b/azure/terraform.tfvars.example
@@ -14,7 +14,7 @@ ninstances = "2"
 az_region = "westeurope"
 
 # Variable for init-nodes.sh script
-init-type = "all"
+init_type = "all"
 
 # SLES4SAP image information
 # If custom uris are enabled public information will be omitted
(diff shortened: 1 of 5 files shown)


azure/terraform.tfvars.example
```diff
@@ -103,6 +103,14 @@ ha_sap_deployment_repo = ""
 # Run provisioner execution in background
 #background = true
 
+# Monitoring variables
+
+# Enable the host to be monitored by exporters
+#monitoring_enable = false
```
MalloZup: ```#monitoring_enabled`

MalloZup:
@juadk  the plan is to make Monitoring disabled by default. If you find out in the doc that is enabled default , feel free to adapt the documentation.

many thx for this PR :+1:

azure/terraform.tfvars.example
juadk: done thx

Issue #169, opened 2019-09-11
Documentation update about workspace name


juadk:
We need to precise that the terraform workspace name does not contain `-` or `_ `characters. Otherwise, HANA installation will failed with a "invalid hostname" message.

MalloZup:
@juadk do you have encountered this with azure? afaik in libvirt is not the case.. But yop we can add it to documentation

juadk:
> @juadk do you have encountered this with azure? afaik in libvirt is not the case.. But yop we can add it to documentation

@MalloZup issue encountered today with AWS, @ldevulder told me that it's the same with Azure but I didn't test it myself for the moment.

MalloZup:
fixed bu https://github.com/SUSE/ha-sap-terraform-deployments/pull/205

PR #205, opened 2019-09-26
Update doc with workspace

MalloZup:
https://github.com/MalloZup/ha-sap-terraform-deployments/blob/doc-work/doc/workspaces-workflow.md

### DIFF
--- a/doc/workspaces-workflow.md
+++ b/doc/workspaces-workflow.md
@@ -15,7 +15,7 @@ Some excerpts:
 > Workspaces are technically equivalent to renaming your state file. They aren't any more complex than that. Terraform wraps this simple notion with a set of protections and support for remote state.
 
 
-```IMPORTANT```: the terraform workspace name must not contain - or _ characters.  Otherwise you will encounter failures by different cloud providers
+```IMPORTANT```: the terraform workspace name **must not contain** `-` or `_` characters.  Otherwise you will encounter failures by different cloud providers
 
 ## TL;DR
 
--- a/doc/workspaces-workflow.md
+++ b/doc/workspaces-workflow.md
@@ -14,17 +14,14 @@ Some excerpts:
 
 > Workspaces are technically equivalent to renaming your state file. They aren't any more complex than that. Terraform wraps this simple notion with a set of protections and support for remote state.
 
-The proposed workflow is the following:
 
-  - Workspaces created by any user must be prefixed with `$USER-` and end with a simple but descriptive name for the deployment (e.g., `$USER-testing-ha`) or the GIT branch (e.g., `$USER-add-ha`).  In the case of OpenQA workers, it could be `openqa-$TEST_NUMBER`.
-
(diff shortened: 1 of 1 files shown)


juadk:
> The unique thing that I would actually comment is that the workspace impacts in the name of the resources as this would become a prefix

Yes that's why we got a failure in AWS and Azure using _ in workspace name (#169).

MalloZup:
yes that's why I have added the note. So can we merge this :grapes:

doc/workspaces-workflow.md
```diff
@@ -14,17 +14,14 @@ Some excerpts:
 
 > Workspaces are technically equivalent to renaming your state file. They aren't any more complex than that. Terraform wraps this simple notion with a set of protections and support for remote state.
 
-The proposed workflow is the following:
 
-  - Workspaces created by any user must be prefixed with `$USER-` and end with a simple but descriptive name for the deployment (e.g., `$USER-testing-ha`) or the GIT branch (e.g., `$USER-add-ha`).  In the case of OpenQA workers, it could be `openqa-$TEST_NUMBER`.
-
-  - Where possible, the tags and labels of the resources created must be prefixed by the workspace, accessible with `${terraform.workspace}` to make it easier to remove resources from the web console when `terraform destroy` can not. Other resources may need to be renamed to conform to this nomenclature to avoid namespace clashes.
```
ldevulder: Why did you removed this part? FMPOV could be still useful.

MalloZup: @ldevulder  I removded because it contains the `-` chars.. :grin: 
Let me re-adapt it

doc/workspaces-workflow.md
```diff
@@ -14,17 +14,14 @@ Some excerpts:
...
+```IMPORTANT```: the terraform workspace name **must not contain** `-` or `_` characters.  Otherwise you will encounter failures by different cloud providers
```
alvarocarvajald: I think `-` is accepted, at least I've used workspace names such as `build-1-35` in Azure the past. Problem with `_` is that it is a character that the DNS RFC forbids as part of the FQDN, and this configuration adds the workspace as a prefix to several resources, including hostnames. Additionally, workspace names which contain `.` would also not be a good idea

doc/workspaces-workflow.md
```diff
@@ -14,17 +14,14 @@ Some excerpts:
...
 
 ## TL;DR
 
 To create a new workspace:
 
-`terraform worspace new $USER-test-amazing-stuff`
+`terraform worspace new $USER`
```
alvarocarvajald: Same as above. AFAIK, dashes `-` are OK.

doc/workspaces-workflow.md
```diff
@@ -40,7 +37,7 @@ To list all workspaces:
 
 To remove the previously created workspace:
 
-`terraform workspace delete $USER-test-amazing-stuff`
+`terraform workspace delete $USER`
```
alvarocarvajald: Idem.

MalloZup:
Ok  closing fo moment. thx for review

PR #209, opened 2019-09-27
lower the iscsi discovery timeout

### LINKED ISSUE #207: iscsi initiator timeout too large
Currently the timeout to wait until an iscsi target is 15000 seconds (more than 4 hours...).
This value looks too large.

https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/salt/hana_node/iscsi_initiator.sls#L26

We should reduce it or add the option to customize that (it would require other extra variable).
I don't know if customization would a good idea (other var in terraform that has to be bypassed to salt).

Opinions?

stefanotorresi:
close #207


stefanotorresi:
45 minutes ought to be enough, no?

MalloZup:
if it is not enough you pay me a coffee

Issue #199, opened 2019-09-25
python3.6 is needed for hanadb_exporter but not available by default on SLE12


ldevulder:
The last issue we are facing is that `hanadb_exporter` requires `python3.6` and this version is not installed by default on 12-SP5 PC images (and should not also on 12-SP4), and it's not possible to install it, as this OS is not available on SCC during Build Validation.

As `python3.6` will be available in SCC after release of 12-SP5, it will be easy for customers to install it after registration or this can be done by the Salt part, but definitely not during Build Validation.

juadk:
Some traces taken this morning with qa_mode = true:
```
----------
          ID: hanadb_exporter
    Function: pkg.installed
      Result: False
     Comment: Attempt 1: Returned a result of "False", with the following comment: "An error was encountered while installing package(s): Zypper command failure: Check Zypper's logs."
              Attempt 2: Returned a result of "False", with the following comment: "An error was encountered while installing package(s): Zypper command failure: Check Zypper's logs."
              An error was encountered while installing package(s): Zypper command failure: Check Zypper's logs.
     Started: 06:44:42.639353
    Duration: 32137.84 ms
     Changes:   
----------
          ID: python3-PyHDB
    Function: pkg.installed
      Result: True
     Comment: The following packages were installed/updated: python3-PyHDB
     Started: 06:45:14.806337
    Duration: 3115.107 ms
     Changes:   
              ----------
              python3-PyHDB:
                  ----------
                  new:
                      0.3.5.dev-1.1
                  old:
----------
          ID: configure_exporter_prd_00
    Function: file.managed
        Name: /etc/hanadb_exporter/prd_00.json
      Result: False
     Comment: One or more requisite failed: hana.exporter.hanadb_exporter
     Started: 06:45:18.002291
    Duration: 0.015 ms
     Changes:   
----------
          ID: start_exporter_prd_00
    Function: service.running
        Name: hanadb_exporter@prd_00
      Result: False
     Comment: One or more requisite failed: hana.exporter.configure_exporter_prd_00
     Started: 06:45:20.610060
    Duration: 0.016 ms
     Changes:   
```
and 
```
zypper in hanadb_exporter
Loading repository data...
Reading installed packages...
Resolving package dependencies...

Problem: nothing provides python(abi) = 3.6 needed by hanadb_exporter-0.3.3-1.1.noarch
 Solution 1: do not install hanadb_exporter-0.3.3-1.1.noarch
[...]

Issue #202, opened 2019-09-26
handle better monitoring repo


MalloZup:
Right now the monitoring repo is hardcode to salt 
https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/salt/monitoring/init.sls#L3

we should use a variable like the ha-factory repo.

Perhaps we should condensate this both 2 vars in a list. (@diegoakechi @stefanotorresi @arbulu89 @ayoub-belarbi )

I think personally having 2 variables, preserving the ha-factory repo and adding the monitoring repo is OK, but i'm ok to have a list also

stefanotorresi:
I'm not sure: one could still customise the pillars to change this, so I don't see a that bad as it stands.

MalloZup:
the problem right now is that the repository is hardcoded to SLE15 sytstems. The repo should vary for os . Like if you have another system we will always inject that repo

ayoub-belarbi:
I agree, it should be dynamic and we shouldn't do it this way.. the easiest solution will be having a separate var for now, the optimal solution will be having the packages in Package Hub!

stefanotorresi:
> The repo should vary for os
Oh, okay! Should all the repos match the same OS version, then?

In any case, a map of `package -> repo` sounds the nicest way to do this, it leaves the door open for extension without modifying existing code.

arbulu89:
Closing! The issue is solved after the last PR

PR #213, opened 2019-09-30
Feature/fix monitoring repo

arbulu89:
Fix:
https://github.com/SUSE/ha-sap-terraform-deployments/issues/199
https://github.com/SUSE/ha-sap-terraform-deployments/issues/202
About this second, now if the `monitoring_image` is set this will be used in the monitoring server. If the value is empty the same image than in the hana nodes will be used.

Besides that, I have fixed the monitoring server usage when the stack is disabled.

Tested and working in:
SLES15 libvirt
SLES12SP3 libvirt
SLES12SP4 azure

### DIFF
--- a/salt/hana_node/monitoring.sls
+++ b/salt/hana_node/monitoring.sls
@@ -1,7 +1,10 @@
+{% set repository = 'SLE_'~grains['osrelease_info'][0] %}
+{% set repository = repository~'_SP'~grains['osrelease_info'][1] if grains['osrelease_info']|length > 1 else repository %}
+
 server_monitoring_repo:
  pkgrepo.managed:
-    - humanname: Server:SLE15:Monitoring
-    - baseurl: https://download.opensuse.org/repositories/server:/monitoring/SLE_15/
+    - humanname: Server:Monitoring
+    - baseurl: https://download.opensuse.org/repositories/server:/monitoring/{{ repository }}/
     - refresh: True
     - gpgautoimport: True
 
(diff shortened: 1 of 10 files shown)


salt/hana_node/monitoring.sls
```diff
@@ -1,7 +1,10 @@
+{% set repository = 'SLE_'~grains['osrelease_info'][0] %}
```
MalloZup: I think that we should add a TF variable here instead of detecting it dynamically.

Because this e.g doesn't work for openSUSE. We should add a TF var 

`MONITORING_REPO`  etc. This will allow us to inject the repo without handling it  imho

diegoakechi: I agree here. I would create a variable and set the `server:monitoring` repository as default for user convenience.

arbulu89: I doubt if the deployment would work even without this change in openSUSE, honestly.

I don't see much advantage of adding a new variable with the repository, just more work expanding this to the cloud providers. 

And if we want to make this work for openSUSE, we could just adapt this code for that, I think this would be the best.

What do you think @stefanotorresi, @diegoakechi , @ayoub-belarbi , @MalloZup ?

MalloZup: I'm ok to merge it now as it is . In future we might just fix the support. I'm ok to be minimalist don't ask me :grin:

arbulu89:
Ok, merging by now.

PR #217, opened 2019-10-01
Add informations and fix 'output.tf'

ldevulder:
`output.tf` file is not consistent across each providers.

This commit fixes this and also add some information.

This is needed to be able to use this project in our Public Cloud tests in openQA.

Tested on AWS, GCP and Azure.
I can't test on libvirt, so if anyone can check if it's good a relevant.

EDIT: libvirt modifications have been removed, to different, will be done in another PR
EDIT2: GCP change is handled in @juadk 's PR #219.

### DIFF
--- a/azure/outputs.tf
+++ b/azure/outputs.tf
@@ -1,26 +1,38 @@
-# Launch SLES-HAE of SLES4SAP cluster nodes
+# Outputs:
+# - Private IP
+# - Public IP
+# - Private node name
+# - Public node name
 
-data "azurerm_public_ip" "monitoring" {
-  count = var.monitoring_enabled == true ? 1 : 0
-  name  = element(azurerm_public_ip.monitoring.*.name, count.index)
-  resource_group_name = element(
-    azurerm_virtual_machine.monitoring.*.resource_group_name,
-    count.index,
-  )
-}
-
-output "monitoring_ip" {
-  value = data.azurerm_public_ip.monitoring.*.ip_address
-}
+# iSCSI server
 
 data "azurerm_public_ip" "iscsisrv" {
   name                = azurerm_public_ip.iscsisrv.name
   resource_group_name = azurerm_virtual_machine.iscsisrv.resource_group_name
 }
 
-output "iscsisrv_ip" {
-  value = data.azurerm_public_ip.iscsisrv.ip_address
+data "azurerm_network_interface" "iscsisrv" {
+  name                = azurerm_network_interface.iscsisrv.name
+  resource_group_name = azurerm_virtual_machine.iscsisrv.resource_group_name
 }
+
+output "iscsisrv_ip" {
+  value = [data.azurerm_network_interface.iscsisrv.private_ip_address]
+}
+
+output "iscsisrv_public_ip" {
+  value = [data.azurerm_public_ip.iscsisrv.ip_address]
+}
+
+output "iscsisrv_name" {
+  value = [azurerm_virtual_machine.iscsisrv.name]
+}
+
+output "iscsisrv_public_name" {
+  value = [data.azurerm_public_ip.iscsisrv.fqdn]
+}
+
+# Cluster nodes
 
 data "azurerm_public_ip" "clusternodes" {
   count = var.ninstances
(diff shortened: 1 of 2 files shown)


ldevulder:
> I would modify that.

Ok, in that case I will simply remove the change for libvirt output, easier. Thanks!

ldevulder:
> I think you removed the libvirt part @ldevulder ? regarding @arbulu89 comment I can't see where he refer too. In tha tcase is imho ok

Yes I remove it, as it is completely different in libvirt.

ldevulder:
I will also remove GCP, as this change is handled in @juadk 's PR #219.

ldevulder:
I try to have the same output for `iscsisrv` as for `monitoring` (it's more coherent) but it seems that I still have some issues, the following message appears 4 times:
`Error: Error in function call: Call to function "element" failed: cannot use element function with an empty list.`

I don't know why. Maybe I missed something (high probability!) or maybe I'm fencing the following bug: https://github.com/hashicorp/terraform/issues/21917 (but I'm really more in favor of the 1st idea).

arbulu89:
@ldevulder As the iscsi server always is just 1, you cannot use the count item. I guess anyway that is OK to print a list for monitoring and a single value for iscsi

juadk:
> Is this some use case of openQA where you parse the output? I think you can add the logic in your parsing script and don't make this project depend on that

@arbulu89 , Exactly, output is parsing in openQA, that's why we thought have identical output everywhere.

arbulu89:
> > Is this some use case of openQA where you parse the output? I think you can add the logic in your parsing script and don't make this project depend on that
> 
> @arbulu89 , Exactly, output is parsing in openQA, that's why we thought have identical output everywhere.

I agree that having some naming is good, but I wouldn't change this project so drastically for that matter. So, my question? How is the output done in all the provider? Some print as a list, others not? I mean the iscsi server IP. Adding the count when is not needed is something that I would like to avoid...

arbulu89:
Anyway, if this makes your life easier and everyone else agrees with that change, I guess that it won't harm much to add this `count` finally...

azure/outputs.tf
```diff
@@ -1,27 +1,47 @@
...
-  name                = azurerm_public_ip.iscsisrv.name
-  resource_group_name = azurerm_virtual_machine.iscsisrv.resource_group_name
+output "iscsisrv_ip" {
+  value = data.azurerm_network_interface.iscsisrv.*.private_ip_address
 }
 
-output "iscsisrv_ip" {
-  value = data.azurerm_public_ip.iscsisrv.ip_address
+output "iscsisrv_public_ip" {
+  value = data.azurerm_public_ip.iscsisrv.*.ip_address
```
arbulu89: Maybe we could just do something like: `list(data.azurerm_publick_ip_iscsisrv.ip_address)` to print as a list and avoid the `count` addition

alvarocarvajald:
> > Is this some use case of openQA where you parse the output? I think you can add the logic in your parsing script and don't make this project depend on that
> 
> @arbulu89 , Exactly, output is parsing in openQA, that's why we thought have identical output everywhere.

Are we going to parse the iscsi server information as well in the openqa tests? I'm with @arbulu89 in this: from a visual standpoint, the output for the iscsi server IP should be a single element since there's only one, while the cluster nodes' IPs should be a list (not a list of lists though). If we need to parse this in openqa, we could do:

```
$ips = $output->{cluster_nodes_public_ip}->{value};
$iscsi_ip = $output->{iscsisrv_public_ip}->{value};
```

Then it would only be a matter of treating `$ips` as an array and `$iscsi_ip` as a single element, or better yet, push `$iscsi_ip` into the array. But looking at the openqa code, don't know how much interaction we would require with the iscsi server there.

azure/outputs.tf
ldevulder: `list` is not supported anymore on TF 0.12 and `tolist` doesn't work, but the correct syntax now is to use `[ ]`: `value = [data.azurerm_public_ip.iscsisrv.ip_address]`.

So I will change for this and simplify my PR :-)

ldevulder:
> But looking at the openqa code, don't know how much interaction we would require with the iscsi server there.

It's mainly to have something coherent, for now yes I don't need the iscsi server information in openQA but I don't know in the future.

But as there is a simple way to do this with TF (now!) I think it's the good place to do it.

alvarocarvajald:
> > But looking at the openqa code, don't know how much interaction we would require with the iscsi server there.
> 
> It's mainly to have something coherent, for now yes I don't need the iscsi server information in openQA but I don't know in the future.
> 
> But as there is a simple way to do this with TF (now!) I think it's the good place to do it.

But we're forcing a list when, semantically, it should be a single element as there's only one iscsi server. It's really a very minor thing though, so no problem from me.

ldevulder:
Tested with `[ ]` on Azure and AWS: **OK**.
GCP was done in PR #219.

MalloZup:
I'm merging this PR because we want to release.

I have seen the changes requested by @arbulu89  were adressed.

Let's release a snapshot. thx

PR #219, opened 2019-10-01
Add monitoring to GCP

juadk:
Enable monitoring for GCP with an option to disable it

GCP documentation needs to be rework as is in AWS or Azure (future PR).

### DIFF
--- a/gcp/disks.tf
+++ b/gcp/disks.tf
@@ -21,3 +21,10 @@ resource "google_compute_disk" "node_data2" {
   zone  = element(data.google_compute_zones.available.names, count.index)
 }
 
+resource "google_compute_disk" "monitoring_data" {
+  count = var.monitoring_enabled == true ? 1 : 0
+  name  = "${terraform.workspace}-${var.name}-monitoring-data"
+  type  = "pd-standard"
+  size  = "20"
+  zone  = element(data.google_compute_zones.available.names, 1)
+}
--- a/gcp/outputs.tf
+++ b/gcp/outputs.tf
@@ -1,10 +1,59 @@
-# Outputs: Public IP address where the service will be listening on
+# Outputs:
+# - Private IP
+# - Public IP
+# - Private node name
+# - Public node name
+
+# iSCSI server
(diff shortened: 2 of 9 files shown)


gcp/outputs.tf
```diff
@@ -1,10 +1,56 @@
...
+output "cluster_nodes_public_name" {
+  value = []
+}
+
+# Monitoring
+
+output "monitoring_nodes_ip" {
+  value = google_compute_instance.monitoring.*.network_interface.0.network_ip
+}
+output "monitoring_nodes_public_ip" {
```
ldevulder: An empty line could help the reading ;-)

gcp/outputs.tf
```diff
@@ -1,10 +1,56 @@
...
+  value = google_compute_instance.monitoring.*.network_interface.0.access_config.0.nat_ip
+}
+output "monitoring_nodes_name" {
+  value = ""
+}
```
ldevulder: Should be: `value = google_compute_instance.monitoring.*.name`

gcp/disks.tf
```diff
@@ -21,3 +21,9 @@ resource "google_compute_disk" "node_data2" {
   zone  = element(data.google_compute_zones.available.names, count.index)
 }
 
+resource "google_compute_disk" "monitoring_data" {
+  name = "${terraform.workspace}-${var.name}-monitoring-data"
```
arbulu89: I guess we don't need to create the disk if the monitoring is not enabled, right?
We could add the same condition than in the VM

juadk: > I guess we don't need to create the disk if the monitoring is not enabled, right?
> We could add the same condition than in the VM

You are right, good catch thx

juadk: @arbulu89 Done

juadk:
> Without testing personally the changes it looks good and ready to be merged! Thanks @juadk

Thanks @arbulu89 , I've tested it quickly with and without the monitoring part and it worked well.

PR #223, opened 2019-10-03
add sbd device visualisation

MalloZup:
The Health of device is the last one at the buttom right `SBD`

image

A potential user will need to change the device/s names and add others. 

___
you can see it here: 

http://10.162.32.94/d/Q5YJpwtZk1/ha-sap-hana-cluster-status?orgId=1


MalloZup:
image


# in case of error (here fake)

MalloZup:
Thx! Lets wait for @dakechi pov

PR #228, opened 2019-10-16
Feature/netweaver implementation

arbulu89:
Here the SAP Netweaver environment deployment!

Some important things to consider and discuss:

- ~~Right now, in order to make the Netweaver env available some changes must be done in the `main.tf` file. We could obviously add other extra parameter like `enable_netweaver` and avoid the `main.tf.netweaver` creation. What do you think?~~ `netweaver_enabled` variable added

- Have a look to the documentation please, this deployment requires more actions than the previous ones.

- In order to use the latest version of `module.run` the `/etc/salt/minion` file is changed in the nodes. This means that after that all of the new `module.run` must use the new version. I think that this is a good idea, but I wanted to mention here as a quite important change. **Now on we need to be extra careful to use the proper syntax here and in the salt formulas**

- Now we need more SAP software to install Netweaver as well. This means that we cannot just add the HANA installation folder using NFS. This affects the salt pillar files, as we need to add there the HANA specific folder. This is might generate some issues. Other solution is to leave as it was and add other extra parameter like `sap_nw_inst_media`. Opinions?

- By now the netweaver pillar files are already set. In the future we can update them to set as example files like with Hana.

- The deployment takes nearly 2 hours, so be patient!


doc/netweaver.md
```diff
@@ -0,0 +1,32 @@
...
+- AAS instance
+- Database instance (this adds the required users, tables, views, etc to the current Hana database)
+
+Besides the standard installation, an additional HA cluster might be added in top of the ASCS and ERS communication to assure high availability between these two components using the *sap_suse_cluster_connector* (HA is enabled by default).
+
+More details in the official [Suse documentation](https://www.suse.com/media/white-paper/sap_netweaver_availability_cluster_740_setup_guide.pdf?_ga=2.211949268.1511104453.1571203291-1421744106.1546416539).
+
+The deployment is performed using the [sapnwbootstrap-formula](https://github.com/SUSE/sapnwbootstrap-formula).
+
+**Disclaimer: Only available for libvirt by now.**
```
MalloZup: I hope we don't forget to update this.. :grin:

salt/pre_installation/upgrade_module_run.sls
```diff
@@ -0,0 +1,6 @@
+upgrade_module_run:
```
MalloZup: perhaps we should add a comment why we need this as rationale

arbulu89: Basically the old style will be deprecated soon, so it's a good moment to adopt the new style:
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.module.html#execution-of-salt-modules-from-within-states

arbulu89: Added a comment just in case

MalloZup:
@arbulu89  regarding the `main.tf.netweaver` I think we should not add it. Afaik this was the decision we took time ago to have only 1 main.tf file. 

I think also make sense to have only 1 main.tf for CI etc.

MalloZup:
regarding:
```
Now we need more SAP software to install Netweaver as well. This means that we cannot just add the HANA installation folder using NFS. This affects the salt pillar files, as we need to add there the HANA specific folder. This is might generate some issues. Other solution is to leave as it was and add other extra parameter like sap_nw_inst_media. Opinions?
```
it make sense to me `sap_nw_inst_media.` and leave the old deployment untouched

doc/netweaver.md
```diff
@@ -0,0 +1,32 @@
+# Netweaver
```
MalloZup: @arbulu89  I would refer this also in the global readme.md

https://github.com/SUSE/ha-sap-terraform-deployments#major-features

arbulu89:
> @arbulu89 regarding the `main.tf.netweaver` I think we should not add it. Afaik this was the decision we took time ago to have only 1 main.tf file.
> 
> I think also make sense to have only 1 main.tf for CI etc.

Ok, in that case we will need to add an extra `enable_netweaver` variable. Easy!

arbulu89:
> regarding:
> 
> ```
> Now we need more SAP software to install Netweaver as well. This means that we cannot just add the HANA installation folder using NFS. This affects the salt pillar files, as we need to add there the HANA specific folder. This is might generate some issues. Other solution is to leave as it was and add other extra parameter like sap_nw_inst_media. Opinions?
> ```
> 
> it make sense to me `sap_nw_inst_media.` and leave the old deployment untouched

Yes, we can to that, but it will be strange in my opinion. I guess that the users will have the SAP installation software in the same folder, so asking to add it twice is not the best. Anyway, let's wait for more comments as what you say makes sense too.

pillar_examples/automatic/hana.sls
```diff
@@ -8,7 +8,7 @@ hana:
       instance: "00"
       password: YourPassword1234
       install:
-        software_path: /root/sap_inst
+        software_path: /root/sap_inst/51053787
```
MalloZup: is this hardcoded number needed? In my last deployment I didn't had it

Simranpal: Earlier we specify this HANA installation dir in "sap_inst_media" in terraform tfvars ..Now this directory will contain both HANA and NW as well other SAP install media in future, we need to specify the one for HANA here..

arbulu89: I commented this in a previous comment. Until now, as we were just using HANA installation software, we added here directly the path to the HANA installer. Now, as we need to install NW as well, we need to update this. At least in our case, HANA and NW installation folders are located in the same folder, that's why I updated the code and the pillar files to point the HANA folder.

As I mentioned before, there 2 solutions: The current one, where we only have one `sap_inst_media` folder and we need to update the pillar files, or adding a new `sap_nw_inst_media` what makes at least our case strange, as both installation folders are in the same folder...

We would need to know more about that. If the users have the installation folders in the same place, It wouldn't make much sense to have a new variable. 

But this is just a design issue...

doc/netweaver.md
```diff
@@ -0,0 +1,34 @@
...
+
+The deployment is performed using the [sapnwbootstrap-formula](https://github.com/SUSE/sapnwbootstrap-formula).
+
+**Disclaimer: Only available for libvirt by now.**
+
+## Quickstart
+
+In order to deploy a SAP Netweaver environment with SAP Hana some changes must be executed in terraform and salt folders. **By default SAP Netweaver is not enabled.**
+
+- In order to enable/disable the Netweaver deployment update the value of `netweaver_enabled` variable to true/false in the `terraform.tfvars` file.
```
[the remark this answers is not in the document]
arbulu89: I don't think we can have such a step by step guide.

Netweaver requires a NFS server, so this has to be created manually right now (in the future we will have NFS/DRBD for this purpose, so everything will be created with this project).

The other points are about updating the `terraform.tfvars`. It looks quite straightforward for me.

And the pillar file edition is quite tricky. The thing is that the user must know something about Netweaver, at least where is the Installation software... In that sense I don't know how I could improve that.

But in summary, the quickstart would be:
1. Have a NFS share with the explained folders structure
2. Update `terraform.tfvars`
3. Update pillar data with the correct installation folder paths

pillar_examples/automatic/hana.sls
MalloZup: as UX pov is pretty ugly :( I mean this number will change all the time making the examples deprecated :( Can we retrieve it dinamically?

stefanotorresi: as far as I can see, the fact that the installation paths are the same is just a coincidence, we shouldn't assume that this will always be the case.

arbulu89: Well, it's really hard to say... Netweaver requires a lot of installation software, so an option is to require that everything is placed in the same folder or add a new parameter for all of them (HANA, SWPM, sapexe, HANA client, NW export). To implement that we would need to add other 4 parameters in the terraform.tfvars and bypass them from terraform to salt.
I will check better options but I don't think there is a perfect solution here...
SAP provides some utilities to get information about the installation folders. I will have a look there....

ldevulder:
Looks very promising!I'm not able to test it on libvirt (don't have the resources) but +1 to test it on PC later :D

salt/netweaver_node/files/pillar/netweaver.sls
```diff
@@ -0,0 +1,80 @@
...
+  swpm_media: {{ grains['sap_inst_media'] }}
+  sapmnt_inst_media: {{ grains['netweaver_nfs_share'] }}
+  swpm_folder: SWPM_10_SP26_6
+  sapexe_folder: kernel_nw75_sar
+  additional_dvds:
+    - '51050829_3' # NW Export folder
+    - '51053787' # HANA HDB Client folder
+
+  hana:
+    host: {{ iprange }}.200
```
Simranpal: Hi @arbulu89 why is the host ip for this hana entry harcoded and 200 I could not understand

arbulu89: It's the Virtual IP address that we create in the SAP HANA environment cluster. I think this is the most logical, as if for some reason the SAP HANA machine is moved to the 2nd machine, this IP address would still work.

pillar_examples/automatic/hana.sls
ldevulder: FMPOV I agree that we can't assume that the NW and HANA software are in the same directory, customers will generally copy HANA and NW binaries in separate dir, as we do (in QA).

Small question: is the HANA and NW directory variable will be in `sls` file now instead of `tfvars`? Or is both files should have to be configured?

salt/netweaver_node/files/pillar/netweaver.sls
ldevulder: And are we sure that this IP value will work on all providers?

pillar_examples/automatic/hana.sls
juadk: I agree with @ldevulder , until now we are using this in tfvars file:
`hana_inst_master = "s3://instmasters/hana/51053381/"`
From a customer point of view, it's easier to set variable in tfvars instead of sls.

salt/netweaver_node/files/pillar/netweaver.sls
arbulu89: NW installation it's not still expanded for other cloud providers, so hart to say. But the virtual address created in the HANA cluster is available in all the providers, so It should work.

arbulu89:
Hi all! I'm checking the comments and I have to admit I expect a bit more of "help". I would appreciate if instead of adding just plain comments, you guys can suggest how would you manage the SAP installation folders topic (request changes if you don't like the current implementation actually). Otherwise we don't progress.

About the SAP software installation folder, the thing (as commented before) is that NW requires a lot of installation files and folders, so I don't know if we shouldn't request the user to put all of them in the same folder, but excluding HANA installation folder. That it's quite strange for me. So the alternatives are, add other additional parameter to ask the NW installation folders (that are many) that is different of the HANA one. The other option is the current implementation. Require to have everything in the same folder (I don't think to as this to the user is crazy, it makes sense for me, to have everything in one place).
More when NW asks for the HANA installation folder itself (in this case, it's already requiring to have everything in the same folder).

So, I will ask directly. Do you prefer to add other parameter to the terraform deployment asking for the NW installation folder (even though HANA installation folder must be there, duplicating you NFS connections to the same folder)? This will require to add other NFS connection.

Do you have any other proposal?

ldevulder:
> you guys can suggest how would you manage the SAP installation folders topic

I would say that we have to follow SAP guidelines for this. But I saw last week that guidelines can change across OS/Product versions... and that two docs can have different approaches...

> About the SAP software installation folder, the thing (as commented before) is that NW requires a lot of installation files and folders

For our NW tests we are using "only" 2 folders: one with NW installation files and one for SWPM. What other directories are needed?

> The other option is the current implementation. Require to have everything in the same folder (I don't think to as this to the user is crazy, it makes sense for me, to have everything in one place).

We could do this yes. And if it's not the "good" way we can change later. This will also let us time to wait for users comment. Honestly don't know either if one way is better to another. But if a huge number of directories are really needed for NW putting all in the same place make sense.

> More when NW asks for the HANA installation folder itself (in this case, it's already requiring to have everything in the same folder).

So if it's requiring then let's go with all in the same place!

> So, I will ask directly. Do you prefer to add other parameter to the terraform deployment asking for the NW installation folder (even though HANA installation folder must be there, duplicating you NFS connections to the same folder)? This will require to add other NFS connection.

In that case I don't want to have another variable. But my only concern is to ***not have*** a variable for the same thing in `tfvars` and in `sls` files like you currently have. This is definitely not so easy to configure ;-)

arbulu89:
As mentioned, the conflicting point in the pillar files and HANA folder removed. I will work on this feature in a future PR adopting the new design discussed in an external RFC.

PR #230, opened 2019-10-17
Disable registration in QA_MODE

juadk:
Now that we have salt-minion included in public cloud images, registration is not needed anymore for HANA/cluster nodes. And we have to make sure to not install packages outside of images.
This change does not concern iscsi/monitoring servers.

Tests:
- AWS with and without QA_MODE: OK
- GCP with and without QA_MODE: OK
- Azure with and without QA_MODE: OK

I didn't test in libvirt but I think it is not impacted by this change, @arbulu89 @MalloZup WDYT?

### DIFF
--- a/salt/salt_provisioner_script.tpl
+++ b/salt/salt_provisioner_script.tpl
@@ -2,8 +2,11 @@
 
 mv /tmp/salt /root || true
 
-# Check is qa_mode is enable
+# Check if qa_mode is enabled
 grep -q 'qa_mode: true' /tmp/grains && QA_MODE=1
+
+# Check if salt-call is installed
+which salt-call > /dev/null 2>&1 && SALT=1
 
 # Disable colors for QA_MODE
 if [[ $${QA_MODE} ]]; then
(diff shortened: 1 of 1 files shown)


arbulu89:
Hi @juadk ,
My first question is. Are you sure we need this change? If salt is already installed it wouldn't actually apply this script, unless the `salt` binaries are located in other path. Maybe this is the case...
To improve that maybe we could just run some other command like `which salt-call` or similar to detect if salt is installed or not. 
What do you think @juadk ?

Second, what about Azure? Working as well?

ldevulder:
> My first question is. Are you sure we need this change?

We mainly **NEED** to ensure that we **DO NOT** install anything on alpha/beta/RC version during Build Validation. This is why adding a test on `qa_mode` could be good.

But +1 for the `which`.

arbulu89:
@juadk @ldevulder 
I think that my comment was misunderstood, so I will rewrite it.

I totally understand the goal of the PR, to avoid registration if you are running QA, this is clear.

My comment comes more about the implementation itself. The thing that the current images bring `salt` already embedded shouldn't impact the current implementation. The script checks if `salt` is installed, and if not, it registers the system and install it. So, if the image comes with `salt` as the latest version, the script won't be executed, so we don't really need to add this new `if` condition with the `QA_MODE` thing.

Actually, now, if you want to use some older version the new changes would make the deployment fail.
Summary: I wouldn't do this change. In any case, I would just update the line:
```
elif [[ ! -e /usr/bin/salt-minion ]]; then
```
to use the output of `which salt-call` (salt installed or not).

I don't know if I'm missing some point that requires this new change...

ldevulder:
> Actually, now, if you want to use some older version the new changes would make the deployment fail.

Could be but we never did this, as Build Validation is only for new version. And the bad thing I can see is the case were a new version that we have to validate has no salt-minion installed, this will be a bug but in that specific case salt will be installed and we have the chance to not see this (as all this TF tests will be done through openQA shortly).

So for me it's better to add the test on `QA_MODE` to ensure. @juadk WDYT?

juadk:
> So for me it's better to add the test on `QA_MODE` to ensure. @juadk WDYT?
Yes I agree with you, we need to know if salt-minion is removed from images. And as you mentioned, we will not do QA with older version.

@arbulu89 Maybe you didn't pay attention but I have already changed `/usr/bin/salt-minion` test with the `which` command.

salt/salt_provisioner_script.tpl
```diff
@@ -2,9 +2,12 @@
 
 mv /tmp/salt /root || true
 
-# Check is qa_mode is enable
+# Check is qa_mode is enabled
```
arbulu89: It should be `check **if** qa_mode is enabled`

juadk: Thx

PR #238, opened 2019-10-29
Grafana dashboard improvements

stefanotorresi:
this PR includes the various changes demonstrated in the 04 Nov 2019 sprint review, plus the "Last CIB change" panel, as shown below:

Screenshot from 2019-11-04 18-37-44

you can play with this new dashboard at http://10.162.32.207/d/Q5YJpwtZk1/ha-cluster-status

### DIFF
--- a/salt/monitoring/prometheus/rules.yml
+++ b/salt/monitoring/prometheus/rules.yml
@@ -3,14 +3,14 @@ groups:
 - name: sap-hana-resource-monitoring
   rules:
   - alert: sap-hana-master-resource-down
-    expr: absent(ha_cluster_pacemaker_resources{resource=~"rsc_saphana_.*",role="master"})
+    expr: absent(ha_cluster_pacemaker_resources{id=~"rsc_saphana_.*",role="master"})
     labels:
       severity: page
     annotations:
       summary: Primary SAP-HANA resource down
 
   - alert: sap-hana-secondary-resource-absent
-    expr: absent(ha_cluster_pacemaker_resources{resource=~"rsc_saphana_.*",role="slave",status="active"})
+    expr: absent(ha_cluster_pacemaker_resources{id=~"rsc_saphana_.*",role="slave",status="active"})
     labels:
       severity: page
     annotations:
--- a/salt/monitoring/prometheus/rules.yml
+++ b/salt/monitoring/prometheus/rules.yml
@@ -102,3 +102,10 @@ groups:
       severity: page
     annotations:
       summary: node filesystem space usage is higher than 85%
+
+  - alert: stonith-disabled
+    expr: ha_cluster_pacemaker_stonith_enabled == 0
+    labels:
+      severity: page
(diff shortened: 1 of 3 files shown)


salt/monitoring/prometheus/rules.yml
```diff
@@ -102,3 +102,10 @@ groups:
       severity: page
     annotations:
       summary: node filesystem space usage is higher than 85%
+
+  - alert: stonith-disabled
+    expr: ha_cluster_pacemaker_stonith_enabled == 0
+    labels:
+      severity: page
+    annotations:
+      summary: stonith is disabled
```
diegoakechi: @stefanotorresi I suggest a more user oriented message here. Like `stonith is disabled. Clusters without a fencing mechanism are not supported and have increased risk of data loss`.

stefanotorresi:
@diegoakechi I have updated the PR and the screenshot according to the latest changes.

MalloZup:
Regarding @arbulu89  comments we can perhaps open an issue as follow-up to gather this feedback and consider it later on yep

stefanotorresi:
@arbulu89 thanks for the feedback!

> * How is evolving this `a few seconds ago`. Shouldn't be just put there the time of the last change (or elapsed seconds)?

This is a date format provided by Grafana: it takes the input as a milliseconds timestamp, and it computes a string by subtracting it from the current time. The main use case for the source metric (`ha_cluster_pacemaker_config_last_change`) is exactly knowing how long ago the CIB changed, so I figured this is the exact representation that the user wanted.
This may very well change in the future, though, because I noticed that the CIB actually changes very often; we originally thought that this would be a relatively stable value, but that's not the case, so I realise that always showing something like "a few seconds ago" or "a few minutes ago" doesn't give much value. Maybe @gao-yan can chime in on this?

> * I would test the dashboard with more than 2 nodes just in case, to see have is rendered
> * Do we need this hana_node_ip filter of the top? Maybe the data is used internally, because for the user it looks totally rendundant

Yeah, I thought about these points! I will work on both in a follow-up... :wink: 

> What is the value of the DC metric? Maybe there we should put the hostname of the machine working as DC

That whole section is just a set of counters. I guess we could add the DC hostname somewhere, but that would go in a separate panel with a dedicated query (afaik we can't extract labels from `count` queries).

> I miss an easy access to the Resources itself. With the current dashboard is not easy to know which are the running resources, and where they are running

Hmm, I will think about this and come up with some visualization. Maybe we could have a dedicated dashboard for the resource details, with a drop-down selection.

---

@MalloZup I'm glad you approve!

> As @diegoakechi mentioned I think we don't need the threshold and failcount resource. We can remove this in a follow-up issue.

I'm not sure it's worth removing those for now. Maybe in the future when we'll rationalize the space utilization further, but for the time being I would leave those panels be; I realise they don't give much actual value, but they're useful to just to show that we have those metrics implemented and at the user disposal.

PR #232, opened 2019-10-24
Improve documentation

juadk:
This PR improves documentation about QA and public cloud:
- Add a global QA documentation
- Public cloud doc standardization 
- Add more details about how to use pillars and centralize documentation to avoid duplication and easier maintenance.

PR #239, opened 2019-10-30
Implement drbd for netweaver

nick-wang:
Add drbd_node cluster for `libvirt` only. Tested on local desktop, server and `gollum` with a proper pillar file(not included in this pr.)

This pr is the minimal implementation of DRBD cluster, except the "isolated network to nat" which no harm to others.

The ongoing changes regarding docs/automatic pillar files/structure polish for hana_node,iscsi,monitor/tuning for [azure best practice](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-nfs) will be the next step.


libvirt/main.tf
```diff
@@ -17,11 +17,14 @@ resource "libvirt_volume" "base_image" {
 resource "libvirt_network" "isolated_network" {
   name      = "${terraform.workspace}-isolated"
   bridge    = var.isolated_network_bridge
-  mode      = "none"
+  mode      = "nat"
```
arbulu89: I think that `none` makes more sense here, This network is only used to connect the VM among them, not for external connectivity.

libvirt/main.tf
```diff
@@ -17,11 +17,14 @@ resource "libvirt_volume" "base_image" {
...
   addresses = [var.iprange]
   dhcp {
     enabled = "false"
   }
+  dns {
```
arbulu89: Why do we need `dns` here? I don't see any usage for it

libvirt/main.tf
```diff
@@ -78,6 +81,35 @@ module "hana_node" {
   sbd_disk_size = "104857600"
 }
 
+// drbd01 and drbd02
+module "drbd_node" {
+  source                 = "./modules/drbd_node"
+  name                   = "drbd"
+  base_image_id          = libvirt_volume.base_image.id
+  drbd_count             = var.drbd_count
```
arbulu89: We will need to change this `drbd_count` to be used with `drbd_enabled` or similar. The idea here is to `enable/disable` drbd with a boolean. Check monitoring stack as example

libvirt/main.tf
```diff
@@ -78,6 +81,35 @@ module "hana_node" {
...
+  vcpu                   = 1
+  memory                 = 1024
+  bridge                 = "br0"
+  drbd_ips               = var.drbd_ips
```
arbulu89: We could just call `host_ips` to this, I have the same implementation in the NW code

libvirt/main.tf
```diff
@@ -78,6 +81,35 @@ module "hana_node" {
...
+  drbd_disk_size         = "1024000000"
+  shared_storage_type    = var.shared_storage_type
+  iscsi_srv_ip           = var.iscsi_srv_ip
+  reg_code               = var.reg_code
+  reg_email              = var.reg_email
+  reg_additional_modules = var.reg_additional_modules
+  ha_sap_deployment_repo = var.ha_sap_deployment_repo
+  qa_mode                = var.qa_mode
```
arbulu89: Do we need `qa_mode` for drbd?

libvirt/modules/drbd_node/main.tf
```diff
@@ -0,0 +1,109 @@
+resource "libvirt_volume" "sbd" {
```
arbulu89: This will need to be change as after NW PR merge, we will have `shared_disk` module that will be used for this matter

salt/drbd_node/files/templates/drbd_cluster.j2
```diff
@@ -0,0 +1,54 @@
+{% set data = pillar.cluster.configure.template.parameters %}
```
arbulu89: For HANA example, we have put this template in the formula project itself. I know that it's not related to drbd exclusively, but I think it will be a good idea to put there, and be aligned with the HANA formula. What do you think?

salt/pre_installation/minion_configuration.sls
```diff
@@ -12,3 +12,13 @@ configure-file-roots:
             - /srv/salt
             - /usr/share/salt-formulas/states
             - /root/salt
+
```
arbulu89: This piece of code might change of file after the NW PR

salt/salt_provisioner_script.tpl
```diff
@@ -19,7 +19,7 @@ fi
 
 # SCC Registration to install salt-minion
 # The iSCSI server won't be de-registered as it is needed to install some additional packages.
-if grep -q 'role: iscsi_srv' /tmp/grains; then
+if grep -Eq 'role: iscsi_srv|role: drbd_node' /tmp/grains; then
```
arbulu89: I don't know if this makes sense...

libvirt/modules/drbd_node/cluster_info.tpl
```diff
@@ -0,0 +1,6 @@
+# -*- mode: yaml -*-
```
MalloZup: do ze need this?

libvirt/modules/drbd_node/main.tf
```diff
@@ -0,0 +1,109 @@
...
+  pool  = var.pool
+  size  = var.sbd_disk_size
+  count = var.sbd_count
+
+  xml {
+    xslt = file("modules/drbd_node/raw.xsl")
+  }
+}
+
+resource "libvirt_volume" "main_disk" {
```
MalloZup: we can be more specific here on the resource name e.g :+ 

https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/libvirt/modules/monitoring/main.tf#L5

salt/drbd_node/files/templates/drbd_cluster.j2
MalloZup: yop

nick-wang:
> I have done an initial review of the code, and my first conclusion is that it's difficult to know if everyting is OK as the NW PR will bring some changes that will affect to this code.
> For this reason, I will wait until merging the other PR (hopefully this week) to adapt the code here.
> What do you think @nick-wang ?

Thanks for the review!
I will fix the issues that Dario and you mentioned after discussion. Then wait for the NW PR merged to resolved the conflict.

nick-wang:
> A part of some change, I think we need to update also the documentation:
> 
> 1. on the README.md we should have a link to drbd as major feature.
> 
Yes, i agree. I intend to update the docs in the later PR, to avoid less conflict to NW PR and [docs change from Julien](https://github.com/SUSE/ha-sap-terraform-deployments/pull/232).

The PR of docs change was on holding by the time i work on this PR :(

> We should have a quickstart to how we can setup a cluster with drbd doc etc.
> 
> Also the drbd is only intended to works with netweaver? looking at PR title.

Do you mean SLE-HA doc, or the quick start with terraform?
I was thinking to create a more generic project/module to create DRBD on demand. I implemented based on this principle in my own terraform project, though not perfect with HCL and still in old structure.
But after talking with Xabi, we think this project is more specific for HANA + Netweaver scenario, so 
the code and parameters is aim to HANA/NW best practice without many flexible DRBD changes. 
> The idea in my head is DRBD module should support tuning some parameters in pillar, change nodes number for DRBD, change attribute of NFS share as demand. But things like multiple volumes/connections per node, different layout of volume, some advanced DRBD usage/configuration may not suitable for terraform and Netweaver scenario.

So i think the module here is for NW only with minor flexible and won't have a generic quick start doc.

libvirt/main.tf
nick-wang: Thanks! Will fix this first after discussion done.

libvirt/main.tf
nick-wang: Thanks! Will fix this first after discussion done.

libvirt/main.tf
nick-wang: Will fix this after all discussion done.

libvirt/modules/drbd_node/main.tf
nick-wang: No problem. I can pending this and adapt after NW PR merged.

salt/drbd_node/files/templates/drbd_cluster.j2
nick-wang: I can understand the reason of putting this into HANA formula. However, it may not suitable to the user want to have different HA resource option than best practice, which is not common for HANA/DRBD resource. If i put this in HANA formula, user either rebuild the package in the repo, or need to modify the template after package installed.

For example, NFS mount option is defined in the resource, example as `options="rw,async,no_subtree_check"`. It is quite common and easier for user to change here, since NFS share is an independent service.

I can also put this NFS option into drbd formula, but it may even worse. The not directly DRBD related parameters like `mount_point`, `file_system`, (especially) `virtual_ip` were questioned by Xabi :)

salt/pre_installation/minion_configuration.sls
nick-wang: Sure, i add this as well for the completion of the PR. I plan to remove this explicit when fixing the conflict with NW PR.

salt/salt_provisioner_script.tpl
nick-wang: I didn't use the same method actually, i don't like add role pre module here:)
Made the PR like this to less conflicts with current design, i was plan to fix this with later PR. Or at least go with `-q 'role:.*_node'` in your NW pr if it is enough for us.

libvirt/modules/drbd_node/main.tf
nick-wang: Will fix this after all discussion done.

libvirt/modules/drbd_node/cluster_info.tpl
nick-wang: Indeed, unnecessary in this PR.
I was using this to change the pillar template with terraform automatically. After talking with Xabi, we decide to give a template to user to change. Also there is `automatic` pillar provide. I will check whether it is suitable to DRBD/NFS scenario later

Will remove is the `automatic` work.

libvirt/main.tf
nick-wang: Because not all host are setup with a bridge like *gollum-p1*. In this case, only one net card will be created/used on that host (but which also need to mask the 1st `network_interface` of `modules/*/main.tf`). With the mode `none`, it won't actually assign the configured IP to node, which may break the address of `outputs.tf`. Not accounting the error of `outputs.tf`, it still need the later `salt` to configure.
I don't think create as a `none` then add ip with `salt` is a perfect solution (though may have other reason for terraform libvirt provider). Since not all cases the net card will named as `eth1`.

And for `nat` mode, it won't allow outside world to access the cluster, so i think it won't cause much trouble but more benefits.

libvirt/main.tf
nick-wang: `dns` is not a must, but may helpful in some cases.

Somehow, i don't have `nameserver` configured in my laptop, but exist when setup on `gollum-p1`. So with dns enabled, i could connect outside world with domain name. But it is not the main reason i enable it.
The better thing is it allow the guest to access the internal network with hostname. For example, as long as the host configured any public service like *iscsi* in intranet, guest nodes could also access it via host name directly. Also, there is no harm to enable the `dns` 
with/without `bridge` card.

nick-wang: As an example, guest setup on `gollum` could find `gollum-p1.qa.suse.de` with `dns` enabled.
Though it is not necessary when host do have a `bridge` configured.

libvirt/main.tf
arbulu89: Ok, if it helps for your use case, and it doesn't affect to the current deployment we can go ahead with that.

libvirt/main.tf
arbulu89: The same as `nat`. If it help for your use case and doesn't affect the other users deployment it's OK

libvirt/main.tf
arbulu89: Ok, having both ( with `drbd_enabled` to false by default, and `drbd_count` with a default value (2 I guess) it's OK)

pillar_examples/libvirt/drbd/drbd.sls
```diff
@@ -0,0 +1,133 @@
+# -*- mode: yaml -*-
+# vim: ft=yaml
+
```
arbulu89: I think we can remove all the comments to make the file a bit shorter, and point to the drbd-formula in the new doc that will come if the users wants an example

salt/drbd_node/files/templates/drbd_cluster.j2
arbulu89: I was not talking about the HANA formula, the DRBD formula itself. I know that it might not be related directly to DRBD but, it makes the shipping easier.

Here some examples:
https://github.com/SUSE/saphanabootstrap-formula/blob/master/templates/scale_up_resources.j2
https://github.com/SUSE/sapnwbootstrap-formula/blob/master/templates/cluster_resources.j2

The idea is to have some specific templates in the formula itself to improve the shipping. In my opinion is better to ship them in the formula and not have them in the terraform project.

We could call something like, `netweaver_cluster_resources.j2`

salt/drbd_node/hosts.sls
```diff
@@ -0,0 +1,7 @@
+{% for ip in grains['drbd_ips'] %}
```
arbulu89: Maybe we should move this salt state to `default` folder. We have exactly the same in `hana_node/hosts.sls`.
I have not considered all things but we should have a look there.
What do you think?

salt/salt_provisioner_script.tpl
arbulu89: What I mean is that this entry was just for iscsi, that's it, all the other `roles` go must pass by the `elif` part

salt/salt_provisioner_script.tpl
```diff
@@ -35,7 +35,7 @@ mkdir -p /etc/salt;mv /tmp/grains /etc/salt || true
 sh /root/salt/deployment.sh || exit 1
 
 # Salt formulas execution
-if grep -q 'role: hana_node' /etc/salt/grains; then
+if grep -Eq 'role: hana_node|role: drbd_node' /etc/salt/grains; then
```
arbulu89: This is already updated to: `if grep -q 'role:.*_node' /etc/salt/grains; then`

libvirt/main.tf
nick-wang: I tested on gollum `libvirt`, it won't effect on the original scenario.
The `nat` will benefit the host without a `bridge` configured and apply the local `qemu_uri`. But for remote deployment, still need `bridge` for connection of `salt_provisioner`

libvirt/main.tf
nick-wang: I tested them on `gollum`. Comment more on the next conversation.

salt/salt_provisioner_script.tpl
nick-wang: Fix together with the later comments of `if grep -q 'role:.*_node' /etc/salt/grains; then`

salt/salt_provisioner_script.tpl
nick-wang: Will fix like this.

salt/drbd_node/hosts.sls
nick-wang: It make more sense to move to default. I prefer to change it in another pr since it effect on other modules.

salt/drbd_node/files/templates/drbd_cluster.j2
nick-wang: > The idea is to have some specific templates in the formula itself to improve the shipping. In my opinion is better to ship them in the formula and not have them in the terraform project.
> 
> We could call something like, `netweaver_cluster_resources.j2`
> 
> PD: I guess that this specific resources template is for NW, right?

@arbulu89  Yes, it is more specific for NW.

I can put this into `saphanabootstrap-formula` templates, but it may have one potential drawback:
As long as customer want to manage the NFS share option as their demand before setup, they can't do this unless build `saphanabootstrap-formula` on their own. Or manually do this after cluster setup.
And need to add dependencies in salt/drbd_nodes with package `saphanabootstrap-formula`.

I can change for sure if you think above issue won't be a problem to customer.

nick-wang: Just thinking, maybe there is an alternative solution for this.
As long as customer put a file with same name in `ha-sap-terraform-project`, is will replace `xxx_with_the_same_filename.j2` from `saphanabootstrap-formula`.

How do you think of it?

libvirt/modules/drbd_node/cluster_info.tpl
nick-wang: Anyway, i will remove the file in this pr.

nick-wang:
One issue may need to fix in future commit, without `zypper --non-interactive --gpg-auto-import-keys refresh --force --services` (actually import gpg key of repo) after basic register, the following register of modules will fail.

salt/drbd_node/files/templates/drbd_cluster.j2
nick-wang: After talking with Xabi in RocketChat:
> Anyway, this is not a blocker right now... We can leave the template in the terraform project and decide later if there is any better place
I will fix it out of this pr.

nick-wang:
@MalloZup @arbulu89 Please help to review, I polish the pull request based on comments, except the two spots below.
+ move drbd_cluster.j2 to `saphanabootstrap-formula`
+ move `host.sls` to default with *hana_node* module together
I will submit them in a new pull request.

salt/drbd_node/hosts.sls
nick-wang: Maybe besides the `hosts.sls`, `iscsi_initiator.sls` for `sbd` also need to move to default. Will do it together.

nick-wang:
> I miss a DRBD documentation where we link it on global readme and provide to user walk-trhough how to use this feature and what is needed to be done in order to do it.

@MalloZup DRBD documentation is added. (Though sections related to automatic pillar and pillar template may need to adopt after more discussion. The current commit is enough for this pull request.)

libvirt/modules/hana_node/salt_provisioner.tf
```diff
@@ -44,7 +44,7 @@ devel_mode: ${var.devel_mode}
 reg_email: ${var.reg_email}
 reg_additional_modules: {${join(", ",formatlist("'%s': '%s'",keys(var.reg_additional_modules),values(var.reg_additional_modules),),)}}
 additional_packages: [${join(", ", formatlist("'%s'", var.additional_packages))}]
-authorized_keys: [${trimspace(file(var.public_key_location))},${trimspace(file(var.public_key_location))}]
+authorized_keys: [${trimspace(file(var.public_key_location))}]
```
arbulu89: I don't really know why this was this way... Anyway, thanks for the fix

salt/drbd_node/cluster_packages.sls
```diff
@@ -0,0 +1,16 @@
+habootstrap-formula:
+  pkg.installed:
+    - retry:
+        attempts: 3
+        interval: 15
+
+install_cluster_packages_prepare:
```
arbulu89: I think we can remove this packages installation as this is done in the habootstrap-formula

nick-wang: Haha, good catch.

In my example pillar, i disable the package installation in habootstrap-formula by `install_packages: false`. 
But i agree it is better to enable in pillar like default but remove from here. Will fix soon.

pillar_examples/libvirt/drbd/cluster.sls
```diff
@@ -0,0 +1,25 @@
...
+    password: linux
+  configure:
+    method: 'update'
+    template:
+      source: /srv/salt/drbd/templates/drbd_cluster.j2
+      parameters:
+        virtual_ip: <ADMIN_VIRTUAL_IP>
+        virtual_ip_mask: 24
+        platform: libvirt
+        prefer_takeover: true
```
arbulu89: I would say `prefer_takeover` and `auto_register` are not using by pacemaker when the cluster is created for drbd, maybe we can remove them

arbulu89: Actually, with the current template only `virtual_ip` is used

pillar_examples/libvirt/drbd/cluster.sls
```diff
@@ -0,0 +1,25 @@
+cluster:
+  name: 'hacluster'
```
arbulu89: Better to call that `drbd_cluster`

pillar_examples/libvirt/drbd/drbd.sls
```diff
@@ -0,0 +1,130 @@
...
+  ## Optional: stop the DRBD resources after initial resync
+  #stop_after_init_sync: true
+
+  ## Optional: format the DRBD resource after initial resync
+  #need_format: true
+
+
+  # Salt deployment/manage related parameters
+  salt:
```
arbulu89: @nick-wang Is this updated? If I'm not wrong be removed that `salt` entry from the formula

nick-wang: Not yet, since the new package is not yet released in update channel. Will fix it after it changed.

pillar_examples/libvirt/drbd/cluster.sls
nick-wang: Fixed in the new commit.

pillar_examples/libvirt/drbd/cluster.sls
nick-wang: Fixed in the new commit.

salt/drbd_node/cluster_packages.sls
nick-wang: Fixed in the new commit.

pillar_examples/libvirt/drbd/drbd.sls
nick-wang: Fixed in the latest commit. Though may need to change the place/format in another pull request...

nick-wang:
@arbulu89 @diegoakechi @MalloZup @stefanotorresi Thanks a lot for the review and valuable comments! I fixed the rest in the latest commit.

Will keep polish this pull request by:
+ move `host.sls` and `iscsi_initiator` to default with hana_node module together
+ Polish and change pillar/template location.

PR #240, opened 2019-10-30
Add multi tenant HANA monitoring dashboard

arbulu89:
Update HANA monitoring dashboard to select the monitoring database.

Now, all the available databases in the metrics are available.

PD: I have not found any option to set the filter for the panels in a variable, to make it easier to change without changing all the panels...


arbulu89:
@diegoakechi Could you check this PR too? It's exactly what I showed in the last sprint review.
I think that the best is to close it, and work in the cleanup in next PRs

PR #242, opened 2019-10-31
Add NFS connection for the NW inst software

arbulu89:
Even though that it looks a big change, mostly is renaming. Having Netweaver now, `sap_inst` doesn't make sense for Hana, that's why I changed. But if you prefer the old version we can just revert this renaming.

The most meaningful change is to add a new NFS share connection for Netweaver.

Depends on: https://github.com/SUSE/sapnwbootstrap-formula/pull/14
https://github.com/SUSE/salt-shaptools/pull/38

Tested with SLE15 and working!


arbulu89:
@juadk @ldevulder @MalloZup Comments? At least about the variables renaming

PR #244, opened 2019-11-06
Enable multi-cluster for hacluster (dashboard and configuration in prometheus) + doc

MalloZup:
- Add documentation
- Modify prometheus config accordingly
- adapt dashboard to reflect change with multi-cluster

# Play with:
http://10.162.31.91/d/Q5YJpwtZk1/ha-cluster-status?orgId=1&refresh=5s

# how it looks like:

See the botton in the top `hacluster` where you can switch

new-dash

### DIFF
--- a/doc/monitoring.md
+++ b/doc/monitoring.md
@@ -30,7 +30,7 @@ Currently supported exporters:
 
 # Multi-cluster monitoring:
 
-For enabling multi-cluster in prometheus and our monitoring solution, you need to follow the schema in `/etc/prometheus/prometheus.yaml`.
+For enabling multi-cluster in prometheus and in our monitoring solution, you need to follow the schema in `/etc/prometheus/prometheus.yaml`.
 
 Each cluster is a different jobname. So if you have 2 cluster you will add 2 jobnames. like :
 
@@ -58,7 +58,7 @@ scrape_configs:
         - "10.162.32.238:9002" # 9002: ha_cluster_exporter metrics
 ```
 
-This will add in prometheus a label `job="cluster-asia-00" from where you can filter your different metrics.
+This will add in prometheus a label `job="hacluster-01` and  `job="hacluster-01`. In the grafana dashboard you will have a special switch on the top to switch clusters.
 
 
 ### SAP HANA database exporter
--- a/doc/monitoring.md
+++ b/doc/monitoring.md
@@ -35,18 +35,27 @@ For enabling multi-cluster in prometheus and our monitoring solution, you need t
 Each cluster is a different jobname. So if you have 2 cluster you will add 2 jobnames. like :
 
 ```
-  - job_name: cluster-europe-00
-    metrics_path: /metrics
+scrape_configs:
+  - job_name: hacluster-01
     static_configs:
       - targets:
+        - "192.168.110.19:8001" # 8001: hanadb exporter port
+        - "192.168.110.20:8001" # 8001: hanadb exporter port
+        - "192.168.110.19:9100" # 9100: node exporter port
+        - "192.168.110.20:9100" # 9100: node exporter port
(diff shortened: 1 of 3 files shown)


salt/monitoring/prometheus/prometheus.yml
```diff
@@ -14,24 +14,17 @@ rule_files:
     - /etc/prometheus/rules.yml
 
 scrape_configs:
-  - job_name: hanadb
+  # in a multi-cluster scenario, add another job name called
+  #  hacluster-europe-02
+  - job_name: hacluster-europe-01
```
MalloZup: perhaps we can rename this better but Imho is ok

doc/monitoring.md
```diff
@@ -28,6 +28,30 @@ Currently supported exporters:
...
 - HA Cluster exporter (hawk-apiserver)
 
+# Multi-cluster monitoring:
+
+For enabling multi-cluster in prometheus and our monitoring solution, you need to follow the schema in `/etc/prometheus/prometheus.yaml`.
+
+Each cluster is a different jobname. So if you have 2 cluster you will add 2 jobnames. like :
+
+```
+  - job_name: cluster-europe-00
```
MalloZup: we could make more real this example with all scraping targets but i'm ok also with fakedata as guideline.

In this example we miss some targates.. (but they are in prometheus.yaml) I can add them

MalloZup:
PR ready to review

stefanotorresi:
@MalloZup I don't quite get how this works in the UI, and the provided demo deployment doesn't seem to be updated, could you please check it?

MalloZup:
image


if you switch the button `hacluster` you choose differnt cluster.

The dashboard is uptodate

stefanotorresi:
uhm, when I change it, the dashboard doesn't show any data anymore!

MalloZup:
@stefanotorresi  yeah that is correct because they are fake jobs. they are fake cluster.

It is to highlight that when you change it, you change cluster. I didn't had 2 cluster but one only :grimacing:

salt/monitoring/provisioning/dashboards/cluster-status-real-time-alerts.json
```diff
@@ -1666,6 +1525,53 @@
...
+      },
+      {
+        "allValue": null,
+        "current": {
+          "tags": [],
+          "text": "cluster-europe",
+          "value": "cluster-europe"
+        },
+        "datasource": "Prometheus SHAP",
+        "definition": "label_values(job)",
```
MalloZup: this is the variable that let us switch to clusters (jobname)

stefanotorresi:
Hmm, we have to test this with two clusters to see if it works properly, because I'm not sure how will it behave with the other variables we use in queries like `$hana_node_ip` and `$hana_node_name`.
Do you think you could arrange a demo, maybe by deploying the netweaver cluster too?

MalloZup:
I think it will work but ok. Let's try. Can you give me your cluster IPS somehow?

We don't need netweaver but 2 hacluster. I will use my and yours

MalloZup:
send me the adress in IRC

MalloZup:
or via mail

salt/monitoring/prometheus/prometheus.yml
stefanotorresi: I'd say to propose just a simple numbered scheme, people may not even have geographical regions.
```suggestion
  - job_name: cluster-01
```

doc/monitoring.md
stefanotorresi: could we make the documentation match the `prometheus.yml` file that we ship, for consistency?

MalloZup:
don't worry, welcome in the club of naming paranoiac. It is good to meet people in the community :grin:

PR #245, opened 2019-11-06
Fix the deployment when execution is done without reg code

arbulu89:
This PR aims to fix:

1. Issue with monitoring server deployment in Azure. Provide customizable image
2. Fix deployment when registration code is not provided. Important things:
 - Registration of iscsi server is managed now in salt
 - Registration is done only if code is provided. Otherwise salt installation will be tried from current repositories.

Pending to be tested properly!

### DIFF
--- a/azure/image.tf
+++ b/azure/image.tf
@@ -54,6 +54,26 @@ variable "iscsi_srv_uri" {
   default = ""
 }
 
+variable "monitoring_public_publisher" {
+  type    = string
+  default = "SUSE"
+}
+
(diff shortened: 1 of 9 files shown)


arbulu89:
Tested in Azure and working fine

arbulu89:
Hi @juadk and @ldevulder ,
This PR manages the deployment for the cases where the reg code is not provided (non BYOS images).
It doesn't affect in any case to the `qa_mode`, but if you want to have a look.
I will merge tomorrow otherwise

PR #246, opened 2019-11-07
Add alert showcase for failcount threshold

MalloZup:
I might need to test this later on.

I couldn't figure out how we can add a generic alert for all resource so it is better to make a specific one and sys-admin can adapt on this showcase

### DIFF
--- a/salt/monitoring/prometheus/rules.yml
+++ b/salt/monitoring/prometheus/rules.yml
@@ -26,8 +26,18 @@ groups:
     annotations:
       summary: A cluster resource failed
 
-# sbd alerts
 
+# failcount exceed migration threshold (example on saphana specific resource)
+- name: resource-failcount-higher-threshold
+  rules:
+  - alert: resource-failcount-higher-threshold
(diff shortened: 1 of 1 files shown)


salt/monitoring/prometheus/rules.yml
```diff
@@ -26,8 +26,18 @@ groups:
     annotations:
       summary: A cluster resource failed
 
-# sbd alerts
 
+# failcount exceed migration threshold (example on saphana specific resource)
+- name: resource-failcount-higher-threshold
+  rules:
+  - alert: resource-failcount-higher-threshold
+    expr: ha_cluster_pacemaker_fail_count{resource="rsc_SAPHana_PRD_HDB00"} > ha_cluster_pacemaker_migration_threshold{resource="rsc_SAPHana_PRD_HDB00"}
```
stefanotorresi: here is the correct query we should use:
```suggestion
    expr: count(ha_cluster_pacemaker_fail_count > ha_cluster_pacemaker_migration_threshold) > 0
```

MalloZup:
ff

PR #250, opened 2019-11-11
Fix the registration error of 'Valid metadata not found'.

nick-wang:
Error message when register like:
[Error]
...
Error building the cache:
[SUSE_Linux_Enterprise_Server_15_SP1_x86_64:
SLE-Product-SLES15-SP1-Updates|https://updates.suse.com/SUSE/Updates/
SLE-Product-SLES/15-SP1/x86_64/update?XXXX]
Valid metadata not found at specified URL
Some of the repositories have not been refreshed because of an error.
No provider of 'sle-module-basesystem' found.'
...

Bad consequence:
The SLE module is not succeed register. Then the following
modules like "SLE-HA" may fail due to lack of requirement.
'Unmet product dependencies. Activate one of these products
first: Server Applications Module 15 SP1 x86_64'

Fix:
1. Refresh the repo with key import:
Retrieving repository 'SLE-Module-Basesystem15-SP1-Pool' metadata [.
Automatically importing the following key:
  Repository:       SLE-Module-Basesystem15-SP1-Pool
  Key Name:         SuSE Package Signing Key <build@suse.de>
  Key Fingerprint:  FEAB5025 39D846DB 2C0961CA 70AF9E81 39DB7C82
  Key Created:      Wed Dec  7 11:57:35 2016
  Key Expires:      Sun Dec  6 11:57:35 2020
  Rpm Name:         gpg-pubkey-39db7c82-5847eb1f
done]
Forcing building of repository cache
Building repository 'SLE-Module-Basesystem15-SP1-Pool' cache [....done]
2. Redo the registration of base module.

Step to reproduce:
1. Register against SLE15SP1.
2. register modules like:
reg_additional_modules = {
    "sle-module-server-applications/15.1/x86_64" = ""
    "sle-module-containers/15.1/x86_64" = ""
    "sle-ha/15.1/x86_64" = "<HA_REG_CODE>"
}

### DIFF
--- a/salt/default/registration.sls
+++ b/salt/default/registration.sls
@@ -2,7 +2,8 @@
 {% if grains['reg_code'] %}
 register_system:
   cmd.run:
-    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }} {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] else "" }}
+    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }} {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] else "" }} ||
+        (zypper --non-interactive --gpg-auto-import-keys refresh --force --services && exit 1)
     - retry:
         attempts: 3
         interval: 15


nick-wang:
> How is this working @nick-wang ? I don't really get what's going on here.
> Is this to prevent an error in `SUSEConnect` itself, or in future zypper usages?
> Is the retry mechanism not enough?
> 
I meet this via register against SLE15SP1 with my code. 100% fail unless manually import the keys and redo the registration.

I am not 100% understand the reason of why SCC fail to load the repo, but i think @ldevulder must met this before, refer to c1893441 with file `install-salt-minion.sh`. He did the same after register, then register the base module again.

```
...
    # Register the system on SCC
    SUSEConnect -r "$regcode"

    # We have to force refresh the repos and the keys (keys may change during lifetime of this OS/image)
    zypper --non-interactive --gpg-auto-import-keys refresh --force --services

    # Register the modules accordingly with the SLE version.
    if [[ $VERSION_ID =~ ^12\.? ]]; then
      SUSEConnect -p sle-module-adv-systems-management/12/x86_64
...
```

The registration is actually failed until keys import.

> I'm not saying that we shouldn't have something like that if it's not enough robust, but the implementation looks quite strange for me... Doesn't have salt something more native for this cases?
It will be better/correct to fix in Salt/SCC side, but need to ask them to involve. And i think it is the best work around before this.

arbulu89:
> > How is this working @nick-wang ? I don't really get what's going on here.
> > Is this to prevent an error in `SUSEConnect` itself, or in future zypper usages?
> > Is the retry mechanism not enough?
> 
> @arbulu89
> I meet this via register against SLE15SP1 with my code. 100% fail unless manually import the keys and redo the registration.
> 
> I am not 100% understand the reason of why SCC fail to load the repo, but i think @ldevulder must met this before, refer to [c189344](https://github.com/SUSE/ha-sap-terraform-deployments/commit/c189344116995b93ff4b4f85895dfd6a7807acbf) with file `install-salt-minion.sh`. He did the same after register, then register the base module again.
> 
> ```
> ...
>     # Register the system on SCC
>     SUSEConnect -r "$regcode"
> 
>     # We have to force refresh the repos and the keys (keys may change during lifetime of this OS/image)
>     zypper --non-interactive --gpg-auto-import-keys refresh --force --services
> 
>     # Register the modules accordingly with the SLE version.
>     if [[ $VERSION_ID =~ ^12\.? ]]; then
>       SUSEConnect -p sle-module-adv-systems-management/12/x86_64
> ...
> ```
> 
> The registration is actually failed until keys import.
> 
> > I'm not saying that we shouldn't have something like that if it's not enough robust, but the implementation looks quite strange for me... Doesn't have salt something more native for this cases?
> > It will be better/correct to fix in Salt/SCC side, but need to ask them to involve. And i think it is the best work around before this.

What about something like:
```
{% if not grains.get('qa_mode') %}
{% if grains['reg_code'] %}
register_system:
  cmd.run:
    - name: /usr/bin/SUSEConnect -r {{ grains['reg_code'] }} {{ ("-e " ~ grains['reg_email']) if grains['reg_email'] else "" }}
    - retry:
        attempts: 3
        interval: 15

refresh_register_system_repos:
  cmd.run:
    - name: zypper --non-interactive --gpg-auto-import-keys refresh --services
    - retry:
        attempts: 3
        interval: 15
{% endif %}

{% if grains['reg_additional_modules'] %}
{% for module, mod_reg_code in grains['reg_additional_modules'].items() %}
{{ module }}_registration:
  cmd.run:
    - name: /usr/bin/SUSEConnect -p {{ module }} {{ ("-r " ~ mod_reg_code) if mod_reg_code else "" }}
    - retry:
        attempts: 3
        interval: 15
{% endfor %}

register_additional_system_repos:
  cmd.run:
    - name: zypper --non-interactive --gpg-auto-import-keys refresh --services
    - retry:
        attempts: 3
        interval: 15
{% endif %}
{% endif %}
```

Would this solve your issue? It's more salt style

nick-wang:
> Would this solve your issue? It's more salt style

I tried this before, but it won't help in some cases actually.

Because as long as the 1st registration failed for 3 times, the *refresh & import* can only help on the next registration, which is not the base module but 'reg_additional_modules'.
As long as the next `reg_additional_modules` is `"sle-module-server-applications/15.1/x86_64"`, it will go smoothly, since it is actually re-register the base module. However, if the next `reg_additional_modules` is modules like `"sle-ha/15.1/x86_64"` which require additional reg-code, it will fail due to lack of pre-requirement module. Then no chance to re-add the `sle-ha` anymore.

It is the reason why i put `refresh&import` in the same state to make sure registration will succeed in the 2nd attempt.

nick-wang:
Since both registration on base_module and additional_modules could fail, may need to add `refresh & import` to both state as a workaround before any fix on registration server side.
@arbulu89 How do you think? Can you reproduce this by adding `refresh & import` to both registration state?

nick-wang:
After talking with Thomas and Diego, we decide not do the change. Because the image should have signing key accepted before registration.
For the image we released, it should have already built-in. Otherwise we would rather raise an error instead of accept silently. Since may have risk to accept all keys automatically.

PR #251, opened 2019-11-12
drbd_nodes: Add automatic pillar of DRBD.

nick-wang:
Will add docs together with DRBD module and move static pillar file after https://github.com/SUSE/ha-sap-terraform-deployments/pull/239 merged.


nick-wang:
Close due to need change a bit.

PR #253, opened 2019-11-12
Update pillar files to work with latest salt formulas

arbulu89:
A quick PR to adapt the pillar files to work with the latest salt formulas.

To make you see which is the requirement for sle12 version mainly.

### DIFF
--- a/pillar_examples/libvirt/performance_optimized/cluster.sls
+++ b/pillar_examples/libvirt/performance_optimized/cluster.sls
@@ -17,7 +17,7 @@ cluster:
   configure:
     method: 'update'
     template:
-      source: /srv/salt/hana/templates/scale_up_resources.j2 #This path changes beyond SLES15SP1
+      source: /usr/share/salt-formulas/states/hana/templates/scale_up_resources.j2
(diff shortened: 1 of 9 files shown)


arbulu89:
Tested in libvirt and working. For the rest of the providers it should be exactly the same.

pillar_examples/libvirt/performance_optimized/cluster.sls
```diff
@@ -17,7 +17,7 @@ cluster:
   configure:
     method: 'update'
     template:
-      source: /srv/salt/hana/templates/scale_up_resources.j2 #This path changes beyond SLES15SP1
+      source: /usr/share/salt-formulas/states/hana/templates/scale_up_resources.j2
```
Simranpal: Hi @arbulu89 I was wondering why was this path to template updated

arbulu89: Because this is the new place where the formulas will be installed.
This comes will all of these last discussion about file permissions to make our formulas compatibles with SUMA

PR #254, opened 2019-11-13
Fix count argument backwards compat

stefanotorresi:
After pulling the latest updates, running any `terraform` command (including destroy)  on existing deployments will fail with the following:

```
Error: Invalid count argument

  on modules/netweaver_node/salt_provisioner.tf line 15, in resource "null_resource" "netweaver_node_provisioner":
  15:   count = var.provisioner == "salt" ? length(libvirt_domain.netweaver_domain) : 0

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.
```

This is due to the `libvirt_domain.netweaver_domain` resource not being initialized.
Using the `netweaver_count` var directly fixes the problem and allows to migrate to the new tfstate.

### DIFF
--- a/aws/salt_provisioner.tf
+++ b/aws/salt_provisioner.tf
@@ -79,7 +79,7 @@ EOF
 }
 
 resource "null_resource" "hana_node_provisioner" {
-  count = var.provisioner == "salt" ? length(aws_instance.clusternodes) : 0
+  count = var.provisioner == "salt" ? var.ninstances : 0
 
   triggers = {
     cluster_instance_ids = join(",", aws_instance.clusternodes.*.id)
--- a/azure/salt_provisioner.tf
+++ b/azure/salt_provisioner.tf
@@ -80,7 +80,7 @@ EOF
 }
 
 resource "null_resource" "hana_node_provisioner" {
-  count = var.provisioner == "salt" ? length(azurerm_virtual_machine.clusternodes) : 0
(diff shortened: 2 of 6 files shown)


stefanotorresi:
@arbulu89 I'd say we change it everywhere then: leaving aside backwards compatibility, which admittedly wouldn't work anyway at the moment but we should start thinking about it sooner or later, I really don't see any value in computing `length` when that value always comes from a variable.
Shall I proceed with updating all the other occurrences of this?

arbulu89:
> @arbulu89 I'd say we change it everywhere then: leaving aside backwards compatibility, which admittedly wouldn't work anyway at the moment but we should start thinking about sooner or later, I really don't see any value in computing `length` when that value always comes from a variable.
> Shall I proceed with updating all the other occurrences of this?

Perfect! Yes, go ahead!

stefanotorresi:
@arbulu89 done! the monitoring ones are addressed in #255

stefanotorresi:
> And in the future we should unify `ninstances` and `hana_count`, but this is another story XD

yeah, I really *hate* `ninstances`, it's incredibly annoying to read :nerd_face: 

in fact, I might just open a PR about that :stuck_out_tongue_closed_eyes:

PR #255, opened 2019-11-13
simplify monitoring resource references

stefanotorresi:
This PR simplifies HCL code that treats monitoring resources as multiple, because we actually support only one single Prometheus/Grafana host.

All monitoring related resources are now consistently looked up via the `0` index, and the outputs are strings instead of tuples.

### DIFF
--- a/aws/outputs.tf
+++ b/aws/outputs.tf
@@ -57,21 +57,17 @@ data "aws_instance" "monitoring" {
 }
 
 output "monitoring_ip" {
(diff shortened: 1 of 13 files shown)


stefanotorresi:
that's exactly why I spotted the problem: I tried to do 
```
ssh root@$(terraform output monitoring_public_ip)
````
and it failed because it wasn't a string, as I would expect, but an array :stuck_out_tongue:

PR #256, opened 2019-11-14
make drbd count setting consistent with latest changes in #254

stefanotorresi:
super-small change I forgot to mention in #239

### DIFF
--- a/libvirt/modules/drbd_node/salt_provisioner.tf
+++ b/libvirt/modules/drbd_node/salt_provisioner.tf
@@ -12,7 +12,7 @@ data "template_file" "drbd_salt_provisioner" {
 }
 
 resource "null_resource" "drbd_node_provisioner" {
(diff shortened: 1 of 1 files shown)


nick-wang:
@stefanotorresi Does it mean all `var.drbd_count` need to change to `length(libvirt_domain.drbd_domain)` or just this one?

stefanotorresi:
> @stefanotorresi Does it mean all `var.drbd_count` need to change to `length(libvirt_domain.drbd_domain)` or just this one?

the other way around, but I've only found just this single occurrence :)

nick-wang:
> > @stefanotorresi Does it mean all `var.drbd_count` need to change to `length(libvirt_domain.drbd_domain)` or just this one?
> the other way around, but I've only found just this single occurrence :)

@stefanotorresi I was misunderstanding in the beginning~ Then delete the msg and approve, lol.
Thanks for the change.

PR #259, opened 2019-11-15
Move hosts.sls from individual modules to default and fix masterless salt-minion error

nick-wang:
+ Move hosts from individual modules (*_node) to default.

+ Stop salt-minion service for masterless salt, since the daemon doesn't help on `salt-call` but cause errors.
```
Error msg:
[ERROR] DNS lookup or connection check of 'salt' failed.
[ERROR] Master hostname: 'salt' not found or not responsive.
        Retrying in 30 seconds
```


nick-wang:
> @nick-wang This is a quite important change.
> Have you tested in cloud providers? And It could even impact NW deployment.
> We need to have all these scenarios in mind

@arbulu89 I think you mean the change of `hosts.sls` instead of masterless one. (Maybe i should split the two commits)

I didn't test against CLOUD providers, only libvirt. I don't think it can cause trouble to NW, unless NW have special usage for hostname/DNS, which not allowed have `ip node` entries in /etc/hosts.
The entries will only include the NW nodes.
The current logic is it will do nothing(like monitoring, iscsi_srv) unless `host_ips` exist(hana_node, drbd_node, netweaver_node).

nick-wang:
@arbulu89 @MalloZup Thanks for the review! Now I think the CI and a `devel` branch is really helpful here XD

arbulu89:
Hey @nick-wang ,
I have an idea about how to deal with duplicated salt states that we are putting in `hana`, `drbd`, `netweaver`, etc.
What we could do, is just to create a new salt states folder like `utils` or `common` (yes I know, we can think a bit more about the name...) and place there the salt states that are common like `hosts.sls`, `iscsi_initiator.sls`, `ssh.sls`, etc.
And them, in the executions we want to use them (in `drbd_node` for example`) we could just import them using `utils.hosts`.

This way we could remove duplicated states and avoid overloading the `default` state.

What do you think?

PD: I don't want to convert this folder in a trash box, so just add we need to be careful what we put there

nick-wang:
> I have an idea about how to deal with duplicated salt states that we are putting in `hana`, `drbd`, `netweaver`, etc.
> What we could do, is just to create a new salt states folder like `utils` or `common` (yes I know, we can think a bit more about the name...) and place there the salt states that are common like `hosts.sls`, `iscsi_initiator.sls`, `ssh.sls`, etc.
> And them, in the executions we want to use them (in `drbd_node` for example`) we could just import them using `utils.hosts`.
> 
> This way we could remove duplicated states and avoid overloading the `default` state.
> 
> What do you think?

I think it is a great idea! With importing the specific statse to individual modules, it remove the duplicate code and many possible `if/else` conditions in `default` module.

Slightly idea is we keep the `default` folder, but *strictly* for the states like(ssh/hostname/registration/xxx) *must* run in all modules. Since with the increasing of modules, it is helpful to emphasis/maintain the states for a node instead of a module/cluster.

And add one folder like `common` for the states shared in some modules but not all. Then we can write the `top.sls` like:
```
base:
  '*':
    - default

  'hana_node':
    - match: grain
    - hana_node
...
```
How do you think? Anyway, i like your idea a lot.
 
> PD: I don't want to convert this folder in a trash box, so just add we need to be careful what we put there
Yes. No matter `default` or `common`. It should be strictly managed to avoid to be a trash libraries bin.

arbulu89:
Hi @nick-wang ,
Actually, after working with DRBD and Netweaver, I have found many duplicates. So, I have started the next PR (that basically covers this whole idea).
https://github.com/SUSE/ha-sap-terraform-deployments/pull/274

If you want, we can remove this part from this current PR, and only merge the salt minion service (changing the destination branch to `develop`).

I'm sorry for duplicating your PR idea, but it makes my implementation easier hehe

nick-wang:
> Hi @nick-wang ,
> Actually, after working with DRBD and Netweaver, I have found many duplicates. So, I have started the next PR (that basically covers this whole idea).
> #274
@arbulu89 Yes, i think we will have common states for all single `node` like *hostname.sls*. And states for all clusters with HA, including this *host.sls*

> If you want, we can remove this part from this current PR, and only merge the salt minion service (changing the destination branch to `develop`).
No problem, i could re-submit even this PR is closed.

> 
> I'm sorry for duplicating your PR idea, but it makes my implementation easier hehe
Don't be, we have the same target~ We can close this with your change. But i do like your suggestion with `import` the shared states (for parts of modules) from `common` folder! I think we need the group know the comment/discussion as well.

nick-wang:
Close due to part of changes about `hosts.sls` is merged.

PR #263, opened 2019-11-18
resources panel in Grafana cluster dashboard 

stefanotorresi:
This PR adds a comprehensive panel with a table representation of all the current resources running in Pacemaker.

Screenshot from 2019-11-18 13-56-42

It also removes the "data source" switcher on top of both dashboards, which had no real use, and hides the IP switcher in the HANA detailed dashboard, which also was redundant since we already have a node switcher.


stefanotorresi:
Well, it's not _completely_ removed, it's just hidden.
If users really want to, they can toggle its visualization back on from the dashboard variables preferences. I think the visualization is tighter this way, I doubt we'll ever have more than one Prometheus instance.

diegoakechi:
Late here, but well done @stefanotorresi. Looks really nice, and if we see that multiple data-sources is not an exception, it is simple to be changed in the future.

PR #269, opened 2019-11-22
Feature/drbd azure

arbulu89:
DRBD deployment implementation for Azure.

There are still a lot of improvements to be made, but by now, this PR provides a functional drbd cluster in Azure.

To use, just:
1. Set `drbd_enabled` to `True` in terraform.tfvars
2. Set `drbd_ips` in terraform.tfvars with IP addresses in the current subnet ip range

Main additions:
1. DRBD module in Azure. For that, I have implemented the `main.tf` file in the root folder. Eventually, we should transform all the components to modules.
2. Add Azure specific topics to the salt implementation. You will notice that now there are many duplicated salt states in `hana_node` and `drbd_node`. Yes, this is a quite bad implementation. I would prefer to work on that in a future PR. The purpose would be to create other salt folder `cluster_nodes` to have all the salt states that are common in the nodes that the cluster will be deployed (hana, drbd, netweaver).
3. I big refactoring of `iscsi` server and initiators. As we need more disks to enable sbd (for each cluster) we need to server more LUNS with iscsi. The current `iscsi-formula` doesn't have this option, so I have done many workaround. The next step in a future PR would be to check the new updates in iscsi-formula upstream to have a proper iscsi server deployment (and initiator).


azure/main.tf
```diff
@@ -0,0 +1,25 @@
+module "drbd_node" {
+  source                 = "./modules/drbd_node"
+  drbd_count             = var.drbd_enabled == true ? 2 : 0
+  resource_group_name    = azurerm_resource_group.myrg.name
+  network_subnet_id      = azurerm_subnet.mysubnet.id
+  sec_group_id           = azurerm_network_security_group.mysecgroup.id
+  availability_set_id    = azurerm_availability_set.myas.id
+  storage_account        = azurerm_storage_account.mytfstorageacc.primary_blob_endpoint
+  loadbalancer_id        = azurerm_lb.mylb.id
+  public_key_location    = var.public_key_location
```
MalloZup: I was thinking that we have lot of variable here.

I think we could `hide` some of them by setting the default to the vars.

E.g  provisioner and other things .

azure/salt_provisioner.tf
```diff
@@ -192,6 +192,7 @@ host_ips: [${join(", ", formatlist("'%s'", [var.monitoring_srv_ip]))}]
 host_ip: ${var.monitoring_srv_ip}
 ha_sap_deployment_repo: ${var.ha_sap_deployment_repo}
 monitored_hosts: [${join(", ", formatlist("'%s'", var.host_ips))}]
+drbd_monitored_hosts: [${join(", ", formatlist("'%s'", var.drbd_ips))}]
```
MalloZup: do we have somthing like this in Libvirt already too? :thinking:

salt/drbd_node/iscsi_initiator.sls
```diff
@@ -0,0 +1,36 @@
+open-iscsi:
+  pkg.installed:
+  - retry:
+    attempts: 3
+    interval: 15
+
+/etc/iscsi/initiatorname.iscsi:
+  file.replace:
+    - pattern: "^InitiatorName=.*"
+    - repl: "InitiatorName=iqn.drbd.{{ grains['server_id'] }}.suse.qa"
```
MalloZup: is this not hardcoded? :thinking:

salt/drbd_node/monitoring.sls
```diff
@@ -0,0 +1,39 @@
+{% set repository = 'SLE_'~grains['osrelease_info'][0] %}
```
MalloZup: regarding this state , to me looks really duplicata in codebase.

I think somehow we should be able to apply the same states  to `hananode` and `drbdnode`perhaps and centralize them. Otherwise we need to change in sync 2 files and we will forget it.

azure/main.tf
arbulu89: I don't really understand that. 
All of the data is coming from the tfvars or default variables. If we remove any of them, it won't be configured with the custom options and it will always use the default data from `modules/drbd_node/variables.tf`.
In fact, we do the same in libvirt, what makes sense in my opinion

salt/drbd_node/iscsi_initiator.sls
arbulu89: It is, and it always have been XD
We just set the server id (unique integer) to make the initiator name unique

salt/drbd_node/monitoring.sls
arbulu89: Yes, it is duplicated. And actually I don't like it.
I pretend to update this in the next PR.
As you proposed, I would like to create other salt group like `cluster_nodes` where all the salt states applying to just nodes that will host a cluster (monitoring, ssh, iscsi_initiator, etc).

By now, I didn't want to add more complexity to this PR

PR #274, opened 2019-11-25
Feature/drbd monitoring improvements

arbulu89:
Follow up of https://github.com/SUSE/ha-sap-terraform-deployments/pull/269

- Improve salt state duplications. `cluster_node` salt state created to gather all of the states related to cluster nodes (hana, drbd, netweaver)

- Add drbd and netweaver monitoring jobs to prometheus

- Add some documentation

- Some updates in all `variables.tf` files to remove deprecated options (type value into quotes)


salt/drbd_node/drbdsetup_link.sls
```diff
@@ -0,0 +1,4 @@
+# Workaround to make drbdsetup available among different SLE versions
+/usr/sbin/drbdsetup:
+  file.symlink:
+    - target: /sbin/drbdsetup
```
stefanotorresi: this is won't be needed anymore as soon as we release next ha_cluster_exporter version, which will happen before the end of the current week, so I'd suggest to skip it!

arbulu89: Removed!

libvirt/modules/netweaver_node/salt_provisioner.tf
```diff
@@ -51,7 +51,9 @@ role: netweaver_node
 netweaver_inst_media: ${var.netweaver_inst_media}
 netweaver_nfs_share: ${var.netweaver_nfs_share}
 ha_sap_deployment_repo: ${var.ha_sap_deployment_repo}
+shared_storage_type: shared-disk
```
diegoakechi: @arbulu89 Is this hardcoded on purpose?

arbulu89: Yes...
By now the code is not adapted to use `sbd` with a disk provided by `iscsi` in Netweaver cluster. The shared disk is always created.
This is other improvement that is pending. But the whole iscsi usage needs a big refinement

PR #276, opened 2019-11-26
Feature/azure impl loadbalance

arbulu89:
Implement a new Load Balancer for Azure following the deployment guide.
Besides that, fix some variables usage.

This PR is carrying over some commits from https://github.com/SUSE/ha-sap-terraform-deployments/pull/274

So only, the last commit is different, but I preferred to create a new PR as the other was about to close

### DIFF
--- a/azure/modules/drbd_node/main.tf
+++ b/azure/modules/drbd_node/main.tf
@@ -165,12 +165,12 @@ resource "azurerm_image" "drbd-image" {
 
 resource "azurerm_virtual_machine" "drbd" {
   count                 = var.drbd_count
-  name                  = "drbd-vm-${count.index}"
+  name                  = "drbd0${count.index + 1}"
   location              = var.az_region
   resource_group_name   = var.resource_group_name
   network_interface_ids = [element(azurerm_network_interface.drbd.*.id, count.index)]
   availability_set_id   = azurerm_availability_set.drbd-availability-set[0].id
-  vm_size               = "Standard_D2s_v3"
+  vm_size               = var.instancetype
 
   storage_os_disk {
     name              = "drbd-os-disk-${count.index}"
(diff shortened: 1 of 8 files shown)


azure/modules/drbd_node/main.tf
```diff
@@ -134,11 +165,11 @@ resource "azurerm_image" "drbd-image" {
...
 resource "azurerm_virtual_machine" "drbd" {
   count                 = var.drbd_count
-  name                  = "drbd-${count.index}"
+  name                  = "drbd-vm-${count.index}"
   location              = var.az_region
   resource_group_name   = var.resource_group_name
   network_interface_ids = [element(azurerm_network_interface.drbd.*.id, count.index)]
-  availability_set_id   = var.availability_set_id
+  availability_set_id   = azurerm_availability_set.drbd-availability-set[0].id
   vm_size               = "Standard_D2s_v3"
```
diegoakechi: @arbulu89 I think will be natural to choose the type for the DRBD machine. I would default to that, but give the option to change. It can be done in a follow-up PR.

azure/modules/drbd_node/salt_provisioner.tf
```diff
@@ -34,8 +34,8 @@ resource "null_resource" "drbd_provisioner" {
     content     = <<EOF
 provider: azure
 role: drbd_node
-name_prefix: ${terraform.workspace}-drbd
-hostname: ${terraform.workspace}-drbd0${count.index + 1}
+name_prefix: drbd
```
diegoakechi: @arbulu89 Won't it break the deployment, in case we want 2 different workspaces at the same time, making all the machines to have the same hostname?

arbulu89: This is just the internal hostname of the machine. It won't have any conflict with other deployments.
The unique reason we did this change it was the GCP requires to have the same VM name and hostname to network properly, other than than is just cosmetics

PR #280, opened 2019-12-02
Update system during deployment

varkoly:


arbulu89:
@varkoly And other important thing. The update cannot be made if `qa_mode` is set. @ldevulder , @juadk Could you confirm this?

varkoly:
@arbulu89  That is why I've pushed the update first into the registration.sls. Update makes only sence if registration was done. Now I'll build in the same conditions like there.

MalloZup:
when merging this, use the squash commits option :pray:

salt/default/pkgs.sls
```diff
@@ -7,4 +7,5 @@ install_additional_packages:
 {% endfor %}
     - require:
       - sls: default.repos
+      - sls: default.update
```
arbulu89: I wouldn't personally add this entry. Actually, I would get rid of tis `pkgs` state. Nobody uses it.

varkoly:
> @varkoly @diegoakechi It looks good to me. As far as I see, it shouldn't bring any issue. Have you tested? At least in libvirt?

I'm just doing it. I'll report if the test has worked.

diegoakechi:
@varkoly This PR should be done to the `develop` branch and not directly to `master`.

varkoly:
I'll resubmit it into devel branch after I've tested IT.

PR #281, opened 2019-12-03
Fix dashboard

### LINKED ISSUE #278: DRBD panel data shown in all the cluster
DRBD panel data shown in all the cluster, so it's not filtered by job.
We should filter that to don't create confusion.

And we can just update the headers to start with upper case

image

MalloZup:
fix #278 

I tested this work fine.


MalloZup:
@stefanotorresi thx done!

MalloZup:
f

PR #282, opened 2019-12-04
Update system during deployment

varkoly:


salt/default/update.sls
```diff
@@ -0,0 +1,10 @@
+{% if not grains.get('qa_mode') %}
+{% if grains['os_family'] == 'Suse' %}
+update_repos:
```
diegoakechi: @varkoly just a nitpick, but I think `update_system_packages` would be more clear than `update_repos`, since we are not refreshing the repos, but updating the packages installed on the system here.

varkoly:
I've run test with libvirt on SLE15-SP1 and SLE15-SP0 and it works.

salt/default/update.sls
varkoly: Renamed.

diegoakechi:
@arbulu89 @MalloZup @stefanotorresi  When we have discussed about that, we had the idea of making this optional, since could have reasons to fully patch the system or not. if is it still the case, I think this PR should also include a new parameter to allow that. What do you think?

stefanotorresi:
hmm, to be honest, I don't exactly fancy this idea. Would be the reason for not updating the system? Can't we just keep it simple and always update it?

diegoakechi:
@stefanotorresi That was my suggestion, just being fair and not ignoring @MalloZup and @arbulu89 raised points ;)

salt/default/update.sls
```diff
@@ -0,0 +1,10 @@
+{% if not grains.get('qa_mode') %}
+{% if grains['os_family'] == 'Suse' %}
+update_system_packages:
+  cmd.run:
+    - name: zypper --non-interactive --gpg-auto-import-keys update
```
diegoakechi: @varkoly I just run 2 tests on this and I am facing a conflict during the upgrade coming from OBS packages versus IBS packages. Nothing wrong with your code, but as the recommended packages are installed by default, we are pulling packages from OBS and causing this conflict. To be on the safe side, I suggest we upgrade only the required packages instead. This is also a good practice.
```suggestion
    - name: zypper --non-interactive --gpg-auto-import-keys update --no-recommends
```

PR #283, opened 2019-12-04
Remove AvSet from Monitoring and iSCSI node, updating naming schema

steffenv-msft:
Individual instances should not be in Availability Set, and never share Availability Set with HA/Clustered instances.


azure/instances.tf
```diff
@@ -72,19 +71,19 @@ resource "azurerm_virtual_machine" "iscsisrv" {
   }
 }
 
-# Cluster Nodes
-
+# Hana Cluster Nodes
+#
 resource "azurerm_virtual_machine" "clusternodes" {
```
arbulu89: I would change this `cluternodes` to `hana`

diegoakechi:
It has worked in my tests, so I am merging it. @steffenv-msft thanks for the changes, looks much nicer now!

PR #284, opened 2019-12-05
Feature/netweaver azure

arbulu89:
Netweaver implementation for Azure.

We need to first merge
https://github.com/SUSE/ha-sap-terraform-deployments/pull/283
https://github.com/SUSE/sapnwbootstrap-formula/pull/22


azure/modules/netweaver_node/main.tf
```diff
@@ -0,0 +1,429 @@
...
+    workspace = terraform.workspace
+  }
+}
+
+resource "azurerm_network_interface" "netweaver" {
+  count                     = var.netweaver_count
+  name                      = "nic-netweaver0${count.index + 1}"
+  location                  = var.az_region
+  resource_group_name       = var.resource_group_name
+  network_security_group_id = var.sec_group_id
```
diegoakechi: @arbulu89  I think we need to include the option `enable_accelerated_networking = true`  similar to PR https://github.com/SUSE/ha-sap-terraform-deployments/pull/292

https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/dbms_guide_general#azure-accelerated-networking

azure/instances.tf
```diff
@@ -22,7 +22,7 @@ resource "azurerm_virtual_machine" "iscsisrv" {
   resource_group_name   = azurerm_resource_group.myrg.name
   network_interface_ids = [azurerm_network_interface.iscsisrv.id]
   availability_set_id   = azurerm_availability_set.myas.id
-  vm_size               = "Standard_D2s_v3"
+  vm_size               = "Standard_B1ms"
```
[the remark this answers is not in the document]
arbulu89: I updated that because in our particular case we don't have enough `Standard_D2s_v3` machines in the US regions. Actually, we don't have any recommendation, so it's hard to say...
Anyway, this machine only hosts the iscsi server, so I guess a small machine would be enough

azure/modules/netweaver_node/main.tf
arbulu89: Only in Netweaver nodes, or should be add it in drbd nodes too?
This option is not supported in all VM types, so it might create issues...

PR #289, opened 2019-12-06
Change the default HANA scraping time to 30s

diegoakechi:
Collecting metrics every 5s seconds can put unnecessary load into the
database

### DIFF
--- a/salt/monitoring/prometheus/prometheus.yml
+++ b/salt/monitoring/prometheus/prometheus.yml
@@ -17,6 +17,8 @@ scrape_configs:
   # in a multi-cluster scenario, add another job name called
   #  hacluster-02
   - job_name: hana-cluster
+    # The HANA scrapping follows a different scrapping time to reduce the execution load into the database
+    # This time was based on users feedback, but should be set accordingly with your environment needs.
     scrape_interval: 30s
     scrape_timeout: 30s
     static_configs:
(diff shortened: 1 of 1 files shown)


witekest:
Is there a reason for not making it global?

diegoakechi:
@witekest Not apart of changing behavior as less as possible, but now that you have mentioned, I think it can be a good idea, since the SAPHanaSR resource agent monitoring operation is set to 60 seconds, and some other cluster operations are set to 20s, so not much will go on every 5 seconds. What do you think if I change the global config to 30 seconds? @arbulu89, @MalloZup @stefanotorresi @gao-yan?

stefanotorresi:
Do we have any load statistics backing up the assumption that 5 seconds is too frequent?

diegoakechi:
@stefanotorresi Yes, on my tests, even with an empty HANA the collection is having some gaps... Also 30s is what some of the customers and partners are using on their monitoring configuration.

stefanotorresi:
fair enough!

salt/monitoring/prometheus/prometheus.yml
```diff
@@ -17,6 +17,8 @@ scrape_configs:
   # in a multi-cluster scenario, add another job name called
   #  hacluster-02
   - job_name: hana-cluster
+    scrape_interval: 30s
```
MalloZup: can we add a comment here why we are putting/increasing this timeout?

like the motivation you stated in pr. THis will help us to know why we increased.

MalloZup:
@diegoakechi  need to update your PR branch to latest develop. A part that feel free to merge afterwards thx!

PR #291, opened 2019-12-10
Fix netweaver hostname

stefanotorresi:
This PR removes the `hostname` parameter from all the libvirt network interfaces, because it doesn't seem to have any actual effect, and fixes the hostname of the NetWeaver nodes which wasn't correctly interpolated in the salt provisioner grains.

### DIFF
--- a/libvirt/modules/netweaver_node/salt_provisioner.tf
+++ b/libvirt/modules/netweaver_node/salt_provisioner.tf
@@ -36,7 +36,7 @@ resource "null_resource" "netweaver_node_provisioner" {
   provisioner "file" {
 content = <<EOF
 name_prefix: ${var.name}
-hostname: ${var.name}${var.netweaver_count > 1 ? "0${count.index + 1}" : ""}
+hostname: ${terraform.workspace}-${var.name}${var.netweaver_count > 1 ? "0${count.index + 1}" : ""}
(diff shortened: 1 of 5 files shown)


libvirt/modules/netweaver_node/salt_provisioner.tf
```diff
@@ -36,7 +36,7 @@ resource "null_resource" "netweaver_node_provisioner" {
   provisioner "file" {
 content = <<EOF
 name_prefix: ${var.name}
-hostname: ${var.name}${var.netweaver_count > 1 ? "0${count.index + 1}" : ""}
+hostname: ${terraform.workspace}-${var.name}${var.netweaver_count > 1 ? "0${count.index + 1}" : ""}
```
arbulu89: I understand that you set this value to align with the other nodes, but we cannot really do this because NW has a quite strict character limit for the hostname (12 or 15 characters). Usually this exceeds the hostname string if we add the `workspace`.
The unique reason we added the `workspace` to the hostname is because `gcp` requires that the vm name and hostname must match to make the network work (I have not tested this personally, was changed by the QA team).
Actually, in azure, we have removed that...
I would like to have a better solution, but so far, I have not worked to find one

stefanotorresi:
superseded by #471

PR #292, opened 2019-12-10
Fix LB rules to match SAP instance 00

steffenv-msft:
Since the template deploys a HANA DB instance 00, HANA ports need to follow the instance number: 3xx13, 3xx14, 3xx15, 3xx17 where xx is the instance number. Same for health probe, which is 625xx. 
Rules in the template were originally for instance 03.
Also updated to use Accelerated Networking for HANA nodes (https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli)

### DIFF
--- a/azure/network.tf
+++ b/azure/network.tf
@@ -234,11 +234,11 @@ resource "azurerm_public_ip" "iscsisrv" {
 }
 
 resource "azurerm_network_interface" "hana" {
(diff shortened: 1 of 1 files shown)


diegoakechi:
@steffenv-msft I could fix it on your branch, as you have allowed maintainers contributions.

PR #297, opened 2019-12-11
Move some Netweaver and HANA OS and Network configurations to variables

diegoakechi:
Remove some Netweaver and HANA hardcodes and move the options to variables to give the possibility to the user to set them.

All of the variables assume as default the values that were hardcoded before.

### DIFF
--- a/azure/modules/netweaver_node/variables.tf
+++ b/azure/modules/netweaver_node/variables.tf
@@ -35,6 +35,7 @@ variable "data_disk_type" {
 }
 
 variable "data_disk_size" {
+  description = "Size of the Netweaver data disks, informed in GB"
   type    = string
   default = "60"
 }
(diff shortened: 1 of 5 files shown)


azure/modules/netweaver_node/variables.tf
```diff
@@ -24,6 +24,26 @@ variable "netweaver_count" {
...
+  default = "Standard_D4s_v3"
+}
+
+variable "data_disk_type" {
+  type    = string
+  default = "Standard_LRS"
+}
+
+variable "data_disk_size" {
+  type    = string
```
arbulu89: I would add a description here to say that the size is in GBs

azure/modules/netweaver_node/variables.tf
```diff
@@ -68,6 +88,11 @@ variable "storage_account_path" {
   type        = string
 }
 
+variable "enable_accelerated_networking" {
+  description = "Enable accelerated networking for netweaver. This function is mandatory for certified Netweaver environments and are not available for all kinds of instances. Check https://docs.microsoft.com/en-us/azure/virtual-network/create-vm-accelerated-networking-cli for more details"
+  default     = true
```
arbulu89: Add `type == boolean` just in case

PR #305, opened 2019-12-19
Initial implementation of Multi-Cluster Grafana dashboard

stefanotorresi:
This is mostly what I've shown in my progress report during the latest sprint review.

I've included some new time charts and fixed a couple of broken queries.

Here is a screenshot:
Screenshot from 2019-12-19 13-52-35

Live demo is here: http://10.162.30.194/d/5b0v3y-Wz1/multi-cluster-overview

It's not 100% incomplete, but I'd rather merge this first version before continuing the work.

### DIFF
--- a/salt/monitoring/provisioning/dashboards/multi-cluster-overview.json
+++ b/salt/monitoring/provisioning/dashboards/multi-cluster-overview.json
@@ -49,19 +49,17 @@
       },
       "id": 6,
       "panels": [],
-      "title": "Cluster Nodes",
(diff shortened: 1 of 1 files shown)


stefanotorresi:
> one improvement I see is that the legend of the top panels are based on the Cluster name (Job Name) but the metrics is shown on the node level, so I think some node name identifier should be used on the legend.

yep, absolutely, that's on my to-do list. It requires some nasty PromQL-fu, but I'll get it done.

PR #310, opened 2020-01-09
Add the option to use ssh keys to create the cluster in libvirt

arbulu89:
Add the usage of ssh keys to create the cluster in libvirt as we do in the cloud providers

### DIFF
--- a/libvirt/main.tf
+++ b/libvirt/main.tf
@@ -77,6 +77,8 @@ module "hana_node" {
   reg_code               = var.reg_code
   reg_email              = var.reg_email
   reg_additional_modules = var.reg_additional_modules
+  cluster_ssh_pub        = var.cluster_ssh_pub
+  cluster_ssh_key        = var.cluster_ssh_key
   ha_sap_deployment_repo = var.ha_sap_deployment_repo
   qa_mode                = var.qa_mode
   hwcct                  = var.hwcct
@@ -111,6 +113,8 @@ module "drbd_node" {
   reg_code               = var.reg_code
   reg_email              = var.reg_email
   reg_additional_modules = var.reg_additional_modules
+  cluster_ssh_pub        = var.cluster_ssh_pub
+  cluster_ssh_key        = var.cluster_ssh_key
   ha_sap_deployment_repo = var.ha_sap_deployment_repo
   devel_mode             = var.devel_mode
   provisioner            = var.provisioner
@@ -170,6 +174,8 @@ module "netweaver_node" {
   reg_code               = var.reg_code
   reg_email              = var.reg_email
   reg_additional_modules = var.reg_additional_modules
+  cluster_ssh_pub        = var.cluster_ssh_pub
+  cluster_ssh_key        = var.cluster_ssh_key
   ha_sap_deployment_repo = var.ha_sap_deployment_repo
   provisioner            = var.provisioner
   background             = var.background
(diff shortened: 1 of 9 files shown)


arbulu89:
> I find that this overlaps with the already present `public_key_location` tfvar which is injected in the `authorized_keys` salt grain; I think we should either clarify the difference in the documentation or remove it altogether.

First of all, this change just normalize how we use this parameters and keys as they are used in the cloud providers, so if we decide to change we will most probably need to change in all of them.

Now,

The 2 key pairs are used for different purposes.
The keys in `public_key_location` are used to give access to the VM itself.
This `cluster_ssh_pub` keys are used to give HA bootstrap formula (and `crmsh`) the option to create the cluster without asking for password (this second ssh keys must be passphraseless).

We have some options here:
- Try to remove `cluster_ssh_pub` pair key at all and use only 1 key pair to access the VM and create the cluster (I don't know if this is feasible as we need to upload some dummy ssh key private key to the machines, and I woulnd't like people uploading their own ssh keys by accident).
- Obviously, document better how they work (as in the cloud).

By now, I would vote for the documentation option, as any other change requires much more work (among all the providers).
What do you think @stefanotorresi ?

stefanotorresi:
I suspected they served a different purpose, but that's still not very clear to me.

I guess we have to better document all of these, because right now they are very confusing. Do you want to do this in another PR that touches all the providers?

stefanotorresi:
Now that I look even deeper into it, I can point out a lot of problems with the current documentation of these variables:

I take the Azure provider for example, but I guess the same goes for all the others:
1. in the readme, we first tell the user to "Generate private and public keys for the cluster nodes" inside a specific folder, and then to have `cluster_ssh_pub` and `cluster_ssh_key` vars match just the file names? But in the tfvars file, we say that these two values are paths, not just file names.
2. Both `private_key_location` and `public_key_location` are said to be local paths to keys that are used to access the virtual machines. If the previous ones are copied in all the vms, couldn't those _also_ be used to access?
3. It is nowhere indicated that the path indicated in `cluster_ssh_key` must point to a passphrase-less private key.

MalloZup:
In the past I have also stumbled why we create so many ssh keys. 
IMHO we shouldn't have much vars and ssh keys.


I think we need a follow-up issue for this trying to fix this problem without introducing vars or reusing the same ssh keys from users

e.g,  some examples:
https://github.com/uyuni-project/sumaform/blob/487230a19dcc0f4cd1c0e8bf75122e8213096942/modules/libvirt/locust/main.tf#L5 
https://github.com/uyuni-project/sumaform/blob/master/README_ADVANCED.md#custom-ssh-keys

arbulu89:
@MalloZup @stefanotorresi 
Here is why we have this 2 ssh key pairs.
Libvirt and cloud providers are a bit different in their creation. In libvirt the machines are created with password access (`linux` by default`) while in the cloud providers not (I'm not 100% sure, but the reason behind that is that the cloud images root user comes without password, so if we enable the ssh access with password (what is possible), it would a safety issue).

So, in libvirt deployment, we were taking advantage of this password usage, and using it to create the cluster (the joining node needs to access without password to the init node). The ha bootstrap formula gives the option to set the password and use (otherwise, ssh keys authorization is mandatory).

So, in the cloud providers, in order to enable this cluster creation, we need to upload a ssh key pair (public and private keys) by force. Obviously, we don't want to upload our personal keys, that's why a "dummy" key pair creation is explained and used.

The unique way to remove one of the keypairs would be to use the `public_key_location` with the same ssh keys created for the cluster, but I don't really like the idea, I personally prefer the to have the flexibility to use the keys we want (what we could do is to have default values for this variables pointing to /home/$user/.ssh/id_rsa and /home/$user/.ssh/ida_rsa.pub).

arbulu89:
All that being said, it's true that this PR is not strictly need, it was just created to add some functionalities that are available in the cloud providers and not in libvirt.

MalloZup:
imho we can close this

arbulu89:
> imho we can close this

Why do you want to close it? I think having this feature makes sense to test different use cases (crmsh with -no-overwrite-ssh for example). It lacks some documentation that's it.

stefanotorresi:
@arbulu89 if you could add some even rudimentary docs on this feature, I'd be happy to merge it!

arbulu89:
Closing as it is stalled

PR #313, opened 2020-01-10
Update the iscsi-formula usage to work with v1.1.1

arbulu89:
Code updated to work with the v1.1.1 iscsi-formula:
https://github.com/saltstack-formulas/iscsi-formula
It had a bug so I have submitted a patch in a new PR here:
https://github.com/saltstack-formulas/iscsi-formula/pull/32

Tested and working. The package in https://build.opensuse.org/project/show/network:ha-clustering:Unstable must be updated together with this PR, and eventually the factory repository to match with the master branch

@juadk With this change all the workarounds for qa are not needed anymore. Basically, I have replaced the old `saveconfig.json` file in the new `iscsi_srv_qa.sls` file replicating the data. Have a look there please


arbulu89:
@juadk @nick-wang Any thoughts on this?

juadk:
> @juadk @nick-wang Any thoughts on this?

Yes, I plan to test it asap, sorry for the delay.

arbulu89:
> LGTM, tested in AWS (not cloud provider specific) with 15SP1 and QA_MODE.
> I used iscsi-formula-1.1.1 and salt-formulas-configuration-0.0.2 from your [home project](https://download.opensuse.org/repositories/home:/xarbulu:/sap-deployment/SLE_15_SP1)
> Just one small comment, I wonder if we couldn't use only one pillar files. Looks like the difference is just two more devices in the qa one. Could be unify later.

Yes, we can do that. I don't really know what for you use this qa specific configuration. In our case, we need 3 different disks to use sbd with hana, drbd and netweaver. That's why we have this configuration.
We can unify though in a next PR.

nick-wang:
> Just one small comment, I wonder if we couldn't use only one pillar files. Looks like the difference is just two more devices in the qa one. Could be unify later.

If only devices number is important, i think we can remove the `qa` configurations.
@juadk @arbulu89  Please help to check & review 
https://github.com/SUSE/ha-sap-terraform-deployments/pull/325

Issue #301, opened 2019-12-18
Netweaver Cluster is wrongly configured on Prometheus


diegoakechi:
All the 4 machines for the Netweaver cluster are being included as part of the netweaver-cluster job inside Prometheus configuration. 

The job_name is used to group the cluster nodes from the monitoring perspective.

As on Netweaver, only the ERS and ASCS hosts are part and maintained by pacemaker cluster, having the PAS and AAS instances inside the configuration causes distortion on the metrics/dashboards. 

The ERS and ASCS should be included as part of the job_name clusters, and the AAS and PAS could be included in a second job_name, called as `netweaver-appservers` as a suggestion.

stefanotorresi:
I don't think we need an additional `job_name`, because this bears no semantic in relation to Pacemaker. On the contrary, it generically identifies a set of targets, and not all of these are `ha_cluster_exporter` instances. In the case of Netweaver, we'll have 4 instances of `node_exporter` and just 2 of `ha_cluster_exporter`.
What we need to change is to only set the `ha_cluster_exporter` targets for the nodes where it actually has to run.

PR #314, opened 2020-01-13
Feature/filter dashboard nodes

arbulu89:
Filter properly the netweaver nodes in the grafana dashboard to only show the cluster members.
Besides that
- I have updated the HANA dashboard to only have the option to select nodes with a HANA db installed and filter properly the other multicheck boxes.
- I have fixed the systemd units filter to have more precision (otherwise some systemd units like fsck with some files with hana in their name were shown).

As a comment, the `node_exporter` data is shown only for the nodes in the HA cluster. If we want to keep the 4 netweaver cluster in the panel and only show the HA data for the nodes having this exporter we will need to do changes.

https://github.com/SUSE/ha-sap-terraform-deployments/issues/301

@stefanotorresi @MalloZup Temporary dashboard available in http://10.162.31.160

PD: And finally, a small bugfix for qa_mode in `repos.sls`


MalloZup:
@arbulu89 thx for pr can we have a screenshot?

arbulu89:
image
image
image

stefanotorresi:
> As a comment, the `node_exporter` data is shown only for the nodes in the HA cluster. If we want to keep the 4 netweaver cluster in the panel and only show the HA data for the nodes having this exporter we will need to do changes.

I think this is how it should be. The HA Cluster dashboard is supposed to show details about a Pacemaker cluster, and only two of the Netweaver nodes apply.

arbulu89:
After internal discussion with @stefanotorresi , we have agreed that the current solution is not what we want. We want to still have the `node_exporter` data in the multi cluster dashboard, and only filter the nodes in the detailed HA dashboad, to avoid having non HA nodes to be selected.

arbulu89:
@stefanotorresi I have added the changes for:
1. Only show the nodes if the export HA metrics in the `Node` multi select option
2. Fix a pattern in the `Resource` table that was adding more metrics than the expected ones

As commented, the unique drawback is that if the ha exporter is not working, we won't see and choose the nodes (what I think is good) in that multi select.

PR #325, opened 2020-01-19
iscsi_srv: use flexible pillar file to reduce duplication.

nick-wang:
Replace the iscsi pillar file with flexible partitions retrieve from grains file.

The current design is using 3 out of 5 partitions,` qa_mode` using the whole 5 partitions. 
Could remove the *iscsi_srv_qa.sls*, If the `alua_tpgs` and `attributes` parameters it not important in `qa_mode`.


nick-wang:
@juadk Could you help to confirm besides partitions number, is the difference of `alua_tpgs` and `attributes` necessary in `qa_mode`?
If not, we can remove the qa specific configuration and condition.
(set to `WIP` due to the question)

salt/pillar/iscsi_srv.sls
```diff
@@ -1,3 +1,10 @@
+{% set devicenum = 'abcdefghijklmnopqrstuvwxyz' %}
+{% set partitions = salt['grains.get']('partitions') %}
```
arbulu89: We could just have `grains['partitions']` here.

salt/pillar/iscsi_srv.sls
```diff
@@ -1,3 +1,10 @@
...
+
+{% if not grains.get('qa_mode', False) %}
+{% set partitions = (partitions|list)[:3] %}
```
arbulu89: Why this `[:3]` hardcoded? It should represent what we have in the partitions entry, right?
If we have 5 partitions, we should offer 5 partitions even though we are not running `qa_mode` in my opinion.

Other option would be to expose other parameter in the terraform part to select the device number to offer.

juadk:
> @juadk Could you help to confirm besides partitions number, is the difference of `alua_tpgs` and `attributes` necessary in `qa_mode`?
> If not, we can remove the qa specific configuration and condition.
> (set to `WIP` due to the question, though the changed pillar file is ready to use/test.)

AFAIK, only partitions number is the difference which matter. `alua_tpgs` and `attributes` are used because there are in the iscsi pillar example. I will give a try with your new pillar file.

nick-wang:
> By now, waiting to the review of some QA member.
> I have added a couple of comments besides that. Specially, I wouldn't hardcode to 3 the number of disk offered.

I just follow the current design. To be honest, i don't know why divide into 5 partitions but only expose 3 by default.
I could change if we get a consensus here.

nick-wang:
> > @juadk Could you help to confirm besides partitions number, is the difference of `alua_tpgs` and `attributes` necessary in `qa_mode`?
> > If not, we can remove the qa specific configuration and condition.
> > (set to `WIP` due to the question, though the changed pillar file is ready to use/test.)
> 
> AFAIK, only partitions number is the difference which matter. `alua_tpgs` and `attributes` are used because there are in the iscsi pillar example. I will give a try with your new pillar file.

Great! So we could drop one of the pillar file without different `alua_tpgs` and `attributes`.

salt/pillar/iscsi_srv.sls
juadk: > Other option would be to expose other parameter in the terraform part to select the device number to offer.
I think it's good idea too.

nick-wang:
@juadk @arbulu89  Is 5 partitions still needed in `qa_mode`?

If so, I am thinking we could expose all configured *partitions* by default. Add another parameters in terraform for the numbers of *partitions* needed in environment. 
For example, the current usage is 3 out of 5. (0 or `bigger than all partitions` will be treated as `all`)
How do you think of this? Or a better idea.

juadk:
> @juadk @arbulu89 Is 5 partitions still needed in `qa_mode`?
> 
> If so, I am thinking we could expose all configured _partitions_ by default. Add another parameters in terraform for the numbers of _partitions_ needed in environment.
> For example, the current usage is 3 out of 5. (0 or `bigger than all partitions` will be treated as `all`)
> How do you think of this? Or a better idea.

Actually, 3 are needed for the moment, one for sbd and two for drbd. But it could possible to add another test that requires more devices later.

nick-wang:
> Actually, 3 are needed for the moment, one for sbd and two for drbd. But it could possible to add another test that requires more devices later.

Got it. I could leave the 5 in `grains` and expose them all by default. 
With another terraform option if need limited partitions attached to iscsi server.(though the needs is weird as well.)

libvirt/main.tf
```diff
@@ -33,6 +33,7 @@ module "iscsi_server" {
   source                 = "./modules/iscsi_server"
   iscsi_count            = var.shared_storage_type == "iscsi" ? 1 : 0
   iscsi_image            = var.iscsi_image
+  iscsi_disks            = var.iscsi_disks
```
juadk: why didn't you update other providers? I only see changes for libvirt here.

nick-wang: I forget this...

nick-wang: Fix in the new commit. Thanks for the review.

nick-wang:
> @nick-wang
> I think the variable `iscsi_disks` is missing in the iscsi module variables file. I don't know if this raises a warning at least.
> I would add it, as the variable is in fact iscsi thing.
> 
> Finally, you have updated documentation only for gcp and aws. I would include the variable in Azure and Libvirt too

Fix in the rebased commit.

juadk:
> Besides that, and if @juadk agrees, I would remove the 5 disks creation to 3 disks creation. It makes more sense. If we need more we can always change, but having them "just in case" is something that I would prefer avoiding.

Yes, I agree for keeping 3 disks.

nick-wang:
@juadk @arbulu89 Thanks for the review.

I changed the iscsi partitions from 5 to 3 in libvirt/gcp/aws/azure in the [latest commit](https://github.com/SUSE/ha-sap-terraform-deployments/pull/325/commits/bf7b6a474456241605524229add947a263f639d0)
```
partitions:
  1:
    start: 1
    end: 33%
  2:
    start: 33%
    end: 67%
  3:
    start: 67%
    end: 100%
```
`PASSED` on my test against *libvirt*. If none of you object to this, i will merge it tomorrow.

juadk:
> `PASSED` on my test against _libvirt_. If none of you object to this, i will merge it tomorrow.

You are welcome Nick, no objection for merging this one.

nick-wang:
Thank you all for review!

PR #326, opened 2020-01-22
Create new terraform module "hana_node"

Simranpal:
Creating a new terraform module to seperate the logic for hana_node


Simranpal:
I have two points to discuss:

- Does it make sense to move the `sles4sap` resource into the `hana_node` module as I did here

- Here I removed the **`boot_diagnostics{storage_uri}`** from hana vm `"azurerm_virtual_machine" "hana"` as it required its resource `"azurerm_storage_account.mytfstorageacc"` to defined inside the `hana_node` module to get access. Is there a way to access a terraform resource (for example access a resource inside resources.tf) from `hana_node` module

arbulu89:
> I have two points to discuss:
> 
>     * Does it make sense to move the `sles4sap` resource into the `hana_node` module as I did here
> 
>     * Here I removed the **`boot_diagnostics{storage_uri}`** from hana vm `"azurerm_virtual_machine" "hana"` as it required its resource `"azurerm_storage_account.mytfstorageacc"` to defined inside the `hana_node` module to get access. Is there a way to access a terraform resource (for example access a resource inside resources.tf) from `hana_node` module

1. I think it does. What you did make sense. What I don't really know yet if we should keep the used variales in the image.tf or move them to the variables.tf file. Maybe it would make sense to keep them as they are to not overload the variables.tf file, and eventually change the name to image_variables.tf

2. I don't really know, but I guess we could access to that resource somehow using `data` element. This needs a bit more of research.

When you have the final version, just remove the `WIP` tag. Test in Azure and if it works I will give a final deep review, but in this first check it looks pretty good!

arbulu89:
@Simranpal Talking about the `boot_diagnostic`, you will need to create a new variable in the `hana_node` as we do with for example with `resource_group_name           = azurerm_resource_group.myrg.name`.
And you will need to pass this value to the module from the main file. This way it will be available in the module. We cannot remove it.

azure/modules/hana_node/main.tf
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  tags = {
+    workspace = terraform.workspace
+  }
+}
+
+resource "azurerm_public_ip" "hana" {
+  count                   = var.ninstances
+  name                    = "pip-hana0${var.ninstances > 1 ? "0${count.index + 1}" : ""}"
```
arbulu89: Replace this `hana0` by ${var.name}

azure/modules/hana_node/variables.tf
```diff
@@ -0,0 +1,215 @@
...
+variable "network_subnet_id" {
+  type = string
+}
+
+variable "name" {
+  type        = string
+  default     = "hana"
+}
+
+variable "ninstances" {
```
arbulu89: Use only one between `ninstances` and `hana_count`. I would use the 2nd one

azure/modules/hana_node/variables.tf
```diff
@@ -0,0 +1,215 @@
...
+variable "host_ips" {
+  description = "ip addresses to set to the nodes"
+  type        = list(string)
+}
+
+variable "instancetype" {
+  type    = string
+}
+
+variable "data_disk_type" {
```
arbulu89: I think this should be `hana_data_disk_type`, right? The same for `hana_data_disk_size`

azure/modules/hana_node/variables.tf
```diff
@@ -0,0 +1,215 @@
+variable "az_region" {
```
arbulu89: The variables are not really properly ordered in my opinion, try to group them depending on their usage

azure/main.tf
```diff
@@ -68,3 +68,47 @@ module "netweaver_node" {
   background                    = var.background
   monitoring_enabled            = var.monitoring_enabled
 }
+
+module "hana_node" {
```
arbulu89: You need to add here the `hana_count` variables. Otherwise it will be always 2. And the idea of having this parameter is to customize it (to have a single hana deployment for example). Now it uses the default 2 values always (from the module variables files)

azure/main.tf
```diff
@@ -68,3 +68,47 @@ module "netweaver_node" {
...
+  source                        = "./modules/hana_node"
+  az_region                     = var.az_region
+  host_ips                      = var.host_ips
+  scenario_type                 = var.scenario_type
+  resource_group_name           = azurerm_resource_group.myrg.name
+  network_subnet_id             = azurerm_subnet.mysubnet.id
+  sec_group_id                  = azurerm_network_security_group.mysecgroup.id
+  storage_account               = azurerm_storage_account.mytfstorageacc.primary_blob_endpoint
```
arbulu89: This doesn't look like an appropriate name. It's not the storage account. It's the primary blob endpoint id or name. Please, use a more specific name.

azure/main.tf
```diff
@@ -68,3 +68,47 @@ module "netweaver_node" {
...
+  enable_accelerated_networking = var.hana_enable_accelerated_networking
+  sles4sap_uri                  = var.sles4sap_uri
+  init_type                     = var.init_type
+  hana_inst_master              = var.hana_inst_master
+  hana_inst_folder              = var.hana_inst_folder
+  hana_disk_device              = var.hana_disk_device
+  hana_fstype                   = var.hana_fstype
+  cluster_ssh_pub               = var.cluster_ssh_pub
+  cluster_ssh_key               = var.cluster_ssh_key
+  instancetype                  = var.instancetype
```
arbulu89: I would put this instancetype at the top as it is azure related thing. Maybe just after `region` parameter

azure/main.tf
```diff
@@ -68,3 +68,47 @@ module "netweaver_node" {
...
+  hana_public_sku               = var.hana_public_sku
+  hana_public_version           = var.hana_public_version
+  admin_user                    = var.admin_user
+  iscsi_srv_ip                  = azurerm_network_interface.iscsisrv.private_ip_address
+  reg_code                      = var.reg_code
+  reg_email                     = var.reg_email
+  reg_additional_modules        = var.reg_additional_modules
+  ha_sap_deployment_repo        = var.ha_sap_deployment_repo
+  devel_mode                    = var.devel_mode
+  qa_mode                       = var.qa_mode
```
arbulu89: Put the `qa_mode` and `hwcct` variables together (at the bottom). They are qa related and better if they go together here.

azure/modules/hana_node/main.tf
```diff
@@ -0,0 +1,289 @@
+# Availability set for the hana VMs
+
+resource "azurerm_availability_set" "hana-availability-set" {
```
arbulu89: Maybe we need to change this, to work with the `hana_count` variables. It doesn't have sense to create it if `hana_count` is 0. In the other hand, as the project is about hana, it's not totally wrong.
Let's keep it this way, just adding this as a comment.

azure/modules/hana_node/main.tf
```diff
@@ -0,0 +1,289 @@
...
+resource "azurerm_lb" "hana-load-balancer" {
+  name                = "lb-hana"
+  location            = var.az_region
+  resource_group_name = var.resource_group_name
+
+  frontend_ip_configuration {
+    name                          = "lbfe-hana"
+    subnet_id                     = var.network_subnet_id
+    private_ip_address_allocation = "static"
+    private_ip_address            = "10.74.1.200"
```
arbulu89: Eventually, we will need to make this IP customizable, but let's do it in a future PR.

azure/main.tf
Simranpal: Thanks @arbulu89 thats correct!

azure/main.tf
Simranpal: Hi @arbulu89 yes agree the name is not appropriate. This is actual used like this in all modules. Should I create a new PR along with other improvements?

azure/modules/hana_node/salt_provisioner.tf
```diff
@@ -0,0 +1,80 @@
...
+    content     = data.template_file.salt_provisioner.rendered
+    destination = "/tmp/salt_provisioner.sh"
+  }
+
+  provisioner "file" {
+    content = <<EOF
+provider: azure
+role: hana_node
+devel_mode: ${var.devel_mode}
+hana_instance_number: ${var.hana_instance_number}
```
arbulu89: Do we use this parameter for something? Why have you added it?

azure/main.tf
arbulu89: So, this was inherited from older code... Sorry, I thought you added it. Let's use this name by now and change in the future if we see any confusion.

azure/README.md
```diff
@@ -136,7 +136,7 @@ In the file [terraform.tfvars.example](terraform.tfvars.example) there are a num
 * **instancetype**: SKU to use for the cluster nodes; basically the "size" (number of vCPUS and memory) of the VM.
 * **hana_data_disk_type**: disk type to use for HANA (Standard_LRS by default).
 * **hana_data_disk_caching**: caching mode for HANA disk, could be None, ReadOnly or ReadWrite (ReadWrite by default).
-* **ninstances**: number of cluster nodes to deploy. Defaults to 2.
+* **hana_count**: number of cluster nodes to deploy. Defaults to 2.
```
arbulu89: I would change the last part of the sentence to `2 by default`

azure/modules/hana_node/salt_provisioner.tf
arbulu89: In fact, to add this we would need to add it in all the providers most probably, and the unique reason would be to change the `hana.sls` pillar file. I think this is a good idea, and we should do it, but let's do it in a future PR.

Actually, we should provide the option to set the SID and instance number from terraform, and tune the pillar files according this (with default values to make it easier)

Simranpal: Yes this is not needed currently. I will revert this, and lets update in future PR.

azure/main.tf
```diff
@@ -68,3 +68,48 @@ module "netweaver_node" {
...
+  network_subnet_id             = azurerm_subnet.mysubnet.id
+  sec_group_id                  = azurerm_network_security_group.mysecgroup.id
+  storage_account               = azurerm_storage_account.mytfstorageacc.primary_blob_endpoint
+  storage_account_name          = var.storage_account_name
+  storage_account_key           = var.storage_account_key
+  enable_accelerated_networking = var.hana_enable_accelerated_networking
+  sles4sap_uri                  = var.sles4sap_uri
+  init_type                     = var.init_type
+  hana_inst_master              = var.hana_inst_master
+  hana_inst_folder              = var.hana_inst_folder
```
arbulu89: We need to by pass the parameter `hana_instance_number` from the the main file to the hana_node module, otherwise it's using always the `00` default value.
We had the same scenario with the `hana_count` value some commits ago.

PR #331, opened 2020-01-25
Create Azure Monitoring terraform module

Simranpal:
Created a new terraform module for monitoring. 
Need to check if any existing files in azure root folder can be removed


Simranpal:
Hi @arbulu89  Do you think that we should remove certain root directory tf files, like `monitoring.tf ``instances.tf` now that we are moving the resources, definitions into modules

arbulu89:
> Hi @arbulu89 Do you think that we should remove certain root directory tf files, like `` monitoring.tf ``instances.tf `` now that we are moving the resources, definitions into modules

I would like that, but I see the value of having different variables files (but they should be renamed to, monitoring_variables.tf or so). Here we would need a bigger consensus, all the variables in one files vs different variables files.

Definetely instances.tf need to be removed.

azure/modules/monitoring/main.tf
```diff
@@ -0,0 +1,126 @@
+# Availability set for the monitoring VMs
+
+resource "azurerm_availability_set" "hana-availability-set" {
```
arbulu89: We don't need an availability set for monitoring, isn't it?
We can remove it

azure/modules/monitoring/main.tf
```diff
@@ -0,0 +1,126 @@
...
+    name              = "disk-monitoring-Data01"
+    caching           = "ReadWrite"
+    create_option     = "Empty"
+    disk_size_gb      = "10"
+    lun               = "0"
+    managed_disk_type = "Standard_LRS"
+  }
+
+  os_profile {
+    computer_name  = "monitoring"
```
arbulu89: Let's change the `computer_name` to `vmmonitoring`

Simranpal:
Thanks for the review @arbulu89 ! If this looks good, I will merge

PR #332, opened 2020-01-28
[master] Fix a typo and some grammar improvements

pablomh:


diegoakechi:
@pablomh Can you push this change also to the `develop` branch, so we avoid that it gets overwritten?

Thanks for your contribution!

PR #334, opened 2020-01-29
Feature/gcp netweaver

arbulu89:
DRBD and NW for GCP.
**INFO: The cluster creation works but the failovers won't be done properly as the GCP resource agents are not configured. They are not shipped yet in SLE or SLES4SAP.**

Summarazing:
1. Create terraform modules for DRBD and NETWEAVER
2. Adapt salt code (mostly how to download NETWEAVER software from storage buckets)
3. Create the static routes as a temporary solution
4. Update docs and pillar examples


gcp/main.tf
```diff
@@ -0,0 +1,58 @@
...
+  gcp_credentials_file      = var.gcp_credentials_file
+  network_domain            = "tf.local"
+  host_ips                  = var.netweaver_ips
+  iscsi_srv_ip              = google_compute_instance.iscsisrv.network_interface.0.network_ip
+  public_key_location       = var.public_key_location
+  private_key_location      = var.private_key_location
+  cluster_ssh_pub           = var.cluster_ssh_pub
+  cluster_ssh_key           = var.cluster_ssh_key
+  netweaver_software_bucket = var.netweaver_software_bucket
+  netweaver_nfs_share       = "${var.drbd_cluster_vip}:/HA1" # drbd cluster ip address is hardcoded by now
```
Simranpal: I see it is `drbd_cluster_vip` is now configurable via terraform.tfvars. Maybe we can remove the comment

salt/netweaver_node/files/pillar/netweaver.sls
```diff
@@ -1,10 +1,15 @@
-{% set iprange = ".".join(grains['host_ips'][0].split('.')[0:-1]) %}
+{%- set iprange = '.'.join(grains['host_ips'][0].split('.')[0:-1]) %}
```
Simranpal: Hi @arbulu89 I noticed you added dash `-` bfore the jinja statements. what is the difference with that

gcp/modules/drbd_node/main.tf
```diff
@@ -0,0 +1,67 @@
...
+
+resource "google_compute_disk" "data" {
+  count = var.drbd_count
+  name  = "${terraform.workspace}-disk-drbd-${count.index}"
+  type  = var.drbd_data_disk_type
+  size  = var.drbd_data_disk_size
+  zone  = element(var.compute_zones, count.index)
+}
+
+# temporary HA solution to create the static routes, eventually this routes must be created by the RA gcp-vpc-move-route
```
Simranpal: Hi Xabi, I did not understand what is the temporary solution for creating routes. Is this managed via gcp somehow?

salt/netweaver_node/files/pillar/netweaver.sls
arbulu89: When you add a `-` the template doesn't create a new line entry in this point of the file. Without this dash we would have an ugly empty line there hehe

gcp/modules/drbd_node/main.tf
arbulu89: The routes should be created by the `gcp-vpc-move-route` RA that we are not shipping properly yet. When the resource agent is there an properly configured we will be able to remove these static routes entries.

salt/netweaver_node/files/pillar/netweaver.sls
Simranpal: ok thanks, thats good!

PR #338, opened 2020-01-30
Make Azure VM sizes configurable for all terraform modules

Simranpal:
As most of the VM sizes `(vm_size)` in Azure were hardcoded, this PR makes them configurable, for terraform modules (drbd, hana, scsi server, netweaver, monitoring)

Also updating the different variable names (instance_type, instancetype, vm_size) to use one name ie `vm_size`

### DIFF
--- a/azure/modules/hana_node/variables.tf
+++ b/azure/modules/hana_node/variables.tf
@@ -82,6 +82,7 @@ variable "hana_public_version" {
 
 variable "vm_size" {
   type    = string
+  default = "Standard_M128s"
 }
 
 variable "admin_user" {
--- a/azure/modules/hana_node/variables.tf
+++ b/azure/modules/hana_node/variables.tf
@@ -80,7 +80,7 @@ variable "hana_public_version" {
   type    = string
 }
 
-variable "instancetype" {
+variable "vm_size" {
   type    = string
 }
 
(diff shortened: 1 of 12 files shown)


azure/modules/hana_node/variables.tf
```diff
@@ -80,7 +80,7 @@ variable "hana_public_version" {
   type    = string
 }
 
-variable "instancetype" {
+variable "vm_size" {
   type    = string
```
arbulu89: Can we add here a default value? The same as the main variables.tf file
`default = "Standard_M128s"`

azure/variables.tf
```diff
@@ -1,9 +1,9 @@
 # Launch SLES-HAE of SLES4SAP cluster nodes
 
 # Variables for type of instances to use and number of cluster nodes
-# Use with: terraform apply -var instancetype=Small -var hana_count=2
+# Use with: terraform apply -var hana_vm_size=Standard_M128s -var hana_count=2
 
-variable "instancetype" {
+variable "hana_vm_size" {
   type    = string
```
arbulu89: Add a description as in the other variables please

azure/variables.tf
```diff
@@ -91,6 +91,12 @@ variable "iscsi_srv_ip" {
   default     = "10.74.1.10"
 }
 
+variable "iscsi_vm_size" {
+  description = "VM size for the iscsi server machine. Default to Standard_D2s_v3"
```
arbulu89: I don't think we need to put the last part of the description `Default to Standard_D2s_v3`;
This is already visible in the default parameter.
I would remove it.

PR #340, opened 2020-02-03
Feature/aws hana improvements

arbulu89:
Improvements to deploy a HANA cluster in aws following the best practices guide:
https://www.suse.com/media/guide/suse_linux_enterprise_server_for_sap_applications_12_sp3_for_the_aws_cloud_setup_guide.pdf?_ga=2.67759635.36618273.1580225010-1421744106.1546416539

**INFO: This PR depends on some crmsh changes**

Changes:
- 2 availability zone
- 2 subnets
- AWS overlay IP address as floating IP address
- (security groups: We have all the ports open internally...)
- Tag the VM instances
- Create a new AWS cli profile. **Now, instead of passing the path to a cuurent aws credentials file, the keys are passed as parameter. The keys are printed in the logs, so it needs an improvement here**
- Disable the source/destinaion check
- Check cloud-netconfig-ec2 is isntalled
- Configure AWS resource agents
- Configure AWS fence agents


MalloZup:
travis is failing @Simranpal  might take look

MalloZup:
@ldevulder @juadk  how about you on this? tia

aws/instances.tf
```diff
@@ -76,7 +94,12 @@ resource "aws_instance" "monitoring" {
     device_name = "/dev/xvdd"
   }
 
+  volume_tags = {
+    Name = "${terraform.workspace}-monitoring"
+  }
+
   tags = {
     Name = "${terraform.workspace} - Monitoring"
```
Simranpal: Hi @arbulu89 the spacing before `=` needs to align with next line

Simranpal:
> travis is failing @Simranpal might take look

@MalloZup @arbulu89 the travis error with ` terraform fmt` check fails, just due to terraform space formatting I think.

ldevulder:
I'm not able to test this PR as it is not synced with the master branch. I tried to manually adapt but there is a weird change in `pillar_examples/automatic/hana/cluster.sls` with `hana_cluster_vip` variable (I don't have it in the current master branch).

Please rebase and I will try to test it ;-)

aws/variables.tf
```diff
@@ -75,9 +81,12 @@ variable "private_key_location" {
   type = string
 }
 
-variable "aws_credentials" {
```
ldevulder: Please don't remove this! It is used by the openQA tests!

I'm OK to add possibility to use variable for `key_id` and `access_key`, but we should keep the ability to use a credential file.

arbulu89: Ok, I will keep both options!

arbulu89:
@ldevulder We cannot merge this PR until some changes in crmsh reach SLE. The requests are already done. I will add some comment here when the PR is ready to be tested properly.
Thank you for your help anyway!

arbulu89:
@ldevulder Option to use an already existing aws credentials file added again (it works exactly like originally works before this changes)

arbulu89:
@diegoakechi @ldevulder @MalloZup Open question here.
This PR depends on some changes in crmsh. This changes are not in SLE yet, so we cannot depend on SCC for that. The changes are in our OBS repos though.

The unique way to make this work is to use the `devel_mode` option.
So, do you think we should merge? Or should we wait until this change reaches SLE? (this 2nd might take long...).
Anyway, the old AWS deployment is far away of being really good... (at least, compared with best practices guide...)

arbulu89:
AWS data provider implementation added. I have added it as a module, because we will need to reuse this in Netweaver. Some docs:
https://d0.awsstatic.com/enterprise-marketing/SAP/aws-data-provider-for-sap.pdfhttps://www.suse.com/media/guide/suse_linux_enterprise_server_for_sap_applications_12_sp3_for_the_aws_cloud_setup_guide.pdf?_ga=2.67759635.36618273.1580225010-1421744106.1546416539 (chapter 4.1.8)

aws/variables.tf
arbulu89: Both options are available now!

PR #348, opened 2020-02-05
Add documentation info on salt pillar encryption

Simranpal:
Adding info and links to the salt pillar encryption

Does the current Readme location makes sense? 
Should we provide the pros/cons on using gpg keys in current approach as I did?


MalloZup:
@arbulu89  imho we can live with this doc only here.

diegoakechi:
@arbulu89 I think we can merge it here, but I agree it should be mentioned on the formulas too, since formulas are intended to be used stand-alone. @Simranpal can you take care of creating the PR for the other projects?

PR #343, opened 2020-02-04
Move remaining resources to main.tf, variables.tf

Simranpal:
I have moved the remaining resources (outside existing modules) into the main.tf.
mainly provider, resource group & network.

Also updated readme to reflect the changes.

PR #352, opened 2020-02-10
Pass the missing tf variables from main.tf to modules for Azure

Simranpal:
Some variables' references were not added in the main.tf file, after creating the terraform modules. These variables however exist in the main variable file and also in the module variable files. But in the last PRs the main.tf was not updated with these values, which leads to broken monitoring.
Here I am adding those missing variables.


Simranpal:
I will test the full deployment with monitoring and remove the WIP label

Simranpal:
Hi @arbulu89 for all the modules variables, I will check them and update in https://github.com/SUSE/ha-sap-terraform-deployments/pull/343
as in this PR, not all modules were created

Simranpal:
Hi @arbulu89 there were some missing variable that were not set. I have updated them now in the main.tf

PR #356, opened 2020-02-18
Move DRBD pillar example together with hana example.

nick-wang:
Polish the pillar example and add index to SAP HANA/DRBD cluster, to avoid forget copy DRBD pillars in a full(HANA/NW/DRBD) deployment.


MalloZup:
thx!

PR #365, opened 2020-02-24
Fix iscsi queue_depth configuration error and hard code of instance type of iscsi/monitoring nodes on aws

nick-wang:
Fix issue#357 and issue#258

Remove the hard code of instance type of iscsi/monitoring nodes, since it is different per region.
Correct the queue_depth configuration of iscsi pillar file.

### DIFF
--- a/aws/README.md
+++ b/aws/README.md
@@ -130,7 +130,11 @@ These are the relevant files and what each provides:
 
 In [terraform.tfvars](terraform.tfvars.example) there are a number of variables that control what is deployed. Some of these variables are:
 
-* **instancetype**: instance type to use for the cluster nodes; basically the "size" (number of vCPUS and memory) of the instance. Defaults to `t2.micro`.
+* **instancetype**: instance type to use for the hana cluster nodes; basically the "size" (number of vCPUS and memory) of the instance. Defaults to `r3.8xlarge`.
+* **netweaver_instancetype**: instance type of netweaver node; basically the "size" (number of vCPUS and memory) of the instance. Defaults to `r3.8xlarge`.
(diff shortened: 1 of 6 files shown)


nick-wang:
> Hi @nick-wang ,
> I think that for some recent changes, the monitoring and iscsi server code is now in modules, so this change would have conflicts (and it has).
> Besides that, I would split the image of monitoring and iscsi in 2 different variables (maybe with the modules they are not even named `instancetype`...).
@arbulu89 No problem, I can modify and test again based on the new change.
For sure, i can split the variables for monitoring/iscsi images. But i assume most of time, they are using the minimum standard image. So my idea
is we have a default less cost(free tier) instancetype, like `t2.micro`. Then have 2 variables for monitoring and iscsi image but by default is empty. It will replace the `free tier` image unless explicit defined. In this case, customer doesn't need to remember modify all the `free tier` images when switch to a new region.
How do you think of it? 

(My AWS account is expired... i will ask Diego for the access of team account if it is OK)

> 
> About the `depth` thing, I don't really know what's the difference, so I trust in your change hehe

I meet a error message in iscsi server about the unknown parser of `queue_depth` of iscsi pillar. I forget to copy the content. If you have the iscsi node, maybe you can find the error in system log.
For sure, the `queue_depth` didn't effect with the original configuration. For the new method of configuring in iscsi configuration, i did check the result in my deployment.

arbulu89:
@nick-wang Yes, I like that. Have the a small instance type by default, being able to be changed. And the same for the image, yes, do this.

Thanks!

nick-wang:
@arbulu89 merged with the latest change and split the instancetype of iscsi/monitoring.

nick-wang:
Thanks for the review.

PR #368, opened 2020-02-24
Use latest available AMI for AWS/GCP/Azure

ldevulder:
in GCP and Azure we always use the latest OS image available.
This commit adds this for AWS.

Tested successfully on AWS with and without a custom image.

**Note:** as SLE15 is in extended support (not sure for SLES4SAP), it's now better to user 15-SP1 by default.


aws/variables.tf
```diff
@@ -1,45 +1,37 @@
...
-    "eu-west-1"    = "ami-0ca96dfbaf35b0c31"
-    "eu-west-2"    = "ami-00189dbab3fd43af2"
-    "eu-west-3"    = "ami-00e70e3421f053648"
+data "aws_ami" "latest_sles4sap" {
+  most_recent = true
+  owners      = ["013907871322"] # SUSE Linux
+
+  filter {
+    name   = "name"
+    values = ["suse-sles-sap-15-sp1-byos-*"]
```
arbulu89: Is there any way to get the latest version using other way? It would be better, instead of having this hardcoded, some kind of regular expressions maybe

ldevulder: It's pretty much what we are doing in GCP or Azure.

On GCP:

    variable "sles4sap_boot_image" {
      type    = string
      default = "suse-byos-cloud/sles-15-sap-byos"
    }

On Azure:

    hana_public_publisher = "SUSE"
    hana_public_offer     = "SLES-SAP-BYOS"
    hana_public_sku       = "15"
    hana_public_version   = "latest"

So I don't think it is a good idea to do something different (or could we do the same for GCP/Azure?).

ldevulder:
> @ldevulder Could we add the `aws_region` part to the filtering to get the latest image? It makes sense to use the image from the region you are launching.

It is automatically done! I tried this at the beginning but I had an error. I checked and the call to this `data` function couldn't be made if `aws_region` is not set globally. So we will only get AMI available in the selected region. 

> Besides that, I would take advantage and create a new variable like `monitoring_image` to have it different of the one use for the cluster, it makes sense.

Yes, more for another PR I think, as I will need this PR to create a new tagged version `v3.1.1`, I would like to only change the minimal things in this PR.

aws/variables.tf
alvarocarvajald: So, latest 15-SP1 in AWS, and latest 15-GM in Azure and GCP? Couldn't this PR include the changes to have all 3 providers in the same latest version as well?

aws/instances.tf
```diff
@@ -57,7 +57,7 @@ resource "aws_instance" "clusternodes" {
 
 resource "aws_instance" "monitoring" {
   count                       = var.monitoring_enabled == true ? 1 : 0
-  ami                         = var.sles4sap[var.aws_region]
+  ami                         = substr(var.sles4sap, 0, 4) == "ami-" ? var.sles4sap : data.aws_ami.latest_sles4sap.image_id
```
arbulu89: I would create an independent variable for this. I guess the monitoring server doesn't need to have the same image than the hana nodes

aws/variables.tf
arbulu89: Agree with Alvaro. We could put them all with SLE15 or SLE15SP1, but the same for all of them

aws/variables.tf
```diff
@@ -1,45 +1,37 @@
...
-    "eu-west-1"    = "ami-0ca96dfbaf35b0c31"
-    "eu-west-2"    = "ami-00189dbab3fd43af2"
-    "eu-west-3"    = "ami-00e70e3421f053648"
+  filter {
+    name   = "virtualization-type"
+    values = ["hvm"]
   }
 }
 
+variable "iscsi_srv" {
```
arbulu89: Could you add the description here? The name itself is not so meaningful)

arbulu89:
@ldevulder Any update on this?
I think if we just add the description to the new variables everything else is fine.
Other thing, did you start working in this issue because you found some issue in the older versions?
I'm facing a connectivity issue with the default image... the obs repositories connection is failling

aws/variables.tf
ldevulder: Agree, done for Azure and GCP.

ldevulder:
> This way we could use the names in a more friendly way

OK, I can try to do this in this PR.

aws/instances.tf
ldevulder: OK.

aws/variables.tf
ldevulder: OK.

ldevulder:
> This way we could use the names in a more friendly way

@arbulu89 In fact after some searches/tests I don't know how to handle this without duplicating the `data "aws_ami"` part, as you requested to have a separate variable for the monitoring server (and I also have one for the iSCSI server).

If I use something like `values = ["{{ var.sles4sap}}-*"]` in that case all server types will have the same OS by default.

So, I see these possibilities (but I will rename `var.sles4sap` to `var.sles4sap_image`, better name IMHO), pretty much what you suggested but I want to be sure:
- like you said, if `var.sles4sap_image` variable starts with `ami-` use it
- if `var.sles4sap_image` is set but not with `ami-`, use it in the filter part but for **ALL** nodes (HANA, iSCSI,  Monitoring, etc.)
- if `var.sles4sap_image` is not set use the latest version defined (15-SP1 currently)

Each server types could be overwritten directly with its variable, for example: `iscsi_os_image = "ami-XXXX"`, but not with a more friendly name. Or, as I said, I will have to duplicate the `data "aws_ami"` part for each server type to be able to do this.

aws/instances.tf
ldevulder: Done.

aws/variables.tf
ldevulder: Done.

aws/variables.tf
```diff
@@ -129,20 +156,10 @@ variable "ninstances" {
...
-    "us-west-1"    = "ami-0f213357578720889"
-    "us-west-2"    = "ami-0fc86417df3e0f6d4"
-    "ca-central-1" = "ami-0811b93a30ab570f7"
-    "eu-central-1" = "ami-024f50fdc1f2f5603"
-    "eu-west-1"    = "ami-0ca96dfbaf35b0c31"
-    "eu-west-2"    = "ami-00189dbab3fd43af2"
-    "eu-west-3"    = "ami-00e70e3421f053648"
-  }
+variable "hana_os_image" {
+  description = "OS image for hana node."
```
arbulu89: I would improve the description as the usage is quite complicated with the new change:
`sles4sap AMI image identifier or a pattern used to find the image name (e.g. suse-sles-sap-15-sp1-byos)`

aws/main.tf
```diff
@@ -11,7 +11,7 @@ module "netweaver_node" {
   aws_account_id            = var.aws_account_id
   aws_region                = var.aws_region
   availability_zones        = data.aws_availability_zones.available.names
-  sles4sap_images           = var.sles4sap
+  netweaver_os_image        = substr(var.netweaver_os_image, 0, 4) == "ami-" ? var.netweaver_os_image : data.aws_ami.latest_os_image.image_id
```
arbulu89: I think that this logic should be implemented inside the module. 
Anyway, we can change it in the future when we add terraform modules for all the components

aws/variables.tf
```diff
@@ -33,6 +33,33 @@ variable "private_key_location" {
...
+variable "default_os_image" {
+  description = "Default OS image"
+  type        = string
+  default     = "suse-sles-sap-15-sp1-byos"
+}
+
+# Data used to get the latest version of an AMI image
+data "aws_ami" "latest_os_image" {
+  most_recent = true
+  owners      = ["013907871322"] # SUSE Linux
```
arbulu89: I think this value should be configurable.

aws/variables.tf
```diff
@@ -33,6 +33,33 @@ variable "private_key_location" {
   type = string
 }
 
+variable "default_os_image" {
```
arbulu89: Most probably, in the future, when we have everything in modules, we will be able to remove this parameter. By now is OK

ldevulder: Why removing this variable? Allowing user to set the default value for all OS images could be fine, no?
Anyway, if you really want to I can remove it.

aws/variables.tf
ldevulder: Make sense, I will do it.

aws/variables.tf
ldevulder: Ok.

aws/main.tf
ldevulder: Ok.

ldevulder:
New version manually tested: we can use official AMI for all or only for some part. And the same for customized AMI. Could be useful for Build Validation: we can use internal AMI for some part (HANA, DRBD or NW_ and keep official AMI for all others.

aws/terraform.tfvars.example
```diff
@@ -36,9 +36,7 @@ public_key_location = "/path/to/your/public/ssh/key"
 private_key_location = "/path/to/your/private/ssh/key"
 
 # Custom AMI for nodes
-#sles4sap = {
-#    "eu-central-1" = "ami-xxxxxxxxxxxxxxxxx"
-#}
+#hana_os_image = "ami-xxxxxxxxxxxxxxxxx"
```
arbulu89, line 39: @ldevulder Can we put as an example just after the next one something like:
```
#hana_os_image = "ami-xxxxxxxxxxxxxxxxx"
# Or
#hana_os_image = "suse-sles-sap-15-sp1-byos"
```
To make more obvious that the 2 options are available. People don't usually check the variables description, and they only check the template, so having here the options makes things better

ldevulder: Done!

aws/variables.tf
ldevulder: In fact, with the refactoring of the code this variable has been removed :-)

PR #370, opened 2020-02-26
Change default instance for azure

juadk:
This PR changes default instance for Azure and adds links about certified systems in documentation.

Azure warned us about the default instance we are using in the project.
They told to use smaller instance which is also supported.
We chose M128S because it is a recommended instance for clustering but Azure told us that clustering means for scale-out scenario and not HA clustering.


diegoakechi:
@juadk I think, as this is not a really critical change, I would submit this to the `develop` branch so we avoid the need to backporting patches between branches.

PR #371, opened 2020-02-26
Change default instance for azure

juadk:
This PR changes default instance for Azure and adds links about certified systems in documentation.

Azure warned us about the default instance we are using in the project.
They told to use smaller instance which is also supported.
We chose M128S because it is a recommended instance for clustering but Azure told us that clustering means for scale-out scenario and not HA clustering.


juadk:
> Hi @juadk ,
> We would need to change this default value as well:
> https://github.com/SUSE/ha-sap-terraform-deployments/blob/develop/azure/modules/hana_node/variables.tf#L85
> 
> Besides that, everything good

Thanks for pointed me this.
Done

PR #372, opened 2020-02-26
Fix salt scripts usage to run the proper top files

arbulu89:
This PR fixes 2 issues:

- `mount_sapmnt` is used in the `sapnwbootstrap-formula`, so we need to change the name to avoid duplicated states
- In order to keep idempotency properly, we need to split the execution of the formulas and the pre deployment. This is done having different salt envs that are executed in the highstate.

Before that, the `iscsi_srv` execution was failing because the `file_root` variable was not getting the folder where iscsi formula states are stored.

I have tested with AWS and works fine now.

### DIFF
--- a/salt/deployment.sh
+++ b/salt/deployment.sh
@@ -1,5 +1,7 @@
 #!/bin/bash -xe
 
+# Execute the states within /root/salt/pre_installation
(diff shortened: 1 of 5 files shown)


MalloZup:
we can merge this and see if it fix

PR #377, opened 2020-03-02
Add timeout for avoiding infinite looping

MalloZup:
this Pr add timeouts 3h when creating cluster.
I had currenlty problem that the cluster creation never stop.

also in future scenario of automated CI is good to have a limit to don't waste resources. I guess 3H should be enough in normal/exceptional cases imho


MalloZup:
ok I wil test and submit later on closing form moment

PR #381, opened 2020-03-06
Improve AWS documentation about IAM policies

arbulu89:
Here updated the documentation about how to use AWS IAM policies.

Sending the changes to `master` as it is a pretty important detail


aws/README.md
```diff
@@ -60,11 +60,31 @@ This file is also used by the `aws` command line tool, so it can be created with
 
 In order to execute the deployment properly using terraform, the used user must have some policies enabled. Mostly, it needs access to manage EC2 instances, S3 buckets, IAM (to create roles and policies) and EFS storage.
 
-Here how it should look like:
+In order to setup the IAM proper rights, 2 options are available:
+- Set the `IAMFullAccess` policy to the user running the project (or to the group which the user belongs to). This is not recommended as this IAM policy give full IAM access to the user.
```
diegoakechi: Maybe we could remove this option from the docs, since documenting the minimal permission needed already implies that more powerful users will also be able to execute the project.

arbulu89:
@juadk I will test again my configuration to see what happens there...

aws/README.md
arbulu89: I would let it as it is. I'm seeing that assumptions here are not helpful. It will come later other request saying to improve the docs...

arbulu89:
@juadk Docs updated with the good policies

PR #382, opened 2020-03-09
Grafana/dashboard: Add initial process table for netweaver

MalloZup:
# What does this PR?

- this pr add an initial dashboard + Process table of process.
- It add also the needed grafana variables like `node_name` etc which we use in other dashboards etc

### DIFF
--- a/salt/monitoring/provisioning/dashboards/netweaver-cluster-status.json
+++ b/salt/monitoring/provisioning/dashboards/netweaver-cluster-status.json
@@ -17,7 +17,92 @@
   "graphTooltip": 0,
   "id": 4,
   "links": [],
-  "panels": [],
+  "panels": [
+    {
+      "columns": [],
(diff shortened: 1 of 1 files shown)


stefanotorresi:
@MalloZup do you have a live deployment of this dashboard?

MalloZup:
@stefanotorresi it was the dashboard I showcased on the live-demo

I think right now we can merge it as minimal step. I will try to deploy a new netweaver cluster right now

stefanotorresi:
I'm not sure about the node switcher, though, which was not in the demo yesterday. :wink:
We have to decide if we want the dashboard to present info of all the nodes, akin to what we do with the multi-cluster dashboard: in that case, we don't need any variable switch.

MalloZup:
ok removed please check thx! now is only a minimal graph which we can change later. 
It is only a skeleton right no

PR #385, opened 2020-03-10
Feature/ha repo automatic

arbulu89:
As we have more clusters and machines now, that might have different SLE version, a hardcoded repository URL doesn't make sense anymore. In order to fix that, I have implemented an autodetection.
So:
- If SLE version is set in the repo url, it will just be used
- If not, the SLE version is automatically detected

PD: We can implement the same for the SCC extensions in the future


arbulu89:
@juadk Thank you for your comment. First things first.
I don't know if we are going to ship `iscsi-formula` in SLE, I don't have plans for that at the moment at least.
Second, I understand your point, and what I would do is just if the coming `ha_sap_deployment_repo` parameters is an empty string, we don't add any new repository. As simple as this. I think it makes sense.
As you say, if all of the involved packages are coming in SLE, we don't really need this repo, so setting it empty (it's the default value in fact) would make the job.

juadk:
> Second, I understand your point, and what I would do is just if the coming `ha_sap_deployment_repo` parameters is an empty string, we don't add any new repository. As simple as this. I think it makes sense.

Totally agree, it's even simpler :smile:

ldevulder:
What could be do I think is:
- use the default provided packages (available in the cloud image) if `ha_sap_deployment_repo` is empty
- use the repos set in the configuration file (whet we currently do)
- use your autodetection code if `ha_sap_deployment_repo = auto` is set

WDYT?

arbulu89:
@ldevulder The autodetection is only to detect the specific SLE version. I doesn't detect any particular repository, this needs to be provided (full path except SLE version).

PR #387, opened 2020-03-11
Enable sap host exporter

MalloZup:
this PR build on https://github.com/SUSE/sapnwbootstrap-formula/pull/33

adding @stefanotorresi  as non-official reviewer so he can see what we need to enable the exporter in salt

### DIFF
--- a/salt/netweaver_node/files/pillar/netweaver.sls
+++ b/salt/netweaver_node/files/pillar/netweaver.sls
@@ -35,7 +35,7 @@ netweaver:
 
   # enable exporter if monitoring enabled
   sap_host_exporter:
-    enabled: {{ grains.get('monitoring_enabled', False) }}
+    enabled: {{ grains.get['monitoring_enabled', False] }}
 
   hana:
 # We have to unify the usage of this parameter, the aws option looks better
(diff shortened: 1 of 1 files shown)


stefanotorresi:
@MalloZup I would take the chance to also add the authentication via the config file.

salt/netweaver_node/files/pillar/netweaver.sls
```diff
@@ -33,6 +33,12 @@ netweaver:
   # apply by default the netweaver solution
   saptune_solution: 'NETWEAVER'
 
+  # enable sap_host_exporter
+  {% if grains.get('monitoring_enabled', False) %}
+   sap_host_exporter:
```
arbulu89: It looks like the indentation is wrong, right? Some extra spaces.
I don't know if this affects, but yaml are quite sensitive with these things.

arbulu89: It could be simplified just with (current solution is good though):
```
sap_host_exporter:
          enabled: {{ grains.get('monitoring_enabled', False) }}
```

MalloZup: done thx

PR #389, opened 2020-03-12
Feature/local exec

arbulu89:
Hello,
Implementation of some local steps that are needed (most of the times) to run the deployment. It includes:
- Copy the content from pillar_examples/automatic to the corresponding folders
- Create cluster sshkeys

The terraform code is implemented in a module located in `generic_modules` that allows it usage to all the providers without duplicating any code.

This execution is optional using the `pre_deployment` parameter (by default false, so the steps are not executed)


generic_modules/local_exec/main.tf
```diff
@@ -0,0 +1,18 @@
...
+    command = <<EOT
+      cp pillar_examples/automatic/hana/* salt/hana_node/files/pillar;
+      cp pillar_examples/automatic/drbd/* salt/drbd_node/files/pillar;
+    EOT
+  }
+  provisioner "local-exec" {
+    working_dir = "${path.module}/../.."
+    command = <<EOT
+      mkdir -p salt/hana_node/files/sshkeys/;
+      rm -rf salt/hana_node/files/sshkeys/*;
```
diegoakechi: @arbulu89 Maybe we should check if the file exist and if not, we create that. Not sure what happens if we run the deployment and change something and re-rerun, if it would make the keys be uploaded again...

arbulu89: Ok, I will test this scenario.

arbulu89:
> looks good
> my thoughts on this.
> 
> why we are introducing the `pre-deployement` variable?
> 
> If my reasoning is correct, it is because we have `performance` and other pillars right, which would be overwritten right?
> 
> Maybe we could use the automatic directory as placeholder, so modifying the doc to say that the user should put the pillar they want into this directory.
> 
> we could call this directory not anymore `automatic` but something like `default` or soemthing similar.
> Imho we could spare in this way the variable perhaps, and copy all pillars by defualt.
> 
> If the user choose other, he will just need to copy it in this dir.
I added this new parameter because I want to give the option to follow the current process, where you do the steps, and you can choose your own pillar files, sshkeys, etc
The whole idea is to remove some manual steps, what you say would create more manual steps (not more, just change where to copy the pillars files, for that we don't do any change I guess...).

The `pillar_examples/automatic` stores pillar files that work with all the providers, and are filled with data coming from terraform, they are more than simple examples. That's why I copy them.

arbulu89:
@diegoakechi Change done to only create the keys if there are not already there.
And after thinking about this, we could just executed this operation always (without the condition) and remove the variables `cluster_ssh_pub` and `cluster_ssh_key` requested to the user, just to make everything a bit more simpler. Or at least put a default value. I guess there is some change where the users want to use some already created keys.

Setting the default value looks like the best option.

aws/README.md
```diff
@@ -158,6 +158,7 @@ In [terraform.tfvars](terraform.tfvars.example) there are a number of variables
 * **scenario_type**: SAP HANA scenario type. Available options: `performance-optimized` and `cost-optimized`.
 * **provisioner**: select the desired provisioner to configure the nodes. Salt is used by default: [salt](../salt). Let it empty to disable the provisioning part.
 * **background**: run the provisioning process in background finishing terraform execution.
+* **pre_deployment**: Enable pre deployment local execution steps (disabled by default).
```
MalloZup: I think will could be more specific in the description. 

```
* **pre_deployment**: Enable pre deployment local execution steps, e.g copy salt pillars (disabled by default).

```

From my pov we will not add more operation but imho is good to know what this `pre-deployment` does in the description without having to look at code.

PR #390, opened 2020-03-13
Add default value for NW storage location

juadk:
This PR adds a default value for Netweaver storage location.
Otherwise, terraform is asking for value even if Netweaver stack is disabled.

Tested successfully on AWS.


arbulu89:
@juadk As this is a `bug` could you redirect the PR to master?

PR #392, opened 2020-03-13
Add default value for NW storage location

juadk:
This PR adds a default value for Netweaver storage location.
Otherwise, terraform is asking for value even if Netweaver stack is disabled.

Tested successfully on AWS.


arbulu89:
@juadk Other user has raised the issue in GCP that `netweaver_virtual_ips` and `drbd_cluster_vip` are requested as well. They don't have any default value. As this is exactly the same reason of this PR, could set default values to them in this change. Empty value for `drbd_cluster_vip` and empty list for `netweaver_virtual_ips` would be enough

juadk:
> @juadk Other user has raised the issue in GCP that `netweaver_virtual_ips` and `drbd_cluster_vip` are requested as well. They don't have any default value. As this is exactly the same reason of this PR, could set default values to them in this change. Empty value for `drbd_cluster_vip` and empty list for `netweaver_virtual_ips` would be enough

@arbulu89 , Done and my test worked well in GCP, commit message updated with all changes.

PR #396, opened 2020-03-18
Fix iscsi initiator cannot automatic login after disconnected.

nick-wang:
Somehow, enable iscsid service may cause iscsi don't
automatical login after disconnecting for any reason,
which need to restart iscsid to fix.

Error log:
iscsid: Kernel reported iSCSI connection 1:0 error
 (1020 - ISCSI_ERR_TCP_CONN_CLOSE: TCP connection closed)

This issue can not reproduce on libvirt, only in AWS/Azure.

### DIFF
--- a/salt/cluster_node/iscsi_initiator.sls
+++ b/salt/cluster_node/iscsi_initiator.sls
@@ -38,6 +38,5 @@ iscsi_discovery:
 
 iscsid:
   service.running:
-    - enable: True
     - watch:
       - cmd: iscsi_discovery


nick-wang:
Unnecessary to enable iscsid, since iscsi will always launch iscsid. 
But somehow it should be harmless, especially no reboot happened.

salt/cluster_node/iscsi_initiator.sls
```diff
@@ -26,6 +26,7 @@ iscsi:
     - watch:
       - file: /etc/iscsi/iscsid.conf
       - file: /etc/iscsi/initiatorname.iscsi
+      - iscsi-queue-depth
```
arbulu89: I'm not totally sure, but is this line doing something?
I mean, the `iscsi-queue-depth` changes `/etc/iscsi/iscsid.conf`, so it looks redundant.
Anyway, is not harmful, and I'm not really sure how the `file` watch works

nick-wang: @arbulu89 Yes, you are right. For god sake, i miss the same file above. I am blind.

nick-wang:
Thanks for review. Rebased on the latest commit.

PR #400, opened 2020-03-25
Update hana_inst_folder default value

arbulu89:
We have found out that the usage of the `/root` folder to store HANA installation software is not really good idea.

The installer complains of having invalid permissions if the used HANA installation is obtained from a SAR file if the files are in `/root` folder.

In order to avoid that, the easy solution is just to have a different default value.
`/hana_inst_media` just works fine, but other options could be good as well.

My unique doubt is if this could have some security risk (I cannot find any).

### DIFF
--- a/aws/terraform.tfvars.example
+++ b/aws/terraform.tfvars.example
@@ -42,7 +42,7 @@ name = "hana"
 hana_inst_master = "s3://path/to/your/hana/installation/master"
 
 # Local folder where HANA installation master will be downloaded from S3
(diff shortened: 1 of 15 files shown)


Simranpal:
I think this location is good. There is no specific recommendations I see from SAP install guides. Another idea is store all SAP media under one directory `/sapmedia/HANA` `/sapmedia/NW` etc

arbulu89:
Pending to update `/hana_inst_media` to `/sapmedia/HANA` and `/netweaver_inst_media` to `/sapmedia/NW`

PR #406, opened 2020-03-31
update prometheus nw targets

stefanotorresi:
This PR changes Prometheus configuration to use floating ips as target of the netweaver exporters.
It also introduces the `monitoring_enabled` variable in the netweaver pillar, which was missing and is leveraged by SUSE/sapnwbootstrap-formula#43

### DIFF
--- a/aws/salt_provisioner.tf
+++ b/aws/salt_provisioner.tf
@@ -106,7 +106,7 @@ additional_packages: [${join(", ", formatlist("'%s'", var.additional_packages))}
 host_ip: ${var.monitoring_srv_ip}
 ha_sap_deployment_repo: ${var.ha_sap_deployment_repo}
 monitored_hosts: [${join(", ", formatlist("'%s'", var.host_ips))}]
-nw_monitored_hosts: [${join(", ", formatlist("'%s'", var.netweaver_enabled ? var.netweaver_ips : []))}]
+nw_monitored_hosts: [${join(", ", formatlist("'%s'", var.netweaver_enabled ? var.netweaver_virtual_ips : []))}]
 network_domain: "tf.local"
 EOF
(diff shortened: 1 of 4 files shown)


arbulu89:
@stefanotorresi Is it safe to enable the `sap_host_exporter`? I mean, if it breaks the system, we shouldn't install it by now.

stefanotorresi:
as discussed in chat: this PR doesn't enable it, it's already enabled. This PR only changes the way Prometheus scrapes the targets, using floating IPs instead of static ones.

arbulu89:
> as discussed in chat: this PR doesn't enable it, it's already enabled. This PR only changes the way Prometheus scrapes the targets, using floating IPs instead of static ones.

I know I know. I'm saying that we should disable it. Not that it was added in this PR hehe

MalloZup:
@arbulu89 good point xabi , I forgot it also :grin:

stefanotorresi:
yeah, you are both right! I will do this shortly.

stefanotorresi:
@arbulu89 I added the changes to aws and azure providers, but I couldn't figure out how to do that for gcp, as that provider has a different structure. Could you give me some pointers?

stefanotorresi:
> @stefanotorresi What about google cloud? Don't we have there the NW monitoring?
> If not we can merge this PR, but we should create a task for gcp

that's what I was asking! :stuck_out_tongue:

arbulu89:
> > @stefanotorresi What about google cloud? Don't we have there the NW monitoring?
> > If not we can merge this PR, but we should create a task for gcp
> 
> that's what I was asking! stuck_out_tongue

Sorry, this is what happen when you answer before reading. I was just updating the previous message hehe

Issue #405, opened 2020-03-29
Please ADD the un-registration of SUSE Server subscription before "destroy"


picoroma:
[FEATURE] Due to some SUSE subscription limit, is usefull to add the un-registration of the VM involved - before to perform the undeploy of the stack (terraform destroy).

diegoakechi:
@picoroma This is a good idea... We will include in our backlog for future implementations.

arbulu89:
@picoroma The feature has been implemented in #412. It will be released in the next version (`5.0.0` most probably) that will come soon, in some few weeks.

PR #408, opened 2020-04-01
Add on_destroy scripts to unregister from SCC

arbulu89:
Feature for https://github.com/SUSE/ha-sap-terraform-deployments/issues/405

This implementation adds the `destroy` hook to the VMs. So, when `terraform destroy` is executed, the `scripts/on_destroy.sh` will be executed. I have created the `scripts` folder because I didn't want to put in `salt` folder, as this hooks are not salt related. I would say the unregistration is something that should **always** occur on destroy. I cannot figure it out any use case where a destroyed machine shouldn't be unregistered.

The question is if the folder name `scripts` is appropriate or if we should choose other (`support_scripts`, `support`, etc)

If you wonder why I didn't put this `provisioner` in `salt-provisioner.tf` file, the reason is that it's not possible. The `destroy` hook must be used in the resource that is destroyed. More info:
https://github.com/hashicorp/terraform/issues/24076

**PD: I will add this to the other providers as soon as this PR is accepted**

**PD2: Bad news... I cannot implement the same approach for the cloud providers. The interpolation is not allowed for destroy time provisioneners: https://github.com/hashicorp/terraform/issues/23679. I will try to find other solution**

### DIFF
--- a/.ci/validate-terraform.sh
+++ b/.ci/validate-terraform.sh
@@ -11,7 +11,7 @@ done
 
 
 echo "executing terraform check , init and validate for each provider"
-for provider in $(find * -maxdepth 0 -type d | grep -Ev 'salt|pillar_examples|doc'); do
+for provider in $(find * -maxdepth 0 -type d | grep -Ev 'scripts|salt|pillar_examples|doc'); do
   echo "============================"
   echo "doing tests for $provider"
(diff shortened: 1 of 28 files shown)


arbulu89:
I'm closing the PR and I will open other, to have a new description

PR #410, opened 2020-04-02
Create iscsi server module for AWS provider

Simranpal:
Creating a new terraform module in AWS to separate the logic for iscsi server.


Simranpal:
Also created `iscsi_srv_ip` with default address = `"10.0.0.254"`

Simranpal:
Hi @arbulu89 with the recent error I noticed, that you added (https://github.com/SUSE/ha-sap-terraform-deployments/commit/5d91553b44ae9d5ecaf707622c633d447e70fc9a) some variables `name` `public_key_location` `private_key_location`. As they are duplicates, I will remove them here

arbulu89:
@Simranpal I think that it's better to group all of the modules in the same pull request. Less reviews to be done. All of them look the same.

About the ssh keys. Yes remove them, please? I hope they don't cause any error at least, right?

aws/main.tf
```diff
@@ -30,7 +54,7 @@ module "netweaver_node" {
   virtual_host_ips          = var.netweaver_virtual_ips
   public_key_location       = var.public_key_location
   private_key_location      = var.private_key_location
-  iscsi_srv_ip              = aws_instance.iscsisrv.private_ip
+  iscsi_srv_ip              = var.iscsi_srv_ip
```
arbulu89, line 57: It will be better something like `module.iscsi_server.aws_instance.iscsisrv.private_ip` (get from the output maybe...).
At least there is an option to provide data from the module. Better to use that, terraform creates internal dependencies in that cases that is really useful

aws/main.tf
```diff
@@ -3,6 +3,30 @@ module "local_execution" {
   enabled = var.pre_deployment
 }
 
+module "iscsi_server" {
+  source                 = "./modules/iscsi_server"
+  aws_region             = var.aws_region
+  availability_zones     = data.aws_availability_zones.available.names
+  subnet_ids             = aws_subnet.hana-subnet.*.id
```
arbulu89, line 10: I'm wondering if we would just provider a unique `subnet_id`, instead of a list. The unique reason I can think about is if in the future we give the option to create multiple iscsi servers.

aws/modules/iscsi_server/salt_provisioner.tf
```diff
@@ -0,0 +1,71 @@
+# Template file for user_data used in resource instances
+data "template_file" "salt_provisioner" {
```
arbulu89, line 2: Just as a comment. I noticed that we have this template duplicated all around the modules...
We should create a unique resource somehow. I don't even know if we need a template.
We should consider for the future to improve this.

aws/variables.tf
```diff
@@ -230,6 +217,12 @@ variable "iscsidev" {
   type        = string
 }
 
+variable "iscsi_srv_ip" {
+  description = "iscsi server address"
+  type        = string
+  default     = "10.0.0.254"
```
arbulu89, line 223: We should say that the ip address must be in the same ip range

aws/main.tf
```diff
@@ -3,6 +3,30 @@ module "local_execution" {
   enabled = var.pre_deployment
 }
 
+module "iscsi_server" {
+  source                 = "./modules/iscsi_server"
+  aws_region             = var.aws_region
+  availability_zones     = data.aws_availability_zones.available.names
```
arbulu89, line 9: I don't know if we need multiple availability zones here, like in netweaver. It should totally work anyway.

Simranpal: Yes it is now passed as list, and in the instance definition we are setting it to the first availability zone from list

aws/main.tf
Simranpal: for other providers like in Azure we are setting a unique `subnet_id`
But for the reason you mention it maybe worth to keep as list

aws/main.tf
Simranpal: Hi @arbulu89 I checked for using the output variables, but could not find if we are using it in main.tf file. I am changing the `iscsi_srv_ip` to get the value from `iscsi_server` as you mentioned. I will change the same for other modules I create, and maybe also check if we can apply this approach to other variables I create.

aws/main.tf
Simranpal: With this approach, in the end, it is set to one `availability_zone`. We can also change it in `main.tf` to pass only 1 value instead of list.

aws/main.tf
Simranpal: Actually, I corrected it now, to get it from the output from iscsi_server module

MalloZup:
what is the status about this?

aws/salt_provisioner.tf
```diff
@@ -102,7 +48,7 @@ hdbserver_sar: ${var.hdbserver_sar}
 hana_extract_dir: ${var.hana_extract_dir}
 hana_disk_device: ${var.hana_disk_device}
 hana_fstype: ${var.hana_fstype}
-iscsi_srv_ip: ${aws_instance.iscsisrv.private_ip}
+iscsi_srv_ip: ${var.iscsi_srv_ip}
```
arbulu89, line 51: I think you need to update this `iscsi_srv_ip` to `module.iscsi_server.iscsisrv_ip`

Simranpal: This value would be logically same, as `var.iscsi_srv_ip` and `aws_instance.iscsisrv.private_ip` take the same value from terraform

arbulu89: But when you use `aws_instance.iscsisrv.private_ip` instead of `var.iscsi_srv_ip` terraform creates some internal dependencies between these 2 resources. For example, the order in which the resources are created. I think updating with `aws_instance.iscsisrv.private_ip` is the correct thing

arbulu89: We have to set `module.iscsi_server.iscsisrv_ip` here

PR #411, opened 2020-04-02
Remove duplicated variables from aws variables

arbulu89:
When I brought change from Master to develop in https://github.com/SUSE/ha-sap-terraform-deployments/commit/5d91553b44ae9d5ecaf707622c633d447e70fc9a
I failed editing the conflicts..
In fact, as I did the merge manually to avoid more PRs, the CI was not executed.

Maybe in the future, we should do proper merge requests to sync develop with master

### DIFF
--- a/aws/variables.tf
+++ b/aws/variables.tf
@@ -4,19 +4,6 @@ variable "aws_region" {
   type = string
 }
 
-variable "name" {
-  description = "hostname, without the domain part"
-  type        = string
-}
-
-variable "public_key_location" {
-  type = string
-}
-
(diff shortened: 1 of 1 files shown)


arbulu89:
> Good catch for the fix.
> Just curious why/when need merge from `master` to `develop`, i though every changes should submit to `develop` first.

We don't have any strict process for that. I did just what it was the easiest thing. Submitting changes to develop first is really complicated, as the files might be already changed for some other reason, and later cherry picking exact commits and fixing conflicts is a nightmare.

PR #412, opened 2020-04-02
Feature/on destroy

arbulu89:
Feature for: https://github.com/SUSE/ha-sap-terraform-deployments/issues/405

Coming back again to the `on_destroy` functionality. After exposing the issues in https://github.com/SUSE/ha-sap-terraform-deployments/pull/408 I have decided the implement other approach.

I have created the `on_destroy` terraform module. I have implemented there the whole logic, and taking advantage of the `trigger` object, that I basically use to store the information to be used later as `self.triggers`. This way, we can connect to all the machines as we do during the creation.

The `on_depends` parameter is used to avoid the destroying of resources that are needed to keep the SSH connection used by the provisioner.

I have tested in all of the providers (PAYS images in Azure included) and everything worked fine.


arbulu89:
> @arbulu89 there is one thing I don't quite understand: how is this related to salt? It doesn't look like it executes anything with salt, but it's placed along with the salt provisioner? why is that? also why do we need the salt provisioner enabled to execute this?

Yes, it's true. It doesn't have anything to do with `salt`. In fact, at the beginning it was stored in a new folder `scripts`. The thing is that `aws` forces me to run the new script with root rights. This makes the usage of the `script` parameter in the `remote-exec` impossible. So, I can only use the `inline` option. For this, I need to have the script in the created machines, and in order to avoid the addition of a new provisioner to copy that file, I just stored the new script in `salt` folder. It's just much more convenient. Otherwise, I would need to add something like the next code in all of the files:
```
provisioner "file" {
    source      = "../scripts"
    destination = "/tmp"
  }
```
About the `var.provisioner = "salt"` thing, I just assumed that if we don't set any provisioner, we are not going to run this operation. 
The question here is, if we want to run this always, independently of the used provisioner (or even if we don't do any provisioning). If we decide that always is always (no exception), I could just move the module call to the `main.tf` files.

salt/on_destroy.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+if [[ ! `SUSEConnect -s | grep "Not Registered"` ]];then
```
diegoakechi, line 3: @arbulu89 I think it is recommended to use `$()` instead of backticks.

diegoakechi:
> The question here is, if we want to run this always, independently of the used provisioner (or even if we don't do any provisioning). If we decide that always is always (no exception), I could just move the module call to the `main.tf` files.

For SUSE SCC database is healthy to unregister all the machines that will eventually not be available anymore, since after destroy they cannot be recovered. 

I don't see big problems on doing like that since we don't have other providers at the moment, and salt is responsible to do the registration, but if it simple to be moved to another location, feel free to do so.

Maybe you could consider also to make this script more generic and reuse it instead of creating a new shell script: https://github.com/SUSE/ha-sap-terraform-deployments/blob/develop/salt/install-salt-minion.sh

azure/modules/monitoring/salt_provisioner.tf
```diff
@@ -61,5 +61,14 @@ EOF
       "return_code=$? && sleep 1 && exit $return_code",
     ] # Workaround to let the process start in background properly
   }
+}
 
+module "monitoring_on_destroy" {
+  source               = "../../../generic_modules/on_destroy"
+  node_count           = var.provisioner == "salt" && var.monitoring_enabled ? 1 : 0
```
nick-wang, line 68: A question, why this is `var.monitoring_enabled` but others like drbd use `var.drbd_count` directly?

arbulu89: Because right now, we only can have a unique monitoring machine. So `monitoring_enabled == true` means 1 machine, and 0 otherwise

nick-wang: Got it, thanks.
Change looks good (didn't test though) and the function helpful to me (don't worry about to payment though).

arbulu89:
@stefanotorresi @diegoakechi Changes done.
Now, the `on_destroy` will be always executed and it's attached to the `main.tf` files.
I have removed the `var.provisioner == salt` condition.
And now the `on_destroy.sh` is stored in the terraform module and copied to the machines from there

PR #413, opened 2020-04-02
Fix AWS credential file

ldevulder:
The current code implies to have space around the `=` sign for `aws_access_key_id` and `aws_secret_access_key` variables in `${HOME}/.aws/config`.

This is not mandatory and could lead to issues during deployment (VIP not working for example).

Tested on AWS, not needed on others providers, as this code is specific to AWS.


arbulu89:
@ldevulder Wer can merge this, right?

PR #414, opened 2020-04-03
Add sapcar related pillar entries and variables for Netweaver

Simranpal:
Adding pillar entries and terraform variables related to sapcar functionality for netweaver


Simranpal:
I had one point to discuss about the corresponding HANA pillar entries.
Using the same approach as NW here, for HANA pillar I could create  
1. `hana_global_inst_folder` entry for global software path for all nodes
2. `hana_inst_folder` entry for local software path for a node
3.  sapcar related entries like `hana_sapcar_exe`, `hana_extract_dir` etc

For this, should I make all these paths relative to `hana_inst_media` 
currently we` software path` = `{{ grains['hana_inst_folder'] }}`  as absolute path in multiple places in terraform

pillar_examples/automatic/netweaver/netweaver.sls
```diff
@@ -24,7 +24,16 @@ netweaver:
   sap_adm_password: SuSE1234
   master_password: SuSE1234
   sapmnt_inst_media: "{{ grains['netweaver_nfs_share'] }}"
+  {%- if grains.get('netweaver_swpm_folder', False) %}
```
arbulu89, line 27: For some reason, salt lint doesn't recommend this syntax. We could change by:
`grains['netweaver_swpm_folder']|default(False)`

libvirt/main.tf
```diff
@@ -177,6 +177,9 @@ module "netweaver_node" {
   netweaver_product_id      = var.netweaver_product_id
   netweaver_inst_media      = var.netweaver_inst_media
   netweaver_swpm_folder     = var.netweaver_swpm_folder
+  nw_sapcar_exe             = var.nw_sapcar_exe
```
arbulu89, line 180: Don't use the `nw` prefix. Check how everything else uses `netweaver`. To be aligned, all the parameters should follow the same naming convention.

libvirt/modules/netweaver_node/variables.tf
```diff
@@ -77,6 +77,24 @@ variable "netweaver_swpm_folder" {
...
+}
+
+variable "nw_swpm_sar" {
+  description = "SWPM installer sar archive containing the installer, path relative from the `netweaver_inst_media` mounted point"
+  type        = string
+  default     = ""
+}
+
+variable "nw_swpm_extract_dir" {
+  description = "Extraction path for Netweaver software SWPM folder, if SWPM sar file is provided"
```
arbulu89, line 93: Maybe be should put the default value here (`/sapmedia/NW/SWPM`). Just to provide more information. I know that the salt formula has already this value, but it would be more informative. At least in the description text.

arbulu89:
> I had one point to discuss about the corresponding HANA pillar entries.
> Using the same approach as NW here, for HANA pillar I could create
> 
>     1. `hana_global_inst_folder` entry for global software path for all nodes
> 
>     2. `hana_inst_folder` entry for local software path for a node
> 
>     3. sapcar related entries like `hana_sapcar_exe`, `hana_extract_dir` etc
> 
> 
> For this, should I make all these paths relative to `hana_inst_media`
> currently we` software path` = `{{ grains['hana_inst_folder'] }}` as absolute path in multiple places in terraform

What I would do is just to change the hana pillar file to store individual `software_path` for the nodes, and just use the global one. For our deployment this makes sense.
This way we don't need to create new parameters, and the functionality would be exactly the same. In fact, this is what we do more or less right now. The software folder is the same for all of them. Let's take advantage of the latest improvements in sap formula.

The sapcar related names make sense. Just add a proper description to the variables.

libvirt/main.tf
Simranpal: Yes I was debating on that, and I noticed `nw_ips` but yes it makes sense to go with `netweaver`

libvirt/modules/netweaver_node/variables.tf
Simranpal: We are also setting default from the formula, so it will have a value for sure. But I can add it here too

pillar_examples/automatic/netweaver/netweaver.sls
Simranpal: Hi @arbulu89 I did not get this error after running the salt-lint on pillar sls files. Also the travis build logs did not show me this error. Where did this error show you saw?

Simranpal:
Hi @arbulu89 I created a seperate PR for HANA related changes here: https://github.com/SUSE/ha-sap-terraform-deployments/pull/417

pillar_examples/automatic/netweaver/netweaver.sls
arbulu89: https://github.com/warpnet/salt-lint/wiki/211

arbulu89:
@Simranpal We can merge this, right?

Simranpal:
Hi @arbulu89 i am testing with azure provider now. Once this finishes I will update

Simranpal:
I tested the deployments with libvirt, AWS and Azure providers, and it works as expected.

PR #416, opened 2020-04-03
Feature/improve provision

arbulu89:
This is a pre step to implemente the `reboot` feature.

Mainly:
- I have improved the provisioning script, storing all the logic in a unique sh script and making the steps executable by their own. This will be need for the reboot, as the it will force us to have at least 2 different processes.
- Remove the `salt_provisioner_script.tpl` usage. It was not really needed. Actually, it introduces as security risk, as the `reg_code` is pasted plainly there, and the file remains in the `/tmp` folder
-  I have renamed `pre_installation` to `os_setup`. This salt states group is just to configure the machine before doing anyting (register to scc, update packages and repos, etc)

**Review specially the updated sh script. I don't have much knowledge on that. Maybe some obvious improvements can be done**


arbulu89:
Guys, I have finished the code implementing the `reboot` functionality (reboot after the os setup). 
Do you prefer I add here or wait until this PR is merged and create a new?
It has some few new things, so maybe is too much (even though everything is quite related).

stefanotorresi:
@arbulu89 I guess you could draft a new PR branching from this one and remove the draft status once this is merged.

arbulu89:
> The changes look good to me but, since we're dealing with some boilerplate, I wanted to ask what do you guys think if we moved the logs into `/var/log` as usual, because looking for them in `/tmp` always threw me a little off.

Agree! If you don't mind I will do it in incoming PR about the `reboot` as I have moved some commands (just calling once some of them).

stefanotorresi:
yeah, fair enough!

salt/provision.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash -xe
```
MalloZup, line 1: can we comment on highlevel manner what the script does without reading 200 lines?

I think we  should do 2 phrases commenting the goal of it purpose

diegoakechi: @arbulu89 Yes, now I think this script deserves a header...

PR #418, opened 2020-04-08
Feature/reboot implementation

arbulu89:
Machines reboot implementation after configuring the OS.

After readjusting how we run the provisioning, I have added the reboot action in a new terraform module, keeping the option to run everything on background.

The logic is in `generic_modules/salt_provisioner/main.tf`. There, we have 2 different options, the background and normal normal/live execution. In this 2nd one, the salt executions are splited in 2 different actions, and between them I run the reboot.
I have to do to run the `reboot` in a new provisioner, because I need the `on_failure` option. The reboot will always make the provisioner fail. More than that, to confirm that the next terraform provisioner doesn't create the SSH connection, I need to stop `sshd` explicitiliy.

Besides that, I have implemented some other small changes:
- Change log files from `/tmp` to `/var/log`
- Fix GCP on_destroy functionality creating dependencies between the terraform resources
- Format the terraform files of the modules


aws/modules/netweaver_node/variables.tf
```diff
@@ -76,16 +76,16 @@ variable "aws_credentials" {
...
 
 variable "aws_secret_access_key" {
-  type    = string
+  type = string
 }
 
 variable "s3_bucket" {
   description = "S3 bucket where Netwaever installation files are stored"
-  type    = string
+  type        = string
```
Simranpal, line 88: I am wondering if terraform fmt does not fix these formatting..

gcp/modules/netweaver_node/variables.tf
```diff
@@ -190,3 +190,9 @@ variable "background" {
   type        = bool
   default     = false
 }
+
+variable "on_destroy_dependencies" {
+  description = "Resources objects need in the on_destroy script (everything that allows ssh connection)"
```
Simranpal, line 195: is it `need/needed/`

aws/modules/netweaver_node/variables.tf
arbulu89: The `terraform fmt` command only affects to the files where the command is executed apparently, so as these files are in subfolders, it wasn't applied here.

Simranpal: Yes the `terraform fmt` works in cwd and not sub-directories

PR #421, opened 2020-04-09
Remove raw and template

### LINKED ISSUE #386: Alert.yaml file should be plenty data only
https://github.com/SUSE/ha-sap-terraform-deployments/commit/07952934aad8f405d40890c00ad4e9cf7d110108

we should remove `raw` so user can copy and paste directly

MalloZup:
fix #386


diegoakechi:
@MalloZup Did you test this? I think it will not work. The https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/salt/monitoring/prometheus/prometheus.yml has Jinja interpolations on it, so I think this change will copy the file as it is and the prometheus server will not start. Or maybe I am missing something here...

MalloZup:
the change works itself. the `prometheus.yaml` file we cannot avoid jinja there.

I will do a more refactor for making it clear

salt/monitoring/init.sls
```diff
@@ -20,22 +20,31 @@ prometheus:
...
-    - name: /etc/prometheus/
-    - makedirs: True
-    - source: salt://monitoring/prometheus
+  file.managed:
+    - name:  /etc/prometheus/prometheus.yml
+    - source: salt://monitoring/prometheus/prometheus.yml.j2
     - template: jinja
-    - include_empty: True
+    - require:
+      - pkg: prometheus
```
stefanotorresi, line 36: Can we get rid of the `shap` prefix from the name of these two states? We don't really need that and external contributors won't know what that means.

MalloZup: make sense thx!

PR #423, opened 2020-04-10
Feature/aws terraform modules

Simranpal:
Creating this PR for AWS monitoring and hana_node modules.
For iscsci I created PR earlier: https://github.com/SUSE/ha-sap-terraform-deployments/pull/410

Note, I change the default instance type for hana isntances from `"r3.8xlarge"` to `"m4.2xlarge" `


Simranpal:
There is dependency between resources of a submodule `hana_node` and ones defined in `network.tf`. Within the changes now, it looks bit different from netweaver module where all subnets are routes are defined within that module.

For hana_node, basically, I moved `hana-cluster-vip` and clusternodes `aws_instance` into the hana_node module, but I left these resources: `netweaver-subnet `and `netweaver-subnet-route-association` in`network.tf`,
because the iscsi and monitoring modules resources depend on it. One way was to expose them as outputs, but I followed structure same as in azure. Let me know what you think.

aws/main.tf
```diff
@@ -47,3 +47,84 @@ module "netweaver_node" {
...
+  aws_secret_access_key  = var.aws_secret_access_key
+  host_ips               = var.host_ips
+  hana_data_disk_type    = var.hana_data_disk_type
+  hana_inst_master       = var.hana_inst_master
+  hana_inst_folder       = var.hana_inst_folder
+  hana_disk_device       = var.hana_disk_device
+  hana_fstype            = var.hana_fstype
+  hana_cluster_vip       = var.hana_cluster_vip
+  private_key_location   = var.private_key_location
+  iscsi_srv_ip           = aws_instance.iscsisrv.private_ip
```
arbulu89, line 78: This won't work as soon as you merge the `iscsi` server module

aws/salt_provisioner.tf
```diff
@@ -59,130 +59,3 @@ EOF
     ] # Workaround to let the process start in background properly
```
arbulu89, line 59: I guess this file will be remove as soon as `iscsi` server module PR is merged

aws/terraform.tfvars.example
```diff
@@ -1,7 +1,7 @@
 # Launch SLES-HAE of SLES4SAP cluster nodes
 
 # Instance type to use for the hana cluster nodes
-instancetype = "r3.8xlarge"
+instancetype = "m4.2xlarge"
```
arbulu89, line 4: Why have you changed this?
I think it was with this value because is the SAP certified instance type

aws/network.tf
```diff
@@ -43,7 +41,7 @@ resource "aws_route_table" "route-table" {
 }
 
 resource "aws_route_table_association" "hana-subnet-route-association" {
-  count          = var.ninstances
+  count          = var.hana_count
```
arbulu89, line 44: This `hana-subnet-route-association` should be moved to the `hana` module

aws/modules/hana_node/main.tf
```diff
@@ -0,0 +1,61 @@
+module "sap_cluster_policies" {
+  enabled           = var.hana_count > 0 ? true : false
+  source            = "../../modules/sap_cluster_policies"
+  name              = var.name
+  aws_region        = var.aws_region
+  cluster_instances = aws_instance.clusternodes.*.id
+  route_table_id    = var.route_table_id
+}
+
+resource "aws_route" "hana-cluster-vip" {
```
arbulu89, line 10: This has to be created only conditionally, if we have at least one `hana` nodes

arbulu89:
> There is dependency between resources of a submodule `hana_node` and ones defined in `network.tf`. Within the changes now, it looks bit different from netweaver module where all subnets are routes are defined within that module.
> 
> For hana_node, basically, I moved `hana-cluster-vip` and clusternodes `aws_instance` into the hana_node module, but I left these resources: `netweaver-subnet `and `netweaver-subnet-route-association` in`network.tf`,
> because the iscsi and monitoring modules resources depend on it. One way was to expose them as outputs, but I followed structure same as in azure. Let me know what you think.

I have shared feelings with this. I would like to have the subnets creation in the `hana` module itself, as it belongs to it. As you said, other auxiliar nodes as iscsi and monitoring are using this subnets.
By now, we can let as you did. In the future

> There is dependency between resources of a submodule `hana_node` and ones defined in `network.tf`. Within the changes now, it looks bit different from netweaver module where all subnets are routes are defined within that module.
> 
> For hana_node, basically, I moved `hana-cluster-vip` and clusternodes `aws_instance` into the hana_node module, but I left these resources: `netweaver-subnet `and `netweaver-subnet-route-association` in`network.tf`,
> because the iscsi and monitoring modules resources depend on it. One way was to expose them as outputs, but I followed structure same as in azure. Let me know what you think.

I have shared feelings on that. The subnet should belong to the hana nodes, but as you said it is used by other machines. By now, we can leave as you did. In the future we can change if we find a better solution.

aws/terraform.tfvars.example
Simranpal: Hi @arbulu89 yes I checked now and the value is `r3.8xlarge` as per the aws doc. So, I have corrected it back. I changed to `m4.2xlarge` earlier because in our aws readme/ Advanced Usage we mentioned `m4.2xlarge` will be deployed by default. Maybe it needs correction there too?

aws/network.tf
Simranpal: as discussed, due to its usage in `on_destroy_dependencies` in `main.tf` I will keep it as such

aws/modules/hana_node/main.tf
Simranpal: I did not understand this @arbulu89. We did not check such condition earlier as well

Simranpal:
Hi @arbulu89 I did some changes.Also moved the remaining resources to infrastructure.tf. Let me know what you think

Simranpal:
Tested now with aws deployment and works as expected!

PR #424, opened 2020-04-10
Register needed extension by default jira#709

MalloZup:
# description:

see jira.


MalloZup:
tested on sle15 works. I will test just for sanity on sle12

salt/os_setup/registration.sls
```diff
@@ -1,6 +1,8 @@
 {% if grains['os_family'] == 'Suse' %}
 {% if not grains.get('qa_mode') or '_node' not in grains.get('role') %}
 {% if grains['reg_code'] %}
+{% set mod_reg_code = grains['reg_code'] %}
```
arbulu89, line 4: Why do you call it `mod_reg_code` to this? Call it `reg_code` in any case. And if you do this, replace all the following `grains['reg_code']` by it

salt/os_setup/registration.sls
```diff
@@ -11,6 +13,57 @@ register_system:
         interval: 15
 {% endif %}
 
+
+{% if not 'iscsi_srv' in grains.get('role') %}
```
arbulu89, line 17: Why are you excluding the `iscsi_srv` here? And only for SLE12?
I would let the `iscsi` server to be registered, I don't see anything wrong with that

salt/os_setup/registration.sls
```diff
@@ -11,6 +13,57 @@ register_system:
...
+{% if '12' in grains['osrelease'] %}
+{% if grains['osrelease'] == '12' %}
```
arbulu89, line 19: Don't use `osrelease`. Use `osrelease_info` or `osmajorrelease` better. It returns a list with OS release numbers. Example
`[12, 4]` or `[15]`

salt/os_setup/registration.sls
```diff
@@ -11,6 +13,57 @@ register_system:
...
+    - retry:
+        attempts: 3
+        interval: 15
+
+{% endif %}
+{% endif %}
+
+# sle15 
+# temporary PackageHub (temporarily for GCP until ECO-1148 is released)
+{% if '15' in grains['osrelease'] %}
```
arbulu89, line 53: The same than for SLE12, don't use `osrelease`.
In fact, we don't need to duplicate this `PackageHub` code. If it's needed by SLE12 and SLE15, and both use the same code, just use a unique state, the `osrelease` and `arch` will do the work for that.

salt/os_setup/registration.sls
```diff
@@ -11,6 +13,57 @@ register_system:
...
+  cmd.run:
+    - name: /usr/bin/SUSEConnect -p sle-module-public-cloud/{{ grains['osrelease'] }}/{{ arch }} {{ "-r $mod_reg_code" if mod_reg_code else "" }}
+    - env:
+        - mod_reg_code: {{ mod_reg_code }}
+    - retry:
+        attempts: 3
+        interval: 15
+
+# temporary PackageHub (temporarily for GCP until ECO-1148 is released)
+default_PackageHub_registration_sle12:
```
arbulu89, line 39: Why are you adding this `sle12` sufix to the state name? I think it's not needed and breaks the naming used in the other registration names

salt/os_setup/registration.sls
```diff
@@ -1,16 +1,51 @@
...
     - env:
-        - reg_code: {{ grains['reg_code'] }}
+        - reg_code: {{ reg_code }}
     - retry:
         attempts: 3
         interval: 15
 {% endif %}
 
+
+{% if 12 == grains['osmajorrelease'] %}
```
arbulu89, line 17: This order is a bit strange. Usually I would put `{% if grains['osmajorrelease'] == 12 %}`

MalloZup:
@arbulu89 let me know thx!

salt/os_setup/registration.sls
```diff
@@ -1,16 +1,67 @@
...
-        - reg_code: {{ grains['reg_code'] }}
+        - reg_code: {{ reg_code }}
     - retry:
         attempts: 3
         interval: 15
 {% endif %}
 
+
+{% if grains['osmajorrelease'] == 12 %}
+# hardcode the 12 version number for the 2 follo
```
arbulu89, line 18: I think that the comment is unfinished

salt/os_setup/registration.sls
```diff
@@ -1,16 +1,67 @@
...
+    - env:
+        - reg_code: {{ reg_code }}
+    - retry:
+        attempts: 3
+        interval: 15
+
+{% endif %}
+{% endif %}
+
+{% if grains['osmajorrelease'] == 15 %}
```
arbulu89, line 41: I think we can change the `if` by an `elif` combining with the previous `if`. It cannot be anything else than SLE12 or SLE15. 
Besides, we can have the provider condition in the same line too. Something like:
```
{% elif grains['osmajorrelease'] == 15 and grains['provider'] in ['gcp', 'aws', 'azure'] %}
```
If you do this, don't forget to remove the previous `endif`

salt/os_setup/registration.sls
MalloZup: ah yes forgot it

salt/os_setup/registration.sls
```diff
@@ -1,16 +1,64 @@
...
         attempts: 3
         interval: 15
 {% endif %}
 
+
+{% if grains['osmajorrelease'] == 12 %}
+# hardcode the 12 version number for the 2 following modules, since they don't offer a sp version only 1. 
+default_sle_module_adv_systems_management_registration:
+  cmd.run:
+    - name: /usr/bin/SUSEConnect -p sle-module-adv-systems-management/12/{{ arch }} {{ "-r $reg_code" if reg_code else "" }}
```
arbulu89, line 21: By the way @MalloZup , we can replace `{{ "-r $reg_code" if reg_code else "" }}` by `-r $reg_code` simply. If we are in this piece of code the `reg_code` always exists.

MalloZup: yes indeed thx!

MalloZup: done

PR #427, opened 2020-04-14
Hotfix/fix log output

arbulu89:
The creation of the files in `/var/log` fixed. I didn't consider that `>>` are not executed as `root` user, so in `AWS` and `Azure` where we don't use `root` user by default, the scripts where failing due `permission denied`.

This change adds the option to the script to specify the output logfile natively, without redirecting. This way we can run it everything as `root`.

Besides, that I have taken advantage to increase `join_timeout` value for the HANA cluster, as I have noticed that the execution failed twice for that reason in our Jenkins builds

### DIFF
--- a/generic_modules/salt_provisioner/main.tf
+++ b/generic_modules/salt_provisioner/main.tf
@@ -46,7 +46,7 @@ resource "null_resource" "provision" {
 
   provisioner "remote-exec" {
     inline = [
-      "sudo sh /tmp/salt/provision.sh -so > /var/log/provisioning.log",
+      "sudo sh /tmp/salt/provision.sh -sol /var/log/provisioning.log",
     ]
   }
 
@@ -59,7 +59,7 @@ resource "null_resource" "provision" {
 
   provisioner "remote-exec" {
     inline = [
-      "sudo sh /root/salt/provision.sh -pdq >> /var/log/provisioning.log",
+      "sudo sh /root/salt/provision.sh -pdql /var/log/provisioning.log",
     ]
   }
 }
--- a/generic_modules/salt_provisioner/main.tf
+++ b/generic_modules/salt_provisioner/main.tf
@@ -19,7 +19,7 @@ resource "null_resource" "provision_background" {
 
   provisioner "remote-exec" {
     inline = [
-      "nohup sudo sh /tmp/salt/provision.sh > /var/log/provisioning.log &",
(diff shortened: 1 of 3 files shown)


generic_modules/salt_provisioner/main.tf
```diff
@@ -46,7 +46,7 @@ resource "null_resource" "provision" {
 
   provisioner "remote-exec" {
     inline = [
-      "sudo sh /tmp/salt/provision.sh -so > /var/log/provisioning.log",
+      "sudo sh /tmp/salt/provision.sh -sol /var/log/provisioning.log",
```
MalloZup, line 49: ``` sudo sh /tmp/salt/provision.sh -sol /var/log/provisioning.log```

could be 

```
# 
sudo sh /tmp/salt/provision.sh -sol /var/log/provisioning.log
```

arbulu89: Everything is documented in the `help` output. I don't thing we need to put so many comments there.

MalloZup: We could have long arguments like `--boostrap` instead of `s` which is more expressive. 

My use-case for this is when people are reviewing or looking at terraform code, without having to `jump` contextually to a scipt arguments helper or code.

arbulu89: This would make the code more complex, and for an internal script, that should be only used during the bootstrap I don't think that's neccesary.
If more people agree, I can add some comments as you asked initially.

MalloZup: I think I'm not suggesting to support both.
I would just instead of using `l` shortcut we just support `log` the fullname . Since the internal script is used in automation, is better to use full-names . We can avoid shortcut name like `l` etc.

MalloZup: but I approved this PR. to me we can create a follow-up issue for this, if team agree

stefanotorresi: While I understand Dario's sentiment, having a more comprehensive CLI is much more difficult to do and I don't think plain `getopts` would be enough. I'd suggest to keep it simple for now. ;)

MalloZup: I approved the PR , so fo me isn't a blocking issue. rather a follow-up for more clarity, since is more human readable

PR #428, opened 2020-04-14
Update anything resource by azure-lb in drbd template

arbulu89:
Update load-balancer resource agent.
I have kept the `rsc_socat` nomenclature as in SUSE distros it uses socat instead `nc`, so it makes sense (replacing by `nc` as the guide suggests is confusing)
https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-high-availability


arbulu89:
> The guide looks a bit confusing, but it seems to me, using socat is fine.
> What is the the `anything` resource used for?

The `anything` resource is a really generic resource, where you can configure a binary and the parameters, that will make the call. The `azure-lb` makes this basically, so there is not any change in the functionality. I guess that having this way have some advantage though (they can choose between `nc` and `socat` automatically for example)

PR #430, opened 2020-04-15
Add or improve terraform variables description

arbulu89:
Description of the variables added or improved as we removed the variables list from the READMEs.


azure/variables.tf
```diff
@@ -221,47 +229,57 @@ variable "hana_extract_dir" {
 }
 
 variable "hana_disk_device" {
-  description = "device where to install HANA"
+  description = "Device where hana is installed"
```
Simranpal, line 232: I guess the previous description was good too, since we are setting these variable _before_ installation

Simranpal: Same for some other variables

azure/variables.tf
```diff
@@ -427,8 +461,9 @@ variable "netweaver_data_disk_size" {
 }
 
 variable "netweaver_data_disk_caching" {
-  type    = string
-  default = "ReadWrite"
+  description = "Diss caching of the disks used to store hana database content"
```
Simranpal, line 464: `Diss/Disk`

gcp/variables.tf
```diff
@@ -1,32 +1,33 @@
...
-  type    = string
-  default = "https://storage.googleapis.com"
+  description = "Path to a SSH private key used to connect to the created machines"
+  type        = string
 }
 
 variable "ip_cidr_range" {
-  description = "internal IPv4 range"
+  description = "Internal IPv4 range of the created network"
+  default     = "10.0.0.0/24"
```
Simranpal, line 30: If we specify this range, as default,  do we also set or mention that the host IPs need to be this range( I see its same range in example)

libvirt/variables.tf
```diff
@@ -81,7 +76,7 @@ variable "background" {
 # Hana related variables
 
 variable "base_image" {
-  description = "Image of the sap hana nodes"
+  description = "qcow2 image used to create the hana machines"
```
Simranpal, line 79: is it always qcow images

libvirt/variables.tf
```diff
@@ -253,15 +253,15 @@ variable "drbd_ips" {
 }
 
 variable "drbd_shared_storage_type" {
-  description = "used shared storage type for fencing (sbd) for DRBD cluster. Available options: iscsi, shared-disk."
+  description = "Ssed shared storage type for fencing (sbd) for drbd cluster. Available options: iscsi, shared-disk."
```
Simranpal, line 256: `Ssed/Used`

azure/variables.tf
arbulu89: @Simranpal You are native in English hehe Most probably you would do it better. We can just fix the typos, and you can open a 2nd round of changes with improvements. What do you think?

libvirt/variables.tf
arbulu89: It's just more descriptive hehe I think that this information can be useful

gcp/variables.tf
arbulu89: Yes, good idea. I will complement the other descriptions

azure/variables.tf
Simranpal: Sounds good, not much improvements needed, but I can check

PR #431, opened 2020-04-16
Feature/gcp terraform modules

Simranpal:
Creating this PR for terraform modules creation and refactoring for GCP provider.


gcp/modules/hana_node/main.tf
```diff
@@ -0,0 +1,104 @@
...
+  count = var.hana_count
+  name  = "${terraform.workspace}-${var.name}-hana-software-${count.index}"
+  type  = "pd-standard"
+  size  = "20"
+  zone  = element(var.compute_zones, count.index)
+}
+
+# temporary HA solution to create the static routes, eventually this routes must be created by the RA gcp-vpc-move-route
+resource "google_compute_route" "hana-route" {
+  name                   = "hana-route"
```
arbulu89, line 30: This entry needs a `count` value, it must only be created if `hana_count` is higher than `0`

Simranpal:
Hi @arbulu89 I have done some further changes with var.name usage> If this looks look good, for GCP I can remove the variable `name` definition as well as we won't need that. Also tested the deployment and looks good

gcp/modules/iscsi_server/variables.tf
```diff
@@ -0,0 +1,99 @@
+variable "name" {
```
arbulu89, line 1: Remove this `name`, we are not using it and it doesn't have any sense in the `iscsi` module

gcp/modules/monitoring/variables.tf
```diff
@@ -0,0 +1,90 @@
+variable "name" {
```
arbulu89, line 1: Remove the `name` variable

gcp/modules/hana_node/variables.tf
```diff
@@ -0,0 +1,235 @@
+variable "name" {
```
arbulu89, line 1: Yes, I would remove the `name` variable. We are not using it in the other modules (unfortunately we use it `aws`...).
I think we can remove it as it doesn't provide any real use case

PR #434, opened 2020-04-16
Azure: Feature/improve ip management and accept existing vNet And Subnets

arbulu89:
This PR brings 2 important features for *Azure* but It doesn't have any non backward compatible change.

1. Use already existing resource group, virtual network and subnets. This is a quite requested feature as network administration is managed by separated teams many times.
2. Auto generate the ip addresses (optionally, they old way still has the preference) to make the deployment easier.

And other smaller but still important change.
Now the hana and drbd cluster virtual addresses are configurable from variables (they can be auto generated too).

The idea is to continue adding the same feature for the other providers.


stefanotorresi:
@arbulu89 have you looked into the terraform network address functions like [`cidrhost`](https://www.terraform.io/docs/configuration/functions/cidrhost.html) ?
I think you can safely use just a starting subnet and then automatically increment the IP addresses.

azure/infrastructure.tf
```diff
@@ -11,15 +11,43 @@ terraform {
...
+}
+
+locals {
+  resource_group_name = var.resource_group_name == "" ? azurerm_resource_group.myrg.0.name : var.resource_group_name
+  vnet_name           = var.vnet_name == "" ? azurerm_virtual_network.mynet.0.name : var.vnet_name
+  subnet_id = var.subnet_name == "" ? azurerm_subnet.mysubnet.0.id : format(
+  "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Network/virtualNetworks/%s/subnets/%s", data.azurerm_subscription.current.subscription_id, var.resource_group_name, var.vnet_name, var.subnet_name)
+  # If vnet_name is not defined, a new vnet is created
+  # If vnet_name is defined, and the vnet_address_range is empty, it will try to get the ip range from the real vnet using the data source. If vnet_address_range is defined it will use it
+  vnet_address_range   = var.vnet_name == "" ? var.vnet_address_range : (var.vnet_address_range == "" ? data.azurerm_virtual_network.mynet.0.address_space.0 : var.vnet_address_range)
```
Simranpal, line 37: Can we store the data source related info in a variable, although I see its not used in much places. It will maybe look clean. Current solution is fine though.

azure/main.tf
```diff
@@ -63,14 +86,15 @@ module "netweaver_node" {
   netweaver_swpm_extract_dir    = var.netweaver_swpm_extract_dir
   netweaver_sapexe_folder       = var.netweaver_sapexe_folder
   netweaver_additional_dvds     = var.netweaver_additional_dvds
-  netweaver_nfs_share           = "10.74.1.201:/HA1" # drbd cluster ip address is hardcoded by now
+  netweaver_nfs_share           = "${local.drbd_cluster_vip}:/HA1"
```
Simranpal, line 89: I am thinking for netweaver SID `HA1` we can avoid hardcoding it

azure/infrastructure.tf
```diff
@@ -31,30 +59,32 @@ resource "azurerm_storage_account" "mytfstorageacc" {
 
 # Network resources: Virtual Network, Subnet
 resource "azurerm_virtual_network" "mynet" {
+  count               = var.vnet_name == "" ? 1 : 0
```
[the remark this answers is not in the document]
arbulu89: Yes

azure/infrastructure.tf
arbulu89: It's already stored in `data.azurerm_virtual_network.mynet`. Why do we need to store it somewhere else?

azure/main.tf
arbulu89: Yes, for that we would need to make the `sid` variable. Now the pillar files are even hardcoded. It would be a good improvement for the future!

PR #438, opened 2020-04-20
Use vpc_security_group_ids instead of security_groups

arbulu89:
This PR fixes the issue in AWS that every apply, recreates the whole deployment because the security groups must be re-attached.

```
 # module.iscsi_server.aws_instance.iscsisrv must be replaced
-/+ resource "aws_instance" "iscsisrv" {
        ami                          = "ami-024f50fdc1f2f5603"
...
      ~ public_ip                    = "18.196.254.113" -> (known after apply)
      ~ security_groups              = [ # forces replacement
          + "sg-0027a15d0a1fe9f43",
        ]
        source_dest_check            = true
```

`security_groups` usage is only used when the default VPC is used. For any other circunstance, we need to use `vpc_security_group_ids` as this will only bring the correct resource.

More info: https://www.terraform.io/docs/providers/aws/r/instance.html#vpc_security_group_ids


arbulu89:
> Looks good! Thanks @arbulu89
> For our use case, we do not need to create/use the default VPCs ?
> I guess not since we are always creating our Instances in a VPC and if we want to reuse the existing VPC, then we can use the one created before, right?

As far as I see it should even work for the default VPC (the documentation doesn't say the opposite). But we can test just in case.

PR #439, opened 2020-04-20
Remove hardcoded storage devices on GCE

ldevulder:
Storage devices in GCE are currently set either in the tfvars file or hardcoded in TF/sls files. This is bad because device name in Linux are handled by udev and the naming is dynamic (`/dev/sdb` could be `/dev/sda` sometimes).

This commit removes all of these by using GCE functionalities to get the correct device name.

Manual tests done successfully and through [openQA tests](http://1b210.qa.suse.de/tests/6633) also.

### DIFF
--- a/pillar_examples/automatic/drbd/drbd.sls
+++ b/pillar_examples/automatic/drbd/drbd.sls
@@ -1,3 +1,5 @@
+{% set drbd_disk_device = salt['cmd.run']('realpath '~grains['drbd_disk_device']) %}
+
 drbd:
   promotion: {{ grains['name_prefix'] }}01
 
@@ -62,7 +64,7 @@ drbd:
   resource:
     - name: "sapdata"
       device: "/dev/drbd1"
-      disk: {{ grains['drbd_disk_device'] }}1
+      disk: {{ drbd_disk_device }}1
 
       file_system: "xfs"
       mount_point: "/mnt/sapdata/HA1"
--- a/salt/hana_node/download_hana_inst.sls
+++ b/salt/hana_node/download_hana_inst.sls
@@ -8,13 +8,14 @@ download_files_from_s3:
 
 {% elif grains['provider'] == 'gcp' %}
 
+{% set hana_inst_disk_device = salt['cmd.run']('realpath '~grains['hana_inst_disk_device']) %}
 hana_inst_partition:
   cmd.run:
     - name: |
-        /usr/sbin/parted -s {{ grains['hana_inst_disk_device'] }} mklabel msdos && \
-        /usr/sbin/parted -s {{ grains['hana_inst_disk_device'] }} mkpart primary ext2 1M 100% && sleep 1 && \
-        /sbin/mkfs -t xfs {{ grains['hana_inst_disk_device'] }}1
-    - unless: ls {{ grains['hana_inst_disk_device'] }}1
+        /usr/sbin/parted -s {{ hana_inst_disk_device }} mklabel msdos && \
+        /usr/sbin/parted -s {{ hana_inst_disk_device }} mkpart primary ext2 1M 100% && sleep 1 && \
+        /sbin/mkfs -t xfs {{ hana_inst_disk_device }}1
+    - unless: ls {{ hana_inst_disk_device }}1
     - require:
       - pkg: parted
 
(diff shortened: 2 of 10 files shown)


salt/hana_node/download_hana_inst.sls
```diff
@@ -8,13 +8,14 @@ download_files_from_s3:
 
 {% elif grains['provider'] == 'gcp' %}
 
+{% set drbd_disk_device = salt['cmd.run']('realpath '~grains['drbd_disk_device']) %}
```
arbulu89, line 11: Is this a typo? Why do we have `drbd_disk_device` in hana salt execution?

pillar_examples/automatic/drbd/drbd.sls
```diff
@@ -1,3 +1,5 @@
+{% set drbd_disk_device = salt['cmd.run']('realpath '~grains['drbd_disk_device']) %}
```
arbulu89, line 1: I wouldn't put this in the template itself. I would run this command in the salt state where the rendering of the template is done, and pass it as context.

salt/hana_node/download_hana_inst.sls
```diff
@@ -8,13 +8,14 @@ download_files_from_s3:
...
 hana_inst_partition:
   cmd.run:
     - name: |
-        /usr/sbin/parted -s {{ grains['hana_inst_disk_device'] }} mklabel msdos && \
-        /usr/sbin/parted -s {{ grains['hana_inst_disk_device'] }} mkpart primary ext2 1M 100% && sleep 1 && \
-        /sbin/mkfs -t xfs {{ grains['hana_inst_disk_device'] }}1
-    - unless: ls {{ grains['hana_inst_disk_device'] }}1
+        /usr/sbin/parted -s {{ hana_inst_disk_device }} mklabel msdos && \
```
arbulu89, line 15: Where is the `hana_inst_diks_device` coming from? Is `{{ grains['hana_inst_disk_device'] }}` ?

salt/hana_node/download_hana_inst.sls
ldevulder: More a typo here I think, I will check.

salt/hana_node/download_hana_inst.sls
ldevulder: Related to the typo you mentioned above I think :D

pillar_examples/automatic/drbd/drbd.sls
ldevulder: I could, but this shouldn't change the result.

ldevulder:
> I wonder if we could do something similar for the other cloud providers. Using the name rather than the mounting point is much better.

It could be possible for AWS as far as I saw. In fact in AWS it seems that you can choose the device name, not exactly the same but at the end the idea is to be sure of the device name.

For Azure I didn't had a look yet.

pillar_examples/automatic/drbd/drbd.sls
ldevulder: Question: where is the salt state you mentioned?

salt/hana_node/download_hana_inst.sls
ldevulder: Fixed!

salt/hana_node/download_hana_inst.sls
ldevulder: Fixed!

ldevulder:
> PD: As far as I have checked, this change shouldn't affect to the other providers where we still use `/dev/sdx` type, right?

For now yes. As I said could be possible with AWS and I didn't check for Azure yet. But this is something I can have  a look after.

arbulu89:
> > PD: As far as I have checked, this change shouldn't affect to the other providers where we still use `/dev/sdx` type, right?
> 
> For now yes. As I said could be possible with AWS and I didn't check for Azure yet. But this is something I can have a look after.

I don't mean if we can implement in other provides, I mean that the change is harmless for them (as they still use the old way, I want to make sure that this change doesn't break anything in other providers, that's it)

ldevulder:
> I don't mean if we can implement in other provides, I mean that the change is harmless for them (as they still use the old way, I want to make sure that this change doesn't break anything in other providers, that's it)

Sorry, I misread :-)

Yes this change doesn't affect other providers, it's only for GCE.

pillar_examples/automatic/drbd/drbd.sls
ldevulder: @arbulu89 do you want me to try a change? Or do we keep it as is?

arbulu89:
@ldevulder Feel free to merge if you think everything is fine. I plan to create a release at the end of this week (or next week), bringing develop to master.

PR #440, opened 2020-04-20
Remove hardcoded SBD device

ldevulder:
Remove the hardcoded SBD device for all PC providers, by using the ~first available~ provided index device for iSCSI device if `sbd_disk_device` is not set.


arbulu89:
@ldevulder I have questions. What is the output of `lsscsi | awk "/LIO-ORG/ { print \$NF; exit }"`.
I'm concerned about the next thing.
Right now, in the iscsi service to serve 3 disks and all the cluster nodes (hana, drbd and nw) are connecting to this iscsi service and retrieving this disks and mounting in `/dev`. The idea is that each cluster uses a different disk. 
So, if iscsi provides `/dev/sda`, `/dev/sdb` and `/deb/sdc` (the names will be different in each node), what is this solution doing?
I guess they will connect to the first one, right?
If this is true, the 3 clusters will be using the same disk for sbd. Even though this is OK for sbd, it breaks the idea of using 3 disks.

ldevulder:
An idea could be to do something like `lsscsi | awk "/LIO-ORG/ { if (NR==$lun_index) print \$NF }"`, with `$lun_index` (or another variable name) the index of the sbd disk to use (1 for hana, 2 for drbd, 3 for netweaver, etc.). And set this index in the TF file. And, of course, use the current behavior if the device name is explicitly set.

In that case we keep the current implementation but without specifying the device name. @arbulu89 WDYT?

libvirt/modules/netweaver_node/salt_provisioner.tf
```diff
@@ -45,7 +45,7 @@ pas_instance_number: ${var.pas_instance_number}
 aas_instance_number: ${var.aas_instance_number}
 ha_sap_deployment_repo: ${var.ha_sap_deployment_repo}
 shared_storage_type: shared-disk
-sbd_disk_device: /dev/vdb1
+sbd_disk_device: "${var.shared_storage_type == "shared-disk" ? "/dev/vdb1"}"
```
alvarocarvajald, line 48: Is it OK that in all other files `sbd_disk_device` references a full disk, but here it references a partition?

alvarocarvajald, line 48: Is it OK that in all other files `sbd_disk_device` references a full disk, but here it references a partition?

arbulu89: Yes, in netweaver the shared disk usage is a bit different. We don't only use the disk for sbd, we use to store other shared files between ASCS and ERS, that's why the disk is splitted in multiple partitions

libvirt/modules/drbd_node/salt_provisioner.tf
```diff
@@ -32,7 +32,7 @@ provider: libvirt
 role: drbd_node
 drbd_disk_device: /dev/vdb
 shared_storage_type: ${var.shared_storage_type}
-sbd_disk_device: "${var.shared_storage_type == "iscsi" ? "/dev/sda" : "/dev/vdc"}"
+sbd_disk_device: "${var.shared_storage_type == "shared-disk" ? "/dev/vdc"}"
```
arbulu89, line 35: @ldevulder Have you tested this condition? What is the output of this `if` if the conditions doesn't match. I think that you need to put the `else` part. Something like: 
```
sbd_disk_device: "${var.shared_storage_type == "shared-disk" ? "/dev/vdc" : ""}"
```

arbulu89:
> An idea could be to do something like `lsscsi | awk "/LIO-ORG/ { if (NR==$lun_index) print \$NF }"`, with `$lun_index` (or another variable name) the index of the sbd disk to use (1 for hana, 2 for drbd, 3 for netweaver, etc.). And set this index in the TF file. And, of course, use the current behavior if the device name is explicitly set.
> 
> In that case we keep the current implementation but without specifying the device name. @arbulu89 WDYT?

@ldevulder Yes, it looks like a better approach. I'm  wondering if passing this index number hardcoded from terraform makes sense. Maybe we should create an output value in the terraform `iscsi_server` with the number of luns that we create there, and based on that number provide the lun index to salt. In fact, we shouldn't have the partition numbers hardcoded in terraform...

Maybe why now the easier thing is just to put the index hardcoded, and a comment why we use this way.

libvirt/modules/drbd_node/salt_provisioner.tf
ldevulder: No I didn't checked, as I can't on libvirtd on my lab. But I think you are right, otherwise the variable should be undefined an not empty.

ldevulder:
> @ldevulder Yes, it looks like a better approach.

It seems to work on my lab, I will update the PR later today I think.

>  I'm wondering if passing this index number hardcoded from terraform makes sense.

We should have a way to identify the differents LUN.

> Maybe we should create an output value in the terraform `iscsi_server` with the number of luns that we create there, and based on that number provide the lun index to salt.

Sounds a good idea, but I don't know how to set the correct LUN index to each cluster, but my Salt knowledge is limited :)

> In fact, we shouldn't have the partition numbers hardcoded in terraform...

Yes it's not so good, but a little bit better than hardcoded the device :)

> Maybe why now the easier thing is just to put the index hardcoded, and a comment why we use this way.

Maybe for now yes.

ldevulder:
> The idea is that each cluster uses a different disk.

I just saw that on GCP booth HANA and NetWeaver cluster are using the same `/dev/sde` device... so we have also an issue with the current code.

arbulu89:
> > The idea is that each cluster uses a different disk.
> 
> I just saw that on GCP booth HANA and NetWeaver cluster are using the same `/dev/sde` device... so we have also an issue with the current code.

No, they use different disks. They might use the same disk for that hehe. As you noticed in this PR, using `/dev/sdx` kind of names if not good, because each machine can mount the same disk with different names depending on how many disks they have. It's quite confusing, but it works fine. 
That's why making the transition to use the disks by name is better.
In fact, It would be better to provide to sbd the disk name reather than the mounting name

libvirt/modules/drbd_node/salt_provisioner.tf
ldevulder: Changes done.

ldevulder:
@arbulu89 Is is ok for you now?

ldevulder:
> If you tested (at least in 1 of the providers) I think we are ready to merge.

I tested on the 3 providers but with the master branch (as there is too much changes/failure in develop :))

PR #442, opened 2020-04-21
Update software_path entry for hana pillar examples

Simranpal:
Removing the hardcoded default software_path for hana media. It now fetches the value from terraform variable `hana_inst_folder`
Tested with deployments providing with/without `hana_platform_folder`


Simranpal:
Yes @arbulu89 I checked all providers and we are setting this grains entry!

Issue #402, opened 2020-03-25
FEATURE request (AWS)


picoroma:
Do you have in the pipeline to deploy this scenario into a pre-existing VPC ?
If, yes - when ?
THX in advance

arbulu89:
Hi @picoroma,
Thanks for the ticket. Yes, it's in our backlog, but we are not going to work on it most probably the next month at least.

PD:
About using an existing VPC, consider that for that terraform needs to import all the information from AWS. At this point, terraform will store all the state, so we assume that the resource is managed by terraform now on. This means that any manual change in the VPC, will put the project out of sync, in an unstable situation. 

Just warming you that using terraform with existing resources can have some side-effects.

arbulu89:
Hi @picoroma . I will start working on this task now. I have a question.
Besides the VPC, do you have in mind any other resource that you would like to reuse (already created resources)?
I don't know, the security group, subnets, etc

picoroma:
Subnet could be fine.SG also, but less important I mean...Il 17 apr 2020 15:14, Xabier Arbulu Insausti <notifications@github.com> ha scritto:
Hi @picoroma . I will start working on this task now. I have a question.
Besides the VPC, do you have in mind any other resource that you would like to reuse (already created resources)?
I don't know, the security group, subnets, etc

arbulu89:
@picoroma Forget about that hehe, When I wrote it I didn't have all the knowledge. Now I know that we can use an already existing VPC without any risk.

picoroma:
👍Let me know when is possible to test itP.

arbulu89:
Issue fixed. It will be available in version `5.0.0`

PR #443, opened 2020-04-21
Feature/aws use existing resources

arbulu89:
Implementation for the IP auto generation and usage of already existing resources (vpc and security group by now).

I have done some more changes that we need to discuss:
- I have created an additional subnet creation for the infrastructure (iscsi, monitoring). Before that they were using the first hana subnet, what looked a bit strange
- I have moved hana subnets cration to the hana module
- When an already existing VPC is provided, I assume that it already has an internet gateway attached, otherwise it will fail. Due to some terraform limitations, we cannot create and attach the gateway if it's not there. To have something like this, we would need to add a new parameter asking the user if they want to create the gateway or not, what it's quite annoying. Opinions about this?

So, things that we need to consider:
- Do you think we should create the subnets (infrastructure in general) in the modules, or in the `infrastructure.tf` file?
- By now, I don't give the option to use already existing subnets. I would like to provide it, but the thing is that subnets usage is not so trivial in AWS. We need one subnet for each cluster component, so I'm afraid if users won't do it really good and create issues in the best practices guides. What do you think? Should we provide this option even having risks of users providing only 1 subnet for example?
- Would you prefer to `locals` that creates the IP addresses in the modules? It might be more readible this way, instead of having all this long conditions in a unique `locals`

https://github.com/SUSE/ha-sap-terraform-deployments/issues/402


Simranpal:
Hi @arbulu89  regarding the subnets, since other modules were using the `hana-subnet` I kept it in infrastructure just like other modules. infrastructure file, according to me, should contain the common resources/shared among some modules.
About the `locals` it is same I think, keeping in infrastructure makes sense.

arbulu89:
> Hi @arbulu89 regarding the subnets, since other modules were using the `hana-subnet` I kept it in infrastructure just like other modules. infrastructure file, according to me, should contain the common resources/shared among some modules.
> About the `locals` it is same I think, keeping in infrastructure makes sense.

That's why I have create an additional subnet only for the infrastructure. The change makes that hana specific subnets only used by hana machines, what makes sense in my opinion.

Please, read all the description about the changes in the PR description. Everything is explained there.

aws/infrastructure.tf
```diff
@@ -32,50 +62,55 @@ resource "aws_vpc" "vpc" {
...
     Workspace = terraform.workspace
   }
 }
 
-resource "aws_subnet" "hana-subnet" {
-  count             = var.hana_count
-  vpc_id            = aws_vpc.vpc.id
-  cidr_block        = cidrsubnet(aws_vpc.vpc.cidr_block, 8, count.index)
-  availability_zone = element(data.aws_availability_zones.available.names, count.index)
+resource "aws_subnet" "infra-subnet" {
```
Simranpal, line 74: Will this not have `count` entry? maybe depends on iscsi, and if monitoring is enabled

Simranpal:
> 
> That's why I have create an additional subnet only for the infrastructure. The change makes that hana specific subnets only used by hana machines, what makes sense in my opinion.
> 
> Please, read all the description about the changes in the PR description. Everything is explained there.

Hi @arbulu89 I reviewed the changes now, and it makes sense. About using existing subnets, it seems complex to me.We would need to document about the current subnet usage for each cluster component, and ask user to provide it correctly.

And due to the VPC and gateway dependence, I would assume the user has provided already working vpc & gateway resources.

diegoakechi:
> Implementation for the IP auto generation and usage of already existing resources (vpc and security group by now).
> 
> I have done some more changes that we need to discuss:
> 
> * I have created an additional subnet creation for the infrastructure (iscsi, monitoring). Before that they were using the first hana subnet, what looked a bit strange

I agree.

> * I have moved hana subnets cration to the hana module
> * When an already existing VPC is provided, I assume that it already has an internet gateway attached, otherwise it will fail. Due to some terraform limitations, we cannot create and attach the gateway if it's not there. To have something like this, we would need to add a new parameter asking the user if they want to create the gateway or not, what it's quite annoying. Opinions about this?
I think this is fine and even correct. The internet gateway is something that exposes the infrastructure to the internet, so HANA being a database is something that most probably not many customers will do. So the existing network needs to make sure they have it properly created if needed.


> 
> So, things that we need to consider:
> 
> * Do you think we should create the subnets (infrastructure in general) in the modules, or in the `infrastructure.tf` file?
I would leave it outside of the modules. The network part is even segregated by teams inside some companies, so keeping this isolated might help to do changes related only with network. Maybe even having a module only for network. 

The application modules should simple receive the resulted network and create their components. With this, we can serve one subnet for each group of machines, but also, some skilled user can adapt main.tf to do adaptations for what they need, eventually same subnet for all, maybe a different one for monitoring and other for iscsi, etc.

> * By now, I don't give the option to use already existing subnets. I would like to provide it, but the thing is that subnets usage is not so trivial in AWS. We need one subnet for each cluster component, so I'm afraid if users won't do it really good and create issues in the best practices guides. What do you think? Should we provide this option even having risks of users providing only 1 subnet for example?
I would ask people to request that. It is a change that we do in a second PR anyway.

> * Would you prefer to `locals` that creates the IP addresses in the modules? It might be more readible this way, instead of having all this long conditions in a unique `locals`

I don't have a preference here. One will spread the same logic everywhere... the other keep it centralized but it is ugly....

aws/infrastructure.tf
arbulu89: This subnet is always created by now, so we don't need to have a count. When we make `iscsi` server optional, we will need to take care of that.

aws/main.tf
```diff
@@ -138,15 +157,15 @@ module "monitoring" {
...
+  security_group_id      = local.security_group_id
+  monitoring_srv_ip      = local.monitoring_ip
   private_key_location   = var.private_key_location
   aws_region             = var.aws_region
   availability_zones     = data.aws_availability_zones.available.names
   sles4sap_images        = var.sles4sap
-  subnet_ids             = aws_subnet.hana-subnet.*.id
-  host_ips               = var.host_ips
+  subnet_ids             = aws_subnet.infra-subnet.*.id
+  host_ips               = [local.hana_cluster_vip]
```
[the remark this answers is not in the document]
arbulu89: I want to change this functionality. We don't need to monitor the 2 hana hosts, as the secondary cannot export metrics. That's why I changed.
Maybe I will just set `local.hana_ips` by now, and make the change when the whole `hanadb_exporter` is integrated in the cluster.

aws/README.md
```diff
@@ -146,6 +146,10 @@ The project has been created in order to provide the option to run the deploymen
 
 Besides the `terraform.tfvars` file usage to configure the deployment, a more advanced configuration is available through pillar files customization. Find more information [here](../pillar_examples/README.md).
 
+## Use already existing network resources
+
+The usage of already existing network resources (vpc and security groups) can be done configuring the `terraform.tfvars` file and adjusting some variables. The example of how to use them is available at [terraform.tfvars.example](terraform.tfvars.example).
```
diegoakechi, line 151: @arbulu89 I think we need to document here that the existing VPC must have an internet gateway associated to it in order to be able to register and install the needed packages.

aws/terraform.tfvars.example
```diff
@@ -1,5 +1,19 @@
 # Launch SLES-HAE of SLES4SAP cluster nodes
 
+# Use an already existing vpc
+#vpc_id = "vpc-xxxxxxxxxxxxxxxxx"
```
diegoakechi, line 4: @arbulu89 I think here worth a mention to the Internet Gateway too.

aws/infrastructure.tf
arbulu89: Changed to `local.hana_ips` by now

aws/README.md
arbulu89: Comment added

aws/terraform.tfvars.example
arbulu89: Comment added

PR #444, opened 2020-04-21
Add qa_mode support for nw and drbd for all providers

Simranpal:
Adding/updating the `qa_mode` support for all providers in DRBD and Netweaver modules.
I have updated the variable description for the modules variables.
For aws, it not have drbd module implemented yet, I added qa_mode for netweaver.

### DIFF
--- a/aws/main.tf
+++ b/aws/main.tf
@@ -71,6 +71,7 @@ module "netweaver_node" {
   reg_additional_modules     = var.reg_additional_modules
   ha_sap_deployment_repo     = var.ha_sap_deployment_repo
   devel_mode                 = var.devel_mode
+  qa_mode                    = var.qa_mode
   provisioner                = var.provisioner
   background                 = var.background
   monitoring_enabled         = var.monitoring_enabled
--- a/aws/modules/hana_node/variables.tf
+++ b/aws/modules/hana_node/variables.tf
@@ -211,7 +211,7 @@ variable "hwcct" {
 }
 
 variable "qa_mode" {
-  description = "Whether or not to install the HA/SAP packages from the `ha_sap_deployment_repo`"
+  description = "Enable test/qa mode (disable extra packages usage not coming in the image)"
   type        = bool
   default     = false
 }
--- a/aws/modules/iscsi_server/variables.tf
+++ b/aws/modules/iscsi_server/variables.tf
@@ -107,7 +107,7 @@ variable "background" {
 }
 
 variable "qa_mode" {
-  description = "Whether or not to install the HA/SAP packages from the `ha_sap_deployment_repo`"
+  description = "Enable test/qa mode (disable extra packages usage not coming in the image)"
   type        = bool
   default     = false
 }
--- a/aws/modules/netweaver_node/variables.tf
+++ b/aws/modules/netweaver_node/variables.tf
@@ -228,7 +228,7 @@ variable "devel_mode" {
 }
 
 variable "qa_mode" {
-  description = "Whether or not to install the HA/SAP packages from the `ha_sap_deployment_repo`"
+  description = "Enable test/qa mode (disable extra packages usage not coming in the image)"
   type        = bool
   default     = false
 }
--- a/azure/main.tf
+++ b/azure/main.tf
@@ -51,6 +51,7 @@ module "drbd_node" {
   reg_additional_modules = var.reg_additional_modules
   ha_sap_deployment_repo = var.ha_sap_deployment_repo
   devel_mode             = var.devel_mode
+  qa_mode                = var.qa_mode
   provisioner            = var.provisioner
   background             = var.background
   monitoring_enabled     = var.monitoring_enabled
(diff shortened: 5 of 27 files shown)


arbulu89:
@Simranpal Thank you for the changes. Now the question is, @ldevulder @juadk , what do you expect from `qa_mode` in drbd and netweaver? Is this enough?
The unique doubt I have is that we install some packages in https://github.com/SUSE/ha-sap-terraform-deployments/blob/develop/salt/drbd_node/drbd_packages.sls.

Could you guys have a check and help @Simranpal closing this changes?

ldevulder:
@Simranpal I will try to test this this week. Thx!

diegoakechi:
@ldevulder @juadk I would appreciate it if you could check it this week since we have included this in our Milestone for release of 5.0.0 this week. We need this release to include some changes on the SLE-15-SP2 release.

juadk:
The deployment is failing for drbd node:
```
----------
          ID: /tmp/cluster.config
    Function: file.managed
      Result: False
     Comment: Unable to manage file: Jinja variable 'dict object' has no attribute 'gcp_instance_name'
     Started: 12:03:36.364736
    Duration: 144.842 ms
     Changes:
----------
          ID: configure-the-cluster
    Function: crm.cluster_configured
        Name: update
      Result: False
     Comment: Error configuring the cluster with method update and file /tmp/cluster.config
     Started: 12:03:36.510105
    Duration: 471.33 ms
     Changes:
```
Looks like `gcp_instance_name` used in `drbd_cluster.j2` is not defined somewhere.

ldevulder:
> The unique doubt I have is that we install some packages in https://github.com/SUSE/ha-sap-terraform-deployments/blob/develop/salt/drbd_node/drbd_packages.sls.

Installation of the packages is not so good as during the Build Validation we don't have access to the repositories, so installation can't be done and deployment will fail.

But we don't have any issue with the current version, so the package should already be part of the PC image. For `parted` yes it's clearly in the default installation, and as DRBD is part of HA I think it's also in the SLES4SAP images.

So to conclude: is fine for now, but a needed package that it's not in the image should be added, if not possible then we will have to re-discuss the Build Validation process for PC image (done after the release of the image? To have SCC populated with the new packages).

ldevulder:
> @ldevulder @juadk I would appreciate it if you could check it this week since we have included this in our Milestone for release of 5.0.0 this week. We need this release to include some changes on the SLE-15-SP2 release.

We will try, but it should not be too much time consuming: again as we already discussed we don't have enough time to do the QA on this tool... We have to use this tool to do the QA.

Simranpal:
Hi @arbulu89 as Julien pointed,  we miss adding grains entry for `gcp_instance_name`. 

 https://github.com/SUSE/ha-sap-terraform-deployments/pull/383

arbulu89:
> @arbulu89 as Julien pointed, we miss adding grains entry for `gcp_instance_name`.
> 
> #383

This is added by the `habootstrap-formula` as it can be obtained using the internal gcp api. So, we don't need to pass it as variable.
https://github.com/SUSE/habootstrap-formula/blob/master/cluster/cloud_detection.sls#L7

arbulu89:
@juadk What version of the `habootstrap-exporter` did you use in your test? This shouldn't happen since version `0.3.4`

arbulu89:
> The deployment is failing for drbd node:
> 
> ```
> ----------
>           ID: /tmp/cluster.config
>     Function: file.managed
>       Result: False
>      Comment: Unable to manage file: Jinja variable 'dict object' has no attribute 'gcp_instance_name'
>      Started: 12:03:36.364736
>     Duration: 144.842 ms
>      Changes:
> ----------
>           ID: configure-the-cluster
>     Function: crm.cluster_configured
>         Name: update
>       Result: False
>      Comment: Error configuring the cluster with method update and file /tmp/cluster.config
>      Started: 12:03:36.510105
>     Duration: 471.33 ms
>      Changes:
> ```
> 
> Looks like `gcp_instance_name` used in `drbd_cluster.j2` is not defined somewhere.

@Simranpal Could you do the changes here and only install the packages if `qa_mode` is false. @juadk What about `drbd-formula`? We are always installing the formula regardless `qa_mode` is true or not. To follow the same logic than the others, we should install it always.

I think we should do the same we do in the other modules, and we decide to change strategy do it in all of them.

juadk:
> @juadk What version of the `habootstrap-exporter` did you use in your test? This shouldn't happen since version `0.3.4`

I'm using the version provided in 15SP2, that is:
`i+ | habootstrap-formula | package    | 0.3.0-1.10 | noarch | (System Packages)`

juadk:
> @juadk What about `drbd-formula`? We are always installing the formula regardless `qa_mode` is true or not. To follow the same logic than the others, we should install it always.
> 
> I think we should do the same we do in the other modules, and we decide to change strategy do it in all of them.

@arbulu89 I am ok for `drbd-formula`, this package seems to be already included in the 15-SP2 image.

arbulu89:
> > @juadk What version of the `habootstrap-exporter` did you use in your test? This shouldn't happen since version `0.3.4`
> 
> I'm using the version provided in 15SP2, that is:
> `i+ | habootstrap-formula | package | 0.3.0-1.10 | noarch | (System Packages)`

This is why it fails. `develop` branch of this project doesn't necessarily match with SLE releases. That's why is develop. 
You should get the packages from: https://build.opensuse.org/project/show/network:ha-clustering:sap-deployments:devel

arbulu89:
Let's postpone this PR review until having some ideas about: https://github.com/SUSE/ha-sap-terraform-deployments/issues/463

MalloZup:
@Simranpal @arbulu89 I believe we can close this because we want to get rid of QA_MODE.

Feel free to reopen in case.

PR #449, opened 2020-04-24
Feature/gcp use existing resources

arbulu89:
Change for GCP:
- Implement the usage of already existing vpc, subnet and firewall rules
- Ip addresses autogeneration
- Add nw and drbd monitoring values properly
- Align iscsi_srv_ip parameter name
- Make `hana_ip` for netweaver pillar file usage as other providers


gcp/terraform.tfvars.example
```diff
@@ -1,13 +1,34 @@
...
 
 # Credentials file for GCP
 gcp_credentials_file = "my-project.json"
 
-# Internal IPv4 range
-ip_cidr_range = "10.0.0.0/24"
+# Region where to deploy the configuration
+region = "europe-west1"
+
+# Use an already existing resource group
```
diegoakechi, line 10: @arbulu89 I think here is meant VPC instead resource group.

PR #451, opened 2020-04-27
Feature/libvirt ip generation

arbulu89:
This PR implements this changes:

- Move infrastructure related resources to `infrastructure.tf`
- Implement IP addresses auto generation
- Implement the usage of already existing network and images

Last two can have a lot of benefits. To list some of them:
- We don't need to define all the ip addresses. Defining the ip range for the network is enough
- Using an already existing network can be really beneficial for the CI, as we can book the ip addresses to assure that nobody else will use them. The would be always there.
- Using already existing images is a major upgrade, as we don't need to wait until the image is uploaded every execution to the storage pool. 

About that last one, I have created a mini terraform project to upload the latest images to our libvirt storage pool, and I plan to create a jenkins job to refresh the image daily so we have the latest images always available, and everything could leverage them.

PD: Many changes are just terraform formating


libvirt/modules/monitoring/variables.tf
```diff
@@ -89,9 +89,16 @@ variable "monitoring_srv_ip" {
 
 // Provider-specific variables
 
-variable "base_image_id" {
```
MalloZup, line 92: this is the technical name from libvirt. I am ok to rename it in more human friendly but is libvirt originally terminology which is somehow abstract :grin:

MalloZup:
@arbulu89  once merged can you send an info to the chat ? we might also need to update the confluence page to the new format so we can post the link. great thing :+1:

libvirt/terraform.tfvars.example
```diff
@@ -1,5 +1,24 @@
...
 qemu_uri = "qemu:///system"
 
+# Use already existing network
+#network_name = "my-network"
+# Due to some internal limitations, the iprange of the existing network must be defined
+# The iprange must be defined for already existing network and to create a new one
+iprange = "192.168.XXX.Y/24"
+
+# Base image configuration. This images will be used for all deployed machines unless the specific image is defined
+# The source image has preference over the `base_image _name` parameter
```
Simranpal, line 11: variable name should be `base_image_name `

libvirt/variables.tf
```diff
@@ -12,10 +12,15 @@ variable "storage_pool" {
...
 
+variable "network_name" {
+  description = "Already existing virtual network name. If it's not provided a new one will be created"
+  type        = string
+  default     = ""
+}
+
 variable "iprange" {
-  description = "IP range of the isolated network"
+  description = "IP range of the isolated network (it must be provided even when the network_name is given, due terraform-libvirt-provider limitations we cannot get the current network data)"
```
Simranpal, line 22: `due /due to`

arbulu89:
> @arbulu89 once merged can you send an info to the chat ? we might also need to update the confluence page to the new format so we can post the link. great thing +1

I already sent a mail to our internal mailing list

libvirt/modules/monitoring/variables.tf
arbulu89: In this case we are using the `name` not the `id`. Are both the same? 
In terraform this is where we define the volume:
```
base_volume_name = var.image_name
```

And this is the description: 
```
variable "image_name" {
  description = "Already existing volume name with a sles4sap image that will be used. It's only used if source_image is not provided"
  type        = string
  type        = string
  default     = ""
}
```

Do you still think we should change it?

arbulu89:
> Thanks @arbulu89 looks good.
> Just the variable names *` _image_name` and `_source_name` could be name better, but thats fine for me

What names do you propose? `*_source_url` and `*_image_name` maybe.
I don't want to be changing this all the time, so let's put now the correct names

libvirt/modules/iscsi_server/variables.tf
```diff
@@ -67,6 +62,18 @@ variable "background" {
 
 // Provider-specific variables
 
+variable "source_image" {
+  description = "qcow2 image used to create the image"
```
stefanotorresi, line 66: Typo
```suggestion
  description = "qcow2 image used to create the volume"
```

PR #454, opened 2020-04-28
Update netweaver dashboard

stefanotorresi:
This PR updates the NetWeaver dashboard to leverage the latest version of `sap_host_exporter`.

It also converts some of the existing graphs to rates.

Screenshot from 2020-04-28 16-31-21


stefanotorresi:
> Have you updated something else in the other panels?

Yes, all the counter metrics, previously presented as continuously growing absolute numbers, are now expressed in per-second rates.

PR #455, opened 2020-04-29
Implement drbd module for AWS provider

Simranpal:
I have some doubts with current change:

1. I created the network resources: subnet, routes for drbd, is this fine or should it use the infrastructure resources?
2. What should I set for the defaul drbd `instance_type` type and size?
3. drbd `drbd_cluster_vip` should use the same subnet?
4 drbd volume_size for block_device: setting a default size to 15 & harcoded the values for `root_block_device`

Please also verify these addresses calculations:
 -current `drbd_ips` calculation using the drbd_count in main.tf
 -`drbd_subnet_address_range` calculation in the infrastructure locals
 
To-do: 
Update/verify the grains entry to set in drbd salt provisioner
add logic to use drbd storage or aws efs
update readme and doc


nick-wang:
> * I created the network resources: subnet, routes for drbd, is this fine or should it use the infrastructure resources?
It would be better if DRBD use different subnet/routes than corosync(infrastructure resources), but it is fine to use the same network.

> * What should I set for the defaul drbd `instance_type` type and size?
As the backing device of NFS share, i think the minimal/economic instance_type should be fine to DRBD. 

Above are my two cents for reference.

nick-wang:
A question about the purpose of implementing DRBD. 
I thought AWS has its EFS service for file storage and initially decide to use this. Any reason or benefit to change to DRBD/NFS?

aws/main.tf
```diff
@@ -9,6 +9,8 @@ module "local_execution" {
 # Monitoring: 10.0.0.5
 # Hana ips: 10.0.1.10, 10.0.2.11 (hana machines must be in different subnets)
 # Hana cluster vip: 192.168.1.10 (virtual ip address must be in a different range than the vpc)
+# DRBD ips: 10.0.0.20, 10.0.0.21
+# DRBD cluster vip: 10.74.0.22
```
arbulu89, line 13: Use the same ip range address than the others virtual addresses: `192.168.1.20` would be fine

aws/infrastructure.tf
```diff
@@ -35,6 +35,10 @@ locals {
   # The 2 is hardcoded because we create 2 subnets for NW always
   netweaver_subnet_address_range = length(var.netweaver_subnet_address_range) != 0 ? var.netweaver_subnet_address_range : [
   for index in range(2) : cidrsubnet(local.vpc_address_range, 8, index + var.hana_count + 1)]
+
+  # The 4 is hardcoded considering 4 ip addresses for NW
+  drbd_subnet_address_range = length(var.drbd_subnet_address_range) != 0 ? var.drbd_subnet_address_range : [
+  for index in range(var.drbd_count) : cidrsubnet(local.vpc_address_range, 8, index + var.hana_count + 4 + 1)]
```
arbulu89, line 41: We don't have `drbd_count` variable, right? 
Let's work with the value `2` hardcoded by now.

arbulu89: We must not use the hardcoded value `4` for in `index + var.hana_count + 4 + 1`. We must use `2`, as only 2 subnets are created for drbd

aws/main.tf
```diff
@@ -9,6 +9,8 @@ module "local_execution" {
 # Monitoring: 10.0.0.5
 # Hana ips: 10.0.1.10, 10.0.2.11 (hana machines must be in different subnets)
 # Hana cluster vip: 192.168.1.10 (virtual ip address must be in a different range than the vpc)
+# DRBD ips: 10.0.0.20, 10.0.0.21
```
arbulu89, line 12: Following the subnets usage, the addresses would be `10.0.5.20` and `10.0.6.21`

aws/main.tf
```diff
@@ -21,12 +23,58 @@ locals {
...
+  vpc_id                 = local.vpc_id
+  subnet_address_range   = local.drbd_subnet_address_range
+  key_name               = aws_key_pair.key-pair.key_name
+  security_group_id      = local.security_group_id
+  route_table_id         = aws_route_table.route-table.id
+  aws_credentials        = var.aws_credentials
+  aws_access_key_id      = var.aws_access_key_id
+  aws_secret_access_key  = var.aws_secret_access_key
+  host_ips               = local.drbd_ips
+  drbd_enabled           = var.drbd_enabled
```
arbulu89, line 53: Do we need `drbd_enabled` for anything. I don't think so

aws/main.tf
```diff
@@ -168,7 +216,7 @@ module "monitoring" {
   availability_zones     = data.aws_availability_zones.available.names
   sles4sap_images        = var.sles4sap
   subnet_ids             = aws_subnet.infra-subnet.*.id
-  host_ips               = [local.hana_ips]
+  host_ips               = local.hana_ips
```
arbulu89, line 219: This fix will come in other PR, but It's OK you have it here

aws/modules/drbd_node/main.tf
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  ebs_block_device {
+    volume_type = var.drbd_data_disk_type
+    volume_size = var.drbd_data_disk_size
+    device_name = "/dev/xvdd"
+  }
+
+  volume_tags = {
+    Name = "${terraform.workspace}-${var.name}${var.drbd_count > 1 ? "0${count.index + 1}" : ""}"
```
arbulu89, line 63: Remove the subfix creation logic like we do here: https://github.com/SUSE/ha-sap-terraform-deployments/pull/457
Basically:
```
Name = "${terraform.workspace}-${var.name}0${count.index + 1}"
```
This applies for all of the places where we do this logic

aws/modules/drbd_node/main.tf
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  tags = {
+    Name      = "${terraform.workspace} - ${var.name}${var.drbd_count > 1 ? "0${count.index + 1}" : ""}"
+    Workspace = terraform.workspace
+    Cluster   = "${terraform.workspace}-${var.name}${var.drbd_count > 1 ? "0${count.index + 1}" : ""}"
```
arbulu89, line 69: The same here. Change `Cluster` by `${terraform.workspace}-cluster`. Check example in:
https://github.com/SUSE/ha-sap-terraform-deployments/pull/457/commits/36e65caecf297d48cc2a611ea9bfa684d8a56a07

aws/modules/drbd_node/salt_provisioner.tf
```diff
@@ -0,0 +1,62 @@
...
+    source      = var.aws_access_key_id == "" || var.aws_secret_access_key == "" ? var.aws_credentials : "/dev/null"
+    destination = "/tmp/credentials"
+  }
+
+  provisioner "file" {
+    content     = <<EOF
+provider: aws
+region: ${var.aws_region}
+role: drbd_node
+name_prefix: vm${var.name}
```
arbulu89, line 25: We don't need to put `vm` here. Follow logic like in:
https://github.com/SUSE/ha-sap-terraform-deployments/pull/457

aws/modules/drbd_node/salt_provisioner.tf
```diff
@@ -0,0 +1,62 @@
...
+hostname: vm${var.name}${var.drbd_count > 1 ? "0${count.index + 1}" : ""}
+network_domain: ${var.network_domain}
+host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]
+host_ip: ${element(var.host_ips, count.index)}
+drbd_disk_device: /dev/sdc
```
arbulu89, line 30: Make sure that the disk names are correct please.
The `sdb_disk_device` must be synchronized with the hana and netweaver disks usage. Is not that simple, as each of them mount the disk in a different name.
Basically, iscsi server providers 3 disk. 1st for hana, 2nd for netweaver and 3rd for drbd. So we need to make sure that this happens. I can help you to found out which are the disks

aws/modules/drbd_node/variables.tf
```diff
@@ -0,0 +1,202 @@
+variable "name" {
+  description = "hostname, without the domain part"
+  type        = string
+  default     = "drbd"
+}
+
+variable "drbd_count" {
+  description = "Number of drbd machines to create the cluster"
+  type        = string
```
arbulu89, line 9: Change this to `number` and `2`. (Yes, I know, we might have string in other places, but let's do it fine this time)

aws/modules/drbd_node/variables.tf
```diff
@@ -0,0 +1,202 @@
...
+}
+
+variable "aws_secret_access_key" {
+  type = string
+}
+
+variable "host_ips" {
+  description = "ip addresses to set to the nodes"
+  type        = list(string)
+  default     = []
```
arbulu89, line 92: This shouldn't have a empty default value

aws/modules/drbd_node/variables.tf
```diff
@@ -0,0 +1,202 @@
...
+}
+
+variable "drbd_enabled" {
```
arbulu89, line 95: Remove this `drbd_enabled` variable, we don't need it

aws/modules/drbd_node/variables.tf
```diff
@@ -0,0 +1,202 @@
...
+  description = "Enable the drbd cluster for nfs"
+  type        = bool
+  default     = false
+}
+
+variable "drbd_cluster_vip" {
+  description = "IP address used to configure the drbd cluster floating IP. It must be in other subnet than the machines!"
+  type        = string
+  default     = ""
```
arbulu89, line 104: Shouldn't have default emtpy value

aws/variables.tf
```diff
@@ -259,6 +259,56 @@ variable "scenario_type" {
   default     = "performance-optimized"
 }
 
+# DRBD related variables
+
+variable "drbd_count" {
```
arbulu89, line 264: Remove `drbd_count` by now. We are going to have this hardcoded to 2. We will see if we need to provide the count in the future

arbulu89:
@Simranpal More things.
- We need to update the nfs share used by netweaver, so the logic would be: If we setup drbd, we use the nfs share created by drbd like in azure/gcp. otherwise use the efs share. So, the efs share creation must be optional. Maybe we have to move it to `infrastructure.tf` (most probably)
- We need to document that usage
- We need to update the pillar files (automatic folder) to make sure that drbd works
- We need to change https://github.com/SUSE/ha-sap-terraform-deployments/blob/develop/salt/drbd_node/files/templates/drbd_cluster.j2 to use aws related resources

We can discuss this points online

Simranpal:
> A question about the purpose of implementing DRBD.
> I thought AWS has its EFS service for file storage and initially decide to use this. Any reason or benefit to change to DRBD/NFS?

This was requested by QA team, to have the option for using drbd inplace of efs

aws/modules/drbd_node/variables.tf
Simranpal: I see its empty for other Virtual addresses for Hana and NW, should we add default to all values?

aws/infrastructure.tf
Simranpal: Updated to hardcoded value `2`

aws/modules/drbd_node/salt_provisioner.tf
Simranpal: It is used like this `vm${var.name}`  in hana and drbd  name_prefix https://github.com/SUSE/ha-sap-terraform-deployments/pull/457

Simranpal:
@arbulu89 I wanted to get opinion on the logic for nfs share used by netweaver

(`aws_efs_mount_target` needs the `file_system_id` from `aws_efs_mount_target` so these two resources are better together.)

1. moving the EFS resouces to infrastructure
They need the `subnet_id` from nw subnet created in the nw module, which could be passed via nw module outputs. Then we can set and pass the `netweaver_nfs_share` variable with efs/nfs share value

2. keep EFS resouces in nw module
here need to create EFS resources based on drbd_enabled variable(maybe add a new variable) and pass drbd nfs share variable, then choose in the salt provisioner between efs/nfs share

I followed the 1st method first but the dependencies are more in that case, so I am planning to use the second approach now.

aws/modules/drbd_node/variables.tf
arbulu89: In this case it's a bit irrelevant, as we always pass the value from the `main.tf`. The thing is that we cannot accept an empty value if we only think about the module individually.
The deployment of drbd always requires this value. The main variables.tf file has this entry with empty default value because the ip address can be auto generated

arbulu89:
> @arbulu89 I wanted to get opinion on the logic for nfs share used by netweaver
> 
> (`aws_efs_mount_target` needs the `file_system_id` from `aws_efs_mount_target` so these two resources are better together.)
> 
>     1. moving the EFS resouces to infrastructure
>        They need the `subnet_id` from nw subnet created in the nw module, which could be passed via nw module outputs. Then we can set and pass the `netweaver_nfs_share` variable with efs/nfs share value
> 
>     2. keep EFS resouces in nw module
>        here need to create EFS resources based on drbd_enabled variable(maybe add a new variable) and pass drbd nfs share variable, then choose in the salt provisioner between efs/nfs share
> 
> 
> I followed the 1st method first but the dependencies are more in that case, so I am planning to use the second approach now.

@Simranpal , I see the issue. I would go with the next option:
We can just move the creation of the efs creation to `infrastructure.tf` and keep the mount targets in the netweaver module. This is a good idea, if for some reason, some customer wants to use an already existing EFS storage (because they share the storage for different deployments for example, and reduce costs). So in this case, we would only create the mounting points with the subnets. So we can implement the usage of an already existing EFS storage in the future would be super easy.

What do you think?

aws/modules/drbd_node/salt_provisioner.tf
```diff
@@ -0,0 +1,65 @@
...
+
+  provisioner "file" {
+    content     = <<EOF
+provider: aws
+region: ${var.aws_region}
+role: drbd_node
+name_prefix: vm${var.name}
+aws_cluster_profile: Cluster
+aws_instance_tag: Cluster
+hostname: vm${var.name}${var.drbd_count > 1 ? "0${count.index + 1}" : ""}
```
arbulu89, line 28: Change the logic here be: `hostname: ${var.name}0${count.index + 1}

aws/modules/drbd_node/salt_provisioner.tf
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  provisioner "file" {
+    content     = <<EOF
+provider: aws
+region: ${var.aws_region}
+role: drbd_node
+name_prefix: vm${var.name}
+aws_cluster_profile: Cluster
+aws_instance_tag: Cluster
```
arbulu89, line 27: `aws_instance_tag` should be `${terraform.workspace}-cluster`

salt/drbd_node/files/templates/drbd_cluster.j2
```diff
@@ -65,6 +65,29 @@ primitive vip_{{ res.name }}_nfs ocf:heartbeat:gcp-vpc-move-route \
 
 {% endif %}
 
+{%- if cloud_provider == "amazon-web-services" %}
```
arbulu89, line 68: I think we could combine this `if` with the last one changing to `elif`

aws/modules/drbd_node/variables.tf
Simranpal: Thanks yes that makes sense I update the default value for the module

aws/modules/drbd_node/salt_provisioner.tf
Simranpal: Hi @arbulu89 I have added the `sbd_disk_index` for drbd, but I am not sure about the `drbd_disk_device` Is this name not correct?

Simranpal:
> @Simranpal The implementation is taking a good shape!
> I think we only miss some small adjustments and efs storage proper implementation.
> Good job. Continue working

I have done the changes, and doing the testing. I faced issue with deployment, about subnets for hana, nw not matching instance addresses. it is not related to this changes, I will update after deployment is successful

aws/modules/drbd_node/salt_provisioner.tf
arbulu89: You would need to use `/dev/xvdd`. This is what you defined at least in the disk definition `aws_instance.drbd`

aws/modules/drbd_node/variables.tf
```diff
@@ -0,0 +1,200 @@
...
+}
+
+variable "aws_secret_access_key" {
+  type = string
+}
+
+variable "host_ips" {
+  description = "ip addresses to set to the nodes"
+  type        = list(string)
+  default     = ["10.0.5.20", "10.0.5.21"]
```
arbulu89, line 92: Remove default values please, better to avoid having them

aws/modules/drbd_node/variables.tf
```diff
@@ -0,0 +1,200 @@
...
+}
+
+variable "drbd_cluster_vip" {
+  description = "IP address used to configure the drbd cluster floating IP. It must be in other subnet than the machines!"
+  type        = string
+  default     = "192.168.1.20"
```
arbulu89, line 98: Remove default value

aws/modules/netweaver_node/main.tf
```diff
@@ -30,20 +30,9 @@ resource "aws_route" "ers-cluster-vip" {
...
-
-  tags = {
-    Name = "${terraform.workspace}-efs"
-  }
-}
-
 resource "aws_efs_mount_target" "netweaver-efs-mount-target" {
-  count           = min(var.netweaver_count, 2)
-  file_system_id  = element(aws_efs_file_system.netweaver-efs.*.id, 0)
+  count           = var.netweaver_count > 0 && var.drbd_enabled == false ? 2 : 0
```
arbulu89, line 34: Instead of using `drbd_enabled` check the value of `netweaver_nfs_share`

aws/main.tf
```diff
@@ -70,7 +118,7 @@ module "netweaver_node" {
   key_name                   = aws_key_pair.key-pair.key_name
   security_group_id          = local.security_group_id
   route_table_id             = aws_route_table.route-table.id
-  efs_performance_mode       = var.netweaver_efs_performance_mode
+  efs_file_system_ids        = aws_efs_file_system.netweaver-efs.*.id
```
arbulu89, line 121: Don't pass a list, we are going to always create a unique `efs` storage. Maybe just pass the id of the efs if this exists, and an empty string otherwise

aws/modules/netweaver_node/main.tf
```diff
@@ -30,20 +30,9 @@ resource "aws_route" "ers-cluster-vip" {
...
+  file_system_id  = element(var.efs_file_system_ids, 0)
```
arbulu89, line 35: Change this when you only pass 1 `efs_file_system_id`

aws/modules/netweaver_node/variables.tf
```diff
@@ -244,6 +248,12 @@ variable "background" {
   default     = false
 }
 
+variable "drbd_enabled" {
```
arbulu89, line 251: Remove this

aws/main.tf
arbulu89: `join("",aws_efs_file_system.netweaver-efs.*.id)`

aws/modules/netweaver_node/main.tf
Simranpal: I will use the `efs_file_system_id` as discussed

aws/modules/drbd_node/salt_provisioner.tf
```diff
@@ -0,0 +1,69 @@
...
+aws_cluster_profile: Cluster
+aws_instance_tag: ${terraform.workspace}-cluster
+aws_credentials_file: /tmp/credentials
+aws_access_key_id: ${var.aws_access_key_id}
+aws_secret_access_key: ${var.aws_secret_access_key}
+hostname: ${var.name}0${count.index + 1}
+network_domain: ${var.network_domain}
+host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]
+host_ip: ${element(var.host_ips, count.index)}
+drbd_disk_device: /dev/xvdd
```
arbulu89, line 35: As an improvement we should check if we can get this device name from terraform. Have a look here: https://www.terraform.io/docs/providers/aws/d/instance.html#ebs_block_device

I think it should be possible

aws/modules/drbd_node/variables.tf
```diff
@@ -0,0 +1,198 @@
...
+
+variable "drbd_cluster_vip" {
+  description = "IP address used to configure the drbd cluster floating IP. It must be in other subnet than the machines!"
+  type        = string
+}
+
+variable "drbd_data_disk_size" {
+  description = "Disk size of the disks used to store drbd content"
+  type        = string
+  default     = "10"
```
arbulu89, line 102: @Simranpal Change the default value to 60. In AWS we download the NW media here to be shared among the machines and don't download in all of them. 60gb should be enough to store the media and the other NW files

Simranpal:
Hi @diegoakechi  @arbulu89, in AWS we are using the shared NFS share to store the NW installation software. Using DRBD, this means that the replicated disk might size around 30GB, is this an issue? I have been trying deployments today and it takes long time and deployment fails with the timeouts.
Should we change the strategy and download the NW installation software to a local disk for all the nodes?

aws/modules/drbd_node/variables.tf
Simranpal: Thanks @arbulu89 i tried to increase the disk sizes to `60` and then `30` the drbd sync takes a very long time, while other nw mount states timeout

salt/drbd_node/files/templates/drbd_cluster.j2
```diff
@@ -63,7 +63,28 @@ primitive vip_{{ res.name }}_nfs ocf:heartbeat:gcp-vpc-move-route \
...
+    stonith-timeout="150s"
+
+primitive rsc_aws_stonith_{{ res.name }}_nfs stonith:external/ec2 \
+    params tag={{ data.instance_tag }} profile={{ data.cluster_profile }} \
+    op start interval=0 timeout=180 \
+    op stop interval=0 timeout=180 \
+    op monitor interval=120 timeout=60 \
+    meta target-role=Started
+
+primitive rsc_aws_vip_{{ res.name }}_nfs ocf:suse:aws-vpc-move-ip \
```
arbulu89, line 80: @Simranpal I have noticed that we need to change the name of this resource to `vip_{{ res.name }}_nfs`. This resource is grouped in `g-nfs_{{ res.name }}`, so it must have this name to use the common names.

aws/modules/netweaver_node/main.tf
Simranpal: Changed the logic to use a new boolean variable efs_enable_mount

aws/terraform.tfvars.example
```diff
@@ -142,10 +142,27 @@ monitoring_srv_ip = "10.0.0.253"
 # true or false (default)
 #hwcct = false
 
+# drbd related variables
+
+# netweaver will use AWS efs for nfs share by default, unless drbd is enabled
+# Enable drbd cluster
+#drbd_enabled = false
+
+#drbd_instancetype = "t2.micro"
```
ldevulder, line 151: `t2.micro` is 1 vCPU and 1 GB of RAM, while `n1-standard-4` is used for GCP (4 vCPUs and 15GB of RAM) and `` for Azure (2 vCPUs and 8GB of RAM).

Could be better to have the same kind of default VM for all providers, no?

salt/drbd_node/files/templates/drbd_cluster.j2
```diff
@@ -63,7 +63,28 @@ primitive vip_{{ res.name }}_nfs ocf:heartbeat:gcp-vpc-move-route \
   op stop interval=0 timeout=180 \
   op monitor interval=60 timeout=60
 
-{% endif %}
```
ldevulder, line 66: Not related to this PR but if you check L53 you can see and `{% -endif }` and a `{% -elif ... }` right after: this `endif` shouldn't be removed?

arbulu89: No, this `endif` is only for the condition where the provider is azure, and the `elif` works with the previous `if`. Maybe it's a bit confusing, but the conditions are right.

aws/terraform.tfvars.example
arbulu89: @ldevulder Thanks for the feedback here, and I agree, our defaults values are set randmonly...
Could you create a ticket to follow up this discussion. We need to research about which are the best VM sizes for each cluster and set them by default. And we need to do this in all providers and all clusters.

salt/drbd_node/files/templates/drbd_cluster.j2
ldevulder: You're right, this is why I like indention in programming languages :-)

aws/terraform.tfvars.example
ldevulder: @arbulu89 done: #466

PR #456, opened 2020-04-29
Improvements/pre v5 improvements

arbulu89:
Involved changes:
- Remove `template` provider. We don't use it, so we can just get rid off it
- Fix a bug in aws monitoring module usage
- Replace `host_ips` by `hana_ips` as it makes more sense


MalloZup:
lgtm

PR #457, opened 2020-04-29
Improvements/remove hostname indexes

arbulu89:
Changes:
- Remove the logic to add a subix or not to the hostname. It doesn't have much sense and any usage. Actually, it makes things worst when you want to create a single hana environment
- Remove terraform workspace from hostname to make it more simple. GCP is an exception, as the hostname and machine name must be the same (unless you use some external dns service)
- Improve AWS cluster tag usage. Otherwise, the `ec2` fence agent could report machines in other clusters as all of them where using the same tag.

### DIFF
--- a/aws/modules/monitoring/salt_provisioner.tf
+++ b/aws/modules/monitoring/salt_provisioner.tf
@@ -17,9 +17,9 @@ resource "null_resource" "monitoring_provisioner" {
 provider: aws
 region: ${var.aws_region}
 role: monitoring
-name_prefix: ${terraform.workspace}-monitoring
+name_prefix: monitoring
+hostname: monitoring
 host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]
-hostname: ${terraform.workspace}-monitoring
 timezone: ${var.timezone}
 reg_code: ${var.reg_code}
 reg_email: ${var.reg_email}
--- a/azure/modules/drbd_node/salt_provisioner.tf
+++ b/azure/modules/drbd_node/salt_provisioner.tf
@@ -17,7 +17,7 @@ resource "null_resource" "drbd_provisioner" {
 provider: azure
 role: drbd_node
 name_prefix: vm${var.name}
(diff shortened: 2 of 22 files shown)


azure/modules/drbd_node/salt_provisioner.tf
```diff
@@ -17,7 +17,7 @@ resource "null_resource" "drbd_provisioner" {
 provider: azure
 role: drbd_node
 name_prefix: vm${var.name}
```
Simranpal, line 19: Should you remove the `vm` prefix here in azure(same for hana_node), like other name-prefixes in aws, libvirt

aws/modules/monitoring/salt_provisioner.tf
```diff
@@ -17,9 +17,9 @@ resource "null_resource" "monitoring_provisioner" {
 provider: aws
 region: ${var.aws_region}
 role: monitoring
-name_prefix: ${terraform.workspace}-monitoring
+name_prefix: monitoring
+hostname: monitoring
 host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]
-hostname: ${terraform.workspace}-monitoring
```
Simranpal, line 22: Should we not keep any prefix for monitoring hostname, incase of multiple deployments to keep them unique

azure/modules/drbd_node/salt_provisioner.tf
arbulu89: No, this was put on purpose by the azure engineers as they use this prefix in the virtual machines. But it only applies to Azure.

aws/modules/monitoring/salt_provisioner.tf
arbulu89: In the case of aws the hostname is just an internal thing. We cannot have duplicates here. What matters is the vm name, which is different and has the workspace.
The limitation to have the same hostname and machine name only exists in gcp.

PR #459, opened 2020-04-29
Update HA cluster dashboards

### LINKED ISSUE #315: Add resource type to the resources table
I miss the resource type in the table with the information of the resources.

image

Could we add a new column showing the resource type?

As far as I checked there is not such an information in our current metrics of the `ha_cluster_exporter`, so it should be implemented.

stefanotorresi:
This PR addresses BC breaking changes that will be introduced by ClusterLabs/ha_cluster_exporter#152.

It also overhauls the "HA Cluster Details" dashboard by introducing some new graphs to track node and resource statuses, as shown in this screenshot:

Screenshot from 2020-04-29 17-41-42

Demo [here](http://10.162.30.4/d/Q5YJpwtZk1/ha-cluster-details)

The DRBD and Alerts panels were removed for now, mainly because they didn't add much value as they were, and another way of handling that information is being explored.

This also fixes #315

### DIFF
--- a/salt/monitoring/provisioning/dashboards/ha-cluster-details.json
+++ b/salt/monitoring/provisioning/dashboards/ha-cluster-details.json
@@ -20,6 +20,12 @@
       "version": "6.3.5"
     },
     {
+      "type": "panel",
+      "id": "graph",
+      "name": "Graph",
+      "version": ""
+    },
+    {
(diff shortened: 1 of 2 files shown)


diegoakechi:
@stefanotorresi Not late to say that it looks gorgeous!

PR #461, opened 2020-04-30
Remove PackageHub extension

arbulu89:
Remove PackageHub extension installation, as we don't need it any more (the gcp cloud packages are available in the cloud extension).
This change is important, as we don't need to use the `devel_mode` flag any longer to make the deployment work.


MalloZup:
thx

PR #464, opened 2020-04-30
Add outputs for openQA tests

ldevulder:
For QA test, we need the ID of the Azure's VMs to be able to stop/start them during our test using the CSP API.

This commit also add `openqa_vms` and `openqa_ips` for all provider to simplify openQA code and share output with QAC team.

Manually tested on master branch.

### DIFF
--- a/aws/outputs.tf
+++ b/aws/outputs.tf
@@ -75,3 +75,13 @@ output "iscsisrv_name" {
 output "iscsisrv_public_name" {
   value = module.iscsi_server.iscsisrv_public_name
 }
+
+# For openQA (QA mode)
+
+output openqa_vms {
+  value = concat(module.hana_node.cluster_nodes_name, module.netweaver_node.netweaver_name)
+}
+
+output openqa_ips {
+  value = concat(module.hana_node.cluster_nodes_public_ip, module.netweaver_node.netweaver_public_ip)
+}
--- a/azure/modules/drbd_node/outputs.tf
+++ b/azure/modules/drbd_node/outputs.tf
@@ -25,3 +25,7 @@ output "drbd_name" {
 output "drbd_public_name" {
   value = data.azurerm_public_ip.drbd.*.fqdn
 }
+
+output "drbd_id" {
+  value = azurerm_virtual_machine.drbd.*.id
+}
--- a/azure/modules/hana_node/outputs.tf
+++ b/azure/modules/hana_node/outputs.tf
@@ -31,3 +31,7 @@ output "cluster_nodes_name" {
 output "cluster_nodes_public_name" {
   value = data.azurerm_public_ip.hana.*.fqdn
 }
+
+output "cluster_nodes_id" {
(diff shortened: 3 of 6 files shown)


arbulu89:
Hi @ldevulder ,
Honestly, I don't really want any reference to `openQA` in the project. As repeated many times, this is not a testing tool project. If you use it for that matter it's fine, but we cannot add specific code for that.

If you need this information, It's fine, and I guess more users will appreciate to have it, but it shouldn't be something that belongs to `qa`.

In fact, I already wanted to improve the outputs. That's why we need to think further how and what we want to output, a general long term strategy, instead of just adding outputs now and them.

All the outputs should have the same shared syntax, and not duplicated things in order to make things easier for openQA.

stefanotorresi:
Yeah, I agree with @arbulu89, we need a more generalized approach to solve the root problem, i.e. outputting more comprehensive data about the created resources.

ldevulder:
> If you need this information, It's fine, and I guess more users will appreciate to have it, but it shouldn't be something that belongs to qa.

Yes I can rename them to `qa_vms` and `qa_ips`.

> All the outputs should have the same shared syntax, and not duplicated things in order to make things easier for openQA.

It partially duplicate yes but this is to also share more code with QA-SLE team on that side.

> As repeated many times, this is not a testing tool project. If you use it for that matter it's fine, but we cannot add specific code for that.

I partially disagree on this: I perfectly understand why this tool is used for now and what the value it have for customers, but check the first commits and what it was supposed to do!

diegoakechi:
@ldevulder reviewing this, apart of the arguments around the project direction, did you consider to get these metadata from the terraform state? It could reduce the level of dependency that openQA has to the terraform output implementation in general. For example, imagine that the output format needs to be changed for some reason, or that I new module is included, it would demand changes on the openQA side or the that developer takes care to include the outputs in such a correct format. 
If you queries the state, this dependency will not exists. You can do it in two different ways:
By resource (terraform specific output)
```
terraform state show module.iscsi_server.data.azurerm_public_ip.iscsisrv
```
or querying the entire state as JSON and parsing this with jq or python:

```
terraform show -json
```

ldevulder:
@diegoakechi the idea of this is to share more code qith QAC team (the team who is working on PC tests for SLES PC images). The idea of this output is to have something generic from an openQA point of view. I already have a code for this who check for drbd, hana, nw entry, but I need to know what to test in the openqa code. This is why I wanted to add this in the TF tool. Having such an output (can be renamed like @arbulu89 suggested) is not an issue for customer I think, it's just an ouput.

diegoakechi:
@ldevulder my point wasn't about the variable name, but in the sense of making the openQA terraform integration more robust. If you don't depend on the output but on the tfstate instead, your solution will work always, and will not break if something is changed on the tf outputs, because it would rely on Terraform metadata instead.

MalloZup:
as discussed we can close this one. THX!

PR #465, opened 2020-05-05
Update monitoring modules

stefanotorresi:
This PR introduces some changes related to the updates in SUSE/saphanabootstrap-formula#82.

It makes monitoring settings consistent across all terraform modules and also the salt formulas. As such, it depends on:
- SUSE/habootstrap-formula#48
- SUSE/sapnwbootstrap-formula#47

It also
- splits up the salt states for Prometheus and Grafana in dedicated sls files
- fixes a configuration issue that would prevent browsing the Prometheus UI from Grafana.
- removes the state that would have Grafana running as `root`
- resets Grafana HTTP listening port to default (`3000`).


aws/main.tf
```diff
@@ -228,10 +227,9 @@ module "monitoring" {
   provisioner            = var.provisioner
   background             = var.background
   monitoring_enabled     = var.monitoring_enabled
-  drbd_enabled           = var.drbd_enabled
-  drbd_ips               = local.drbd_ips
-  netweaver_enabled      = var.netweaver_enabled
-  netweaver_ips          = local.netweaver_virtual_ips
+  hana_targets           = concat(local.hana_ips, [local.hana_cluster_vip])
```
arbulu89, line 230: I took my a while to understand how this works, and in fact it's documented in the variables.tf file...
I hope we don't change this by accident, as it's not so obvious and we don't tend to check the variable description.
Should we add a comment in top of this line explaining what it does?

pillar_examples/automatic/netweaver/cluster.sls
```diff
@@ -37,13 +37,7 @@ cluster:
...
       max_messages: 20
   {% endif %}
-  {%- for node in netweaver.netweaver.nodes if node.host == grains['host'] %}
-  {%- if grains.get('monitoring_enabled', False) and node.sap_instance in ['ascs', 'ers'] %}
-  ha_exporter: true
-  {%- else %}
-  ha_exporter: false
-  {%- endif %}
-  {%- endfor %}
+  monitoring_enabled: {{ grains['monitoring_enabled']|default(False) }}
```
arbulu89, line 40: This means that the HA exporter is enabled in all the nodes, even though the cluster is not there.
Why don't you like the previous code, which only enables the exporter in clustered nodes?

pillar_examples/automatic/netweaver/cluster.sls
```diff
@@ -75,4 +69,3 @@ cluster:
         route_table: {{ grains['route_table'] }}
         vpc_network_name: {{ grains['vpc_network_name'] }}
         {%- endif %}
-        monitoring_enabled: {{ grains['monitoring_enabled']|default(False) }}
```
arbulu89, line 78: Why have you removed this line? Am I missing something? Without this the HA resources to move the exporter won't be created, right?
We use this in the nw template: 
```
{%- set data = pillar.cluster.configure.template.parameters %}
```
I think we miss to change the `sapnwbootstrap-formula` for that:
https://github.com/SUSE/sapnwbootstrap-formula/blob/master/templates/cluster_resources.j2#L14

pillar_examples/automatic/netweaver/cluster.sls
arbulu89: Sorry, I was wrong, this pillar file is not executed in non HA nodes, so the exporter must be installed always

pillar_examples/automatic/netweaver/cluster.sls
stefanotorresi: as discusssed, we now use the topmost `cluster.monitoring_enabled` field to enable monitoring on all the HA templates in the formulas.

aws/main.tf
stefanotorresi: adding a comment!

salt/monitoring/grafana/provisioning/dashboards/dashboards.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: 1
+
+providers:
+- name: SUSE
+  folder: ''
+  type: file
+  allowUiUpdates: true
+  editable: true
```
MalloZup, line 8: i'm ok if we allow editable mode

salt/monitoring/init.sls
```diff
@@ -2,126 +2,18 @@
...
- pkgrepo.managed:
+  pkgrepo.managed:
     - humanname: Server:Monitoring
     - baseurl: https://download.opensuse.org/repositories/server:/monitoring/{{ repository }}/
     - refresh: True
     - gpgautoimport: True
     - retry:
         attempts: 3
         interval: 15
+    - require_in:
```
MalloZup, line 13: I find `require_in` statement of salt kind of confusing.
In sake of simplicity, from my pov we should not use it. I just searched and this is the first time it is present.

I had to google the difference of require_in and vs require. Imho we could live without the `require_in`

stefanotorresi: It's just the inversed semantic of `require`. What's wrong with it? It's kinda like dependency injection.

PR #468, opened 2020-05-08
Feature/improve iscsi module

arbulu89:
Implement the option to use iscsi optionally, and not always as we were doing in the cloud providers.

Here the most important things:
- Create multiple iscsi machines. This is only implemented in the module level. In order to use more, we should change the `main.tf` file.
- Add `sbd_storage_type` and `*_cluster_sbd_enabled` variables. They are used to decide if sbd is used or not, and which is the type. Cloud providers only provide iscsi option
- Make iscsi disk size customizable (10GB by default)
- Now, we can decide the LUN count provided by Iscsi, and this is not hardcoded anymore
- Change how `sbd_disk_device` is used by salt. Now, if iscsi is used, this grain will be updated in the iscsi salt part. So, the logic is removed from the pillar files, what makes everything cleaner.
- Change the documentation to reference sbd more than iscsi, as sbd is what we want to deploy, more than a iscsi server perse.

**A lot of files are changed, but most of them are just really small changes.**

Pending things:
- Iscsi output is different in each cloud provider, Some offer the data as list, other just individual data, we need to unify this in the future


aws/modules/iscsi_server/variables.tf
```diff
@@ -1,15 +1,16 @@
 variable "aws_region" {
-  type = string
+  type        = string
+  description = "AWS region where the deployment machines will be created"
 }
 
 variable "availability_zones" {
   type        = list(string)
-  description = "Used availability zones"
+  description = "Availability zone nameas where the machines are deployed"
```
yandroskaos, line 8: Typo: names instead of nameas

aws/modules/iscsi_server/variables.tf
```diff
@@ -28,16 +29,15 @@ variable "iscsi_srv_images" {
   }
 }
 
-variable "min_instancetype" {
-  description = "The minimum cost/capacity instance type, different per region."
-  type        = string
-  default     = "t2.micro"
+variable "iscsi_count" {
```
yandroskaos, line 32: Nice iscsi module is now parametrized to generate a number of them

libvirt/variables.tf
```diff
@@ -185,14 +185,27 @@ variable "hana_cluster_vip" {
   default     = ""
 }
 
+variable "hana_cluster_sbd_enabled" {
+  description = "Enable sbd usage in the hana HA cluster"
+  type        = bool
+  default     = true
```
yandroskaos, line 191: libvirt and azure has sbd enabled as default, gcp and aws don't... why?

arbulu89: Because, `gcp` and `aws` have their own fence agents installed by default, so sbd is a good to have thing. In libvirt and azure we don't deploy other fence agents, so sbd is mandatory to have the cluster working

salt/cluster_node/iscsi_initiator.sls
```diff
@@ -47,3 +46,14 @@ iscsid:
   service.running:
     - watch:
       - cmd: iscsi_discovery
+
+# This state sets the sbd_disk grain value. As we cannot run the lssci command directly to get the output (the output is change in the latest command)
+# this workaround that scenario to render the ouptut during execution time
```
Simranpal, line 51: `ouptut // output`

libvirt/terraform.tfvars.example
```diff
@@ -34,9 +34,20 @@ hdbserver_sar = "IMDB_SERVER.SAR"
...
+# In order to enable SBD, an ISCSI server is needed as right now is the unique option
+# All the clusters will use the same mechanism
+# In order to enable the iscsi machine creation sbd_enabled must be set to true for any of the clusters
+
+#sbd_storage_type = "iscsi"
 #iscsi_srv_ip = "192.168.XXX.Y+6"
+# Number of LUN (logical units) to serve with the iscsi server. Each LUN can be used as a unique sbd disk
+#iscsi_lun_count = 3
+# Disk size in Bytes used to create the LUNs and partitions to be served by the ISCSI service
+#sbd_disk_size = 10737418240
```
Simranpal, line 50: can this not be in GB :)

libvirt/modules/netweaver_node/salt_provisioner.tf
```diff
@@ -45,8 +45,11 @@ ers_instance_number: ${var.ers_instance_number}
 pas_instance_number: ${var.pas_instance_number}
 aas_instance_number: ${var.aas_instance_number}
 ha_sap_deployment_repo: ${var.ha_sap_deployment_repo}
-shared_storage_type: shared-disk
-sbd_disk_device: /dev/vdb1
+sbd_enabled: ${var.sbd_enabled}
+sbd_storage_type: ${var.sbd_storage_type}
+sbd_disk_device: "${var.sbd_storage_type == "shared-disk" ? "/dev/vdb1" : ""}"
```
Simranpal, line 50: Will this not result in extra double quotes? same for other modules and providers grains entries

gcp/modules/iscsi_server/salt_provisioner.tf
```diff
@@ -16,33 +16,22 @@ resource "null_resource" "iscsi_provisioner" {
...
-  2:
-    start: 33%
-    end: 67%
-  3:
-    start: 67%
-    end: 100%
+${yamlencode(
+  {partitions: {for index in range(var.lun_count) :
+    tonumber(index+1) => {
+      start: format("%.0f%%", index*100/var.lun_count),
```
Simranpal, line 30: the `.0f` for floating type?

gcp/modules/iscsi_server/main.tf
```diff
@@ -1,14 +1,16 @@
...
+  zone  = element(var.compute_zones, 0)
 }
 
 resource "google_compute_instance" "iscsisrv" {
-  name         = "${terraform.workspace}-iscsisrv"
+  count        = var.iscsi_count
+  name         = "${terraform.workspace}-iscsisrv-${count.index + 1}"
   description  = "iSCSI server"
-  machine_type = var.machine_type_iscsi_server
+  machine_type = var.machine_type
```
Simranpal, line 13: Should we set the machine type default for other providers as well

aws/terraform.tfvars.example
```diff
@@ -69,16 +66,25 @@ hana_extract_dir = "/sapmedia/HDBSERVER"
...
-# iSCSI server address. It should be in same iprange as hana_ips
-iscsi_srv_ip = "10.0.0.254"
+# SBD related variables
+# In order to enable SBD, an ISCSI server is needed as right now is the unique option
+# All the clusters will use the same mechanism
+# In order to enable the iscsi machine creation sbd_enabled must be set to true for any of the clusters
 
-# Device used by the iSCSI server to provide LUNs
-iscsidev = "/dev/xvdd"
+# iSCSI server address. It should be in same iprange as hana_ips
```
Simranpal, line 82: should we also add comment, that all IPs will be auto generated if not provided

aws/modules/iscsi_server/main.tf
```diff
@@ -1,14 +1,19 @@
 # iscsi server resources
 
+locals {
```
Simranpal, line 3: I guess we can make use of locals here and elsewhere in modules for keeping other hard-coded values as well

libvirt/main.tf
```diff
@@ -31,11 +31,14 @@ locals {
   netweaver_ip_start    = 30
   netweaver_ips         = length(var.netweaver_ips) != 0 ? var.netweaver_ips : [for ip_index in range(local.netweaver_ip_start, local.netweaver_ip_start + 4) : cidrhost(local.iprange, ip_index)]
   netweaver_virtual_ips = length(var.netweaver_virtual_ips) != 0 ? var.netweaver_virtual_ips : [for ip_index in range(local.netweaver_ip_start, local.netweaver_ip_start + 4) : cidrhost(local.iprange, ip_index + 4)]
+
+  # Check if iscsi server has to be created
+  iscsi_enabled = var.sbd_storage_type == "iscsi" && (var.hana_cluster_sbd_enabled == true || (var.drbd_enabled && var.drbd_cluster_sbd_enabled == true) || (var.netweaver_enabled && var.netweaver_cluster_sbd_enabled == true)) ? true : false
```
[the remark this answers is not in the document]
arbulu89: It's an extra check that never comes bad. If you disable hana's sbd, disable netweaver, but for some reason you missed to set netweaver_cluster_sbd_enabled to false, the iscsi machine will be created, but not used. This is not what we want.
More in Azure for example, where there is not any other fencing mechanism than sbd, so the values are set to true. In this case, we need to check whether they are enabled or not.

aws/main.tf
```diff
@@ -25,22 +25,25 @@ locals {
...
-  iscsi_instancetype     = var.iscsi_instancetype
-  min_instancetype       = var.min_instancetype
+  instance_type          = var.iscsi_instancetype
   key_name               = aws_key_pair.key-pair.key_name
   security_group_id      = local.security_group_id
   private_key_location   = var.private_key_location
-  iscsi_srv_ip           = local.iscsi_ip
-  iscsidev               = var.iscsidev
-  iscsi_disks            = var.iscsi_disks
+  host_ips               = [local.iscsi_ip]
```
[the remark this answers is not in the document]
arbulu89: No, we cannot. Now iscsi module is designed to create multiple machines if needed, and all of them need a different IP address.
Take this as example: https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker

We don't deploy such 3 iscsi machines setup, but now it's doable.

salt/cluster_node/sbd.sls
```diff
@@ -0,0 +1,7 @@
+# This state is executed at the end of iscsi_initiator.sls state
+{% set sbd_disk_device = salt['cmd.run']('lsscsi | grep "LIO-ORG" | awk "{ if (NR=='~grains['sbd_disk_index']~') print \$NF }"', python_shell=true) %}
```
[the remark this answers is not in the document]
arbulu89: It runs the command, get the value and set in the grains, end of story hehe
What don't you understand?

libvirt/terraform.tfvars.example
arbulu89: Nope XD

gcp/modules/iscsi_server/salt_provisioner.tf
arbulu89: This syntax is used to round the value, to avoid decimals usage:
https://www.terraform.io/docs/configuration/functions/format.html

gcp/modules/iscsi_server/main.tf
arbulu89: I don't understand this comment. This doesn't want to have any default value. It's just removing the `iscsi_server` part, as it's redundant. We are here in the iscsi module, obviously the machine type is for that server

aws/modules/iscsi_server/main.tf
arbulu89: Yes, I'm trying to do so for many variables that are reused. Feel free to use this method in the future to keep the code clean and easier to maintain

aws/terraform.tfvars.example
arbulu89: Yes, this applies to all the ip addresse. Let's let this for a future improvement. Could you create a ticket in our backlog?

libvirt/modules/netweaver_node/salt_provisioner.tf
arbulu89: No, this sets just once the quotes. It's true that we could remove the first and last quotes. But the syntax accepts string with and without quotes, so no issues here.
This is the result:
```
sbd_storage_type: shared-disk
sbd_disk_device: "/dev/vdc"
sbd_disk_index: 1
```

aws/main.tf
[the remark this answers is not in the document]
Simranpal: Thanks yes I noticed that

salt/cluster_node/sbd.sls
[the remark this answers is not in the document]
Simranpal: I ran the command and get the output. But I did not get whats the `LIO-ORG` and the `awk` output.

gcp/modules/iscsi_server/main.tf
Simranpal: What I meant is that the variable has default `"custom-1-2048"` for `machine_type` if we are setting a default value for all machines in other providers, like here in GCP

aws/terraform.tfvars.example
Simranpal: Okay I created the ticket with 'Update doc on automatic IP generation'

PR #469, opened 2020-05-08
Add first workflow for github actions

MalloZup:
THIS pr introduce github workflow action for run PR. 

The test run for libvirt every PR. Is not optimized, we could add some filter like if it  modify libvirt etc, but I would propose to merge this and see how the worker behave. 

In future we can add saltstack execution and some condition filters, like don't run tests if PR modify documentation files etc etc


MalloZup:
closing in favor of new PR

PR #470, opened 2020-05-11
GitHub actions test for prs

MalloZup:
# description:

As discussed this PR introduce the needed file for testing PR with terraform.

It introduce 2 secrets as env. vars as discussed. I think it is a good compromise.

I need to check now why it doesn't run/triggered. I didn't change much


.github/workflows/tf-validation.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: terraform apply netweaver and hana
+      env: #
+        QEMU_URI: ${{ secrets.qemu_uri }}
+        SOURCE_IMAGE: ${{ secrets.source_image }}
+     
+      run: |
+        cp .github/hana-netweaver-tf-only.tfvars libvirt/terraform.tfvars
+        cd libvirt
+        export TF_VAR_qemu_uri="$QEMU_URI"
+        export TF_VAR_source_image="$SOURCE_IMAGE"
```
stefanotorresi, line 24: You could avoid defining the vars and then exporting them again with another name: you can just name the env vars in the top section with the TF_VAR prefix.
```suggestion
      env: #
        TF_VAR_qemu_uri: ${{ secrets.qemu_uri }}
        TF_VAR_source_image: ${{ secrets.source_image }}
     
      run: |
        cp .github/hana-netweaver-tf-only.tfvars libvirt/terraform.tfvars
        cd libvirt
```

.github/workflows/tf-validation.yml
```diff
@@ -0,0 +1,43 @@
...
+      env: #
+        QEMU_URI: ${{ secrets.qemu_uri }}
+        SOURCE_IMAGE: ${{ secrets.source_image }}
+    
+      if: ${{ always() }}
+      run: |
+        cp .github/hana-netweaver-tf-only.tfvars libvirt/terraform.tfvars
+        cd libvirt
+        export TF_VAR_qemu_uri="$QEMU_URI"
+        export TF_VAR_source_image="$SOURCE_IMAGE"
```
stefanotorresi, line 40: same as above ;)

.github/hana-netweaver-tf-only.tfvars
```diff
@@ -0,0 +1,47 @@
...
+netweaver_inst_media     = "10.162.32.134:/sapdata/sap_inst_media"
+netweaver_swpm_folder     =  "SWPM_10_SP26_6"
+
+# Install NetWeaver
+netweaver_sapexe_folder   =  "kernel_nw75_sar"
+netweaver_additional_dvds = ["51050829_3", "51053787"]
+
+
+# NFS share to store the Netweaver shared files
+netweaver_nfs_share    = "192.168.108.201:/HA1"
```
stefanotorresi, line 37: I think we can omit this variable entirely: it will be interpolated with the default network address and path (by the way the current value wouldn't work: the network is declared above in the 192.168.16 range, but here it uses 192.168.108).

.github/workflows/tf-validation.yml
MalloZup: done

.github/hana-netweaver-tf-only.tfvars
MalloZup: @arbulu89  can we remov this?

arbulu89: Yes, with the latest changes, if drbd is enabled, this variable is automatically generated

.github/hana-netweaver-tf-only.tfvars
```diff
@@ -0,0 +1,47 @@
+# the following 2 vars are aquired via gh env variables
+# qemu_uri            = 
+# source_image        = 
+
+hana_inst_media     = "10.162.32.134:/sapdata/sap_inst_media/51053787"
+iprange             = "192.168.25.0/24"
```
arbulu89, line 6: I would create a fixed virtual network in our agent, and set there the name with `network_name   = "gh_agent"`.
This way, we don't have the risk that someones uses that IP address accidentally

.github/hana-netweaver-tf-only.tfvars
```diff
@@ -0,0 +1,47 @@
...
+
+storage_pool = "terraform"
+
+# Enable pre deployment to automatically copy the pillar files and create cluster ssh keys
+pre_deployment = true
+
+# For iscsi, it will deploy a new machine hosting an iscsi service
+shared_storage_type = "shared-disk"
```
arbulu89, line 14: I would use ISCSI, it tests more code

.github/hana-netweaver-tf-only.tfvars
```diff
@@ -0,0 +1,47 @@
...
+# NFS share to store the Netweaver shared files
+netweaver_nfs_share    = "192.168.108.201:/HA1"
+
+# DRBD variables
+
+# Enable the DRBD cluster for nfs
+drbd_enabled = true
+
+# IP of DRBD cluster
+drbd_shared_storage_type = "shared-disk"
```
arbulu89, line 45: I would use iscsi even here

.github/workflows/tf-validation.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  terraform-deploy-netweaver-hana:
+    runs-on: self-hosted 
+
+    steps:
+    - uses: actions/checkout@v2
+
+       
+    - name: terraform apply netweaver and hana
```
arbulu89, line 15: I would just call `terraform apply`

.github/workflows/tf-validation.yml
```diff
@@ -0,0 +1,42 @@
+# github-actions workflow
+# this test will just run terraform without salt
+name: e2e tests
+
+on: [pull_request]
+
+jobs:
+  terraform-deploy-netweaver-hana:
```
arbulu89, line 8: I would call it `terraform SAP deployment`

.github/workflows/tf-validation.yml
```diff
@@ -0,0 +1,42 @@
...
+                ## ${{ secrets.qemu_uri }}
+                ## TF_VAR_source_image: ${{ secrets.source_image }}
+     
+      run: |
+        cp .github/hana-netweaver-tf-only.tfvars libvirt/terraform.tfvars
+        cd libvirt
+        terraform workspace new gh-terraform
+        terraform workspace select gh-terraform
+        terraform init 
+        terraform apply -auto-approve
```
arbulu89, line 28: I would run `terraform plan` before the `apply`. It catches many errors without creating issues, so it's a good starting point.

.github/workflows/tf-validation.yml
```diff
@@ -0,0 +1,42 @@
...
+      env: 
+        TF_VAR_qemu_uri: "qemu+ssh://vmm@gollum-p1.qa.suse.de/system"
+        TF_VAR_source_image: "http://download.suse.de/ibs/Devel:/SAP:/Terraform:/Images/images/SLES4SAP-15_SP0-JeOS.x86_64.qcow2"
+                ## ${{ secrets.qemu_uri }}
+                ## TF_VAR_source_image: ${{ secrets.source_image }}
+     
+      run: |
+        cp .github/hana-netweaver-tf-only.tfvars libvirt/terraform.tfvars
+        cd libvirt
+        terraform workspace new gh-terraform
```
arbulu89, line 25: If you create always the workspace, the 2nd time It will fail, as it fails when a workspace exist with that name. To do this, you need to delete the workspace at the end running 
```
terraform workspace select default
terraofmr workspace delete --force gh-terraform
```
The `--force` is optional, but I would do it to assure that we will have a clean workspace next alwasy

arbulu89, line 25: I wouldn't use a workspace with a `-`. Some providers fail in that case (azure for example). For libvirt it's fine though

MalloZup: done

MalloZup: ok

.github/workflows/tf-validation.yml
MalloZup: dione

.github/hana-netweaver-tf-only.tfvars
MalloZup: I dunno how to do this and how we can update, maybe we can do it later  or sync about

MalloZup: I think we can create a follow-up issue for this

MalloZup: created

MalloZup:
as discussed, let's merge this and see how it behaves. For fixing that other problem i will open follow-up issue

Issue #474, opened 2020-05-19
The SUSEConnect -d command fails in AWS with Netweaver nodes 3 and 4


arbulu89:
The SUSEConnect -d gets blocked during the `on_destroy` stage in AWS in the NW nodes 3 and 4.

Only happens here. If you run `SUSEConnect -d` manually, you can notice that the command gets blocked.

In order to workaround this and at least destroy the AWS resources, "generic_modules/on_destroy/main.tf" has been changed to add a timeout in the destroy command.

You will find this error log after this workaround:

```
#<Thread:0x000055d2e7e61e18@/usr/lib64/ruby/2.5.0/open3.rb:264 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	1: from /usr/lib64/ruby/2.5.0/open3.rb:264:in `block (2 levels) in capture3'
/usr/lib64/ruby/2.5.0/open3.rb:264:in `read': stream closed in another thread (IOError)
#<Thread:0x000055d2e7e61c38@/usr/lib64/ruby/2.5.0/open3.rb:265 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
	1: from /usr/lib64/ruby/2.5.0/open3.rb:265:in `block (2 levels) in capture3'
/usr/lib64/ruby/2.5.0/open3.rb:265:in `read': stream closed in another thread (IOError)
```

MalloZup:
@arbulu89  closing as solved

arbulu89:
@MalloZup This issue is not fixed. Having merged PRs is just a coincidence

yeoldegrove:
This is not valid anymore as newer images use `registercloudguest` instead of `SUSEConnect`.
See https://github.com/SUSE/ha-sap-terraform-deployments/commit/ff5954abedc4e830467fe5034dfb5f300575e966 for reference.

PR #475, opened 2020-05-19
Hotfix/fix pre v500 issues

arbulu89:
Hotfixing branch to solve all the issues before creating release v5.0.0.

Fixes:
- EFS usage in AWS if Netweaver is not enabled
- DRBD creation in GCP
- Netweaver routing for GCP.
   - Depends on: https://github.com/SUSE/sapnwbootstrap-formula/pull/50
- Open port 3000 to make grafana available in the cloud providers
- Use netcat to find available NFS share. Showmount has several issues (it needs more ports, 111 and 20048, so Azure load balancer and AWS EFS doesn't serve them)
- Workaround error using SUSEConnect to deregister in AWS (doesn't fix the issue!): https://github.com/SUSE/ha-sap-terraform-deployments/issues/474

### DIFF
--- a/aws/infrastructure.tf
+++ b/aws/infrastructure.tf
@@ -242,3 +242,14 @@ resource "aws_security_group_rule" "prometheus_server" {
 
   security_group_id = local.security_group_id
 }
+
+resource "aws_security_group_rule" "grafana_server" {
+  count       = local.create_security_group_monitoring
+  type        = "ingress"
+  from_port   = 3000
+  to_port     = 3000
+  protocol    = "tcp"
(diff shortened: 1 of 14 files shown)


arbulu89:
> LGTM, although I'm not sure anymore about the need to open ports 80 and 443.
> In fact, it looks like we might need some tightening of all the security-related settings (at first glance it looks like we're opening the same the ports in all the nodes), but this topic is for another PR.

Totally agree. The current security approach is really random, and we need to work on that. In fact, we already had this topic identified and we have tickets in our backlog. This will be something we might want to implement in the next version (post `5.0.0`)

PR #477, opened 2020-05-20
Add support to extract zip,rar,exe,sar hana media

Simranpal:
Adding the feature to automatically detect the HANA archive type and extract the archive accordingly. This is related to https://github.com/SUSE/saphanabootstrap-formula/pull/84

Tested with various extraction scenarios in Libvirt, I am testing with cloud providers and will update.


Simranpal:
Hi @arbulu89 @diegoakechi  while working in these changes, I was thinking about enhancement to this feature. What do you think about asking the user just for software path, rather than asking for archive files, seperately? So we would have 1 user input variable.

Based on software path extension, we can extract or use as such for installation. As some other variables/path are dependent on the software path we may need to update those.

libvirt/variables.tf
```diff
@@ -161,14 +161,14 @@ variable "hana_sapcar_exe" {
   default     = ""
 }
 
-variable "hdbserver_sar" {
-  description = "Path to the HANA database server installation sar archive, relative to the 'hana_inst_media' mounting point"
+variable "hana_archive_file" {
+  description = "Path to the HANA installation archive, relative to the 'hana_inst_media' mounting point"
```
arbulu89, line 165: I would specify here what is a `HANA installation archive`. We have to say to the user that, this might be `HANA database server installation SAR archive` or a `HANA platform RAR file`. All the options basically. And we should add the same content in the `tfvars.example`. More help the better

libvirt/terraform.tfvars.example
```diff
@@ -23,14 +23,14 @@ hana_inst_media = "url-to-your-nfs-share:/sapdata/sap_inst_media/51053381"
...
 # Specify the path to already extracted HANA platform installation media, relative to hana_inst_media mounting point.
-# This will have preference over hdbserver sar archive installation media
+# This will have preference over hana archive installation media
 #hana_platform_folder = "51053381"
 
-# Or specify the path to the sapcar executable & HANA database server installation sar archive, relative to the 'hana_inst_media' mounting point
-# The sar archive will be extracted to path specified at hdbserver_extract_dir (optional, by default /sapmedia/HANA)
-# Make sure to use the latest/compatible version of sapcar executable, otherwise file may be extracted incorrectly
+# Or specify the path to the HANA installation archive, relative to the 'hana_inst_media' mounting point
+# If using hana sar archive, please also provide the compatible version of sapcar executable to extract the sar archive
```
arbulu89, line 30: I think we need to improve the description to talk about all the available options (SAR, RAR, ZIP, EXE), etc

arbulu89:
> Hi @arbulu89 @diegoakechi while working in these changes, I was thinking about enhancement to this feature. What do you think about asking the user just for software path, rather than asking for archive files, seperately? So we would have 1 user input variable.
> 
> Based on software path extension, we can extract or use as such for installation. As some other variables/path are dependent on the software path we may need to update those.

Yes, this might make sense. Give a try and see how it's done.

Simranpal:
> Yes, this might make sense. Give a try and see how it's done.

I will add a single user input option in a seperate PR, maybe similar changes can be also used in NW

aws/modules/hana_node/variables.tf
```diff
@@ -102,14 +102,14 @@ variable "hana_sapcar_exe" {
   default     = ""
 }
 
-variable "hdbserver_sar" {
-  description = "Path to the HANA database server installation sar archive, relative to the 'hana_inst_master' mounting point"
+variable "hana_archive_file" {
```
arbulu89, line 105: I would change a bit the description to comment that this parameter is optional, that only applies if the archive is not already extracted

PR #481, opened 2020-05-21
Update cluster pillar file for azure for better tuning

arbulu89:
Update cluster tuning for Azure to follow the configuration provided in:
https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/high-availability-guide-suse-pacemaker#default-pacemaker-configuration-for-sbd

Now, sbd resource parameters,  cluster properties, rsc_defaults and op_defaults can be updated.

@diegoakechi Eventually we would want to drop the defaults configuration done in the formulas templates, as it is really inflexible. Having the option to make changes from the pillar file it's better in my opinon.

Depends on:
https://github.com/SUSE/salt-shaptools/pull/63
https://github.com/SUSE/habootstrap-formula/pull/49

bsc#1170702

### DIFF
--- a/pillar_examples/automatic/hana/cluster.sls
+++ b/pillar_examples/automatic/hana/cluster.sls
@@ -16,6 +16,15 @@ cluster:
   {% if grains['sbd_enabled'] %}
   sbd:
     device: {{ grains['sbd_disk_device'] }}
+    {% if grains['provider'] == 'azure' %}
+    configure_resource:
+      params:
+        pcmk_delay_max: 15
+      op:
+        monitor:
+          timeout: 15
+          interval: 15
+    {% endif %}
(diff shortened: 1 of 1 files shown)


diegoakechi:
@arbulu89 I agree, but the defaults should be the defaults =P. So, if not changed, it should still assume it.

arbulu89:
> @arbulu89 I agree, but the defaults should be the defaults =P. So, if not changed, it should still assume it.

I don't get your point here. What do you mean? That we should still have the defaults in the template?

diegoakechi:
@arbulu89 no, that if not defined, AFAIK pacemaker will set the default anyway, so I think it is fine to remove it from the templates.

PR #483, opened 2020-05-22
Feature/sap scenarios

arbulu89:
This PRs provides the option to create some basic SAP scenarios, depending on if the cluster is HA or not.

**By now the changes are only for libvirt, but if you like the idea, I will expand the implementation to the other cloud providers**

It adds a new variables, `ha_enabled` (yes, yet another new variable), for HANA and NETWEAVER.

This way, 
- if `hana_ha_enabled`, HANA system replication and HA cluster will be enabled. We need to play with `hana_count` for that. (I didn't want to change the `hana_count` variable by now).
- if `netweaver_ha_enabled` is set together with `netweaver_enabled`, this will create the environment we were creating until now: ASCS, ERS, PAS and AAS. If `netweaver_ha_enabled` if false, we will create a simple NW deployment with ASCS and PAS only.

With this new variable, I think we can remove the variable `init_type`. In fact, I don't see much sense on having the option to create an empty cluster if we set `init_type=skip-hana`. @juadk , @ldevulder do you use this variable for something?

I have removed the `drbd_count` variable too. I think we only use a 2 nodes drbd cluster always. There is not any other logical scenario if drbd is used.

The `ha_enabled` data is not only for salt, depending on this the created infrastructure is different, specially on the cloud, where we don't need to create load balancers, routing entries, etc, so it makes sense.

I think that we need to provide all of this flexibility with terraform and salt, and if eventually we create a wizard to manage all of these scenarios, we can remove several variables, and just offer some customized scenarios, but the backed (terraform+salt) would still be able to manage more scenarios.


stefanotorresi:
yeah, I agree that `init_type` has to go, it's super ambiguous.
I'm okay with the new vars and especially the now hardcoded node counts, which didn't make any sense as variables because I don't think they would work at all.

PR #484, opened 2020-05-22
Create the initial version of the log retreiver

arbulu89:
This PR implements a new support tool used to get the deployment log files to the local machine.

Currently, debugging is quite annoying, as you need to log in multiple machines and find the logs.
This tool does this for us!
It copies the files using scp and getting the ip addresses from `terraform output` command.

Basically it creates a folder like `provider/workspace` and additional folder for each machine, storing there all the log files defined in the code.

Just run `python3 support/log_retriever.py -h` to see the options.

What do you think?
If you like it, I can update the docs to explain how it works and how to use.

**Things to discuss:**
- The cloud providers don't allow the ssh connections with `root` user by default (only gcp does). Actually this happens because terraform hacks the ssh authorized files, to force the user to not use `root` user. We could add a new option in terraform to authorize connections using `root`. Actually, this already exists for `libvirt`. We could describe the variable to be used in development and testing deployments. It's super easy to implement.
- I have only added some basic log files to retrieve. We could add more (SAP specific files maybe), if we run `supportconfig` during the execution for example. Or even this tool could run `supportconfig` as this is just for debugging purposes and running in the deployment itself doesn't make much sense.
- The tool has been only tested in linux, but it should work the same way using `pscp` in windows, but I have not personally tried (mostly, because I can't, I don't have a MS machine).

### DIFF
--- a/support/log_retreiver.py
+++ b/support/log_retreiver.py
@@ -1,0 +1,187 @@
+"""
+Support tool to retrieve the execution logs to the local machine
+
+:author: xarbulu
+:organization: SUSE LLC
+:contact: xarbulu@suse.com
+
+:since: 2020-05-22
+"""
+
+import os
+import argparse
+import subprocess
+import shutil
+import shlex
+import json
+import logging
+import pprint
+
+
+DESCRIPTION = """This tool provides the option to download the logs generated during the \
+deployment from all of the machines in paralell and store them in a folder so they can be debugged
+and shared easily"""
+
+OUTPUT_NAMES = {
+    "cluster_nodes_public_ip": "hana",
+    "drbd_public_ip": "drbd",
+    "netweaver_nodes_public_ip": "netweaver",
+    "iscsisrv_public_ip": "iscsi",
+    "monitoring_public_ip": "monitoring"
+}
+
+LOG_LIST = [
+    "/var/log/provisioning.log",
+    "/var/log/salt-predeployment.log",
+    "/var/log/salt-os-setup.log",
+    "/var/log/salt-deployment.log",
+    "/var/log/ha-cluster-bootstrap.log"
+]
+
+LOGGER = logging.getLogger(__name__)
+
+
+def parse_arguments():
+    """
+    Parse command line arguments
+    """
+    parser = argparse.ArgumentParser(description=DESCRIPTION)
+
+    parser.add_argument(
+        "-v", "--verbosity", default="INFO", choices=["DEBUG", "INFO", "WARN", "ERROR"],
+        help="Python logging level")
+    parser.add_argument(
+        "-p", "--provider", choices=["aws", "azure", "gcp", "libvirt"], required=True,
+        help="The provider to get the logs from")
+    parser.add_argument(
+        "-w", "--workspace",
+        help="Get the logs for a specific workspace")
+    parser.add_argument(
+        "-u", "--user", default="root",
+        help="User used to create the scp connection")
+    parser.add_argument(
+        "-i", "--identity",
+        help="Path to a SSH private key used to connect to the created machines")
+    parser.add_argument(
+        "-o", "--output-dir",
+        help="Folder where the logs will be stored")
+
+    args = parser.parse_args()
+    return parser, args
+
+
+def sanity_check(args):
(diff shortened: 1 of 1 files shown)


MalloZup:
We have already the support config tooll which gathers logs for ha and Sap. I think we should extend this tool which we own.
If we want to have a centralized fashion to call it on each machine, we could keep this approach

MalloZup:
a part of this, I think we should really be cautious with this, since this PR will add `python` as dependency for the entiere project, which can be really difficult to maintain also in windows env.

support/log_retreiver.py
```diff
@@ -0,0 +1,187 @@
+"""
```
stefanotorresi, line 1: can we add a shebang and make this executable?

arbulu89: Of course

arbulu89:
> Not super-thrilled by this, but I understand the immediate value it provides.
> 
> That said, as it is, it will be super painful to maintain because it's so tightly coupled with the Terraform conventions.
> We should consider a different approach for determining the logs to download: instead of hardcoding those, the list of log files should be part of the terraform output.
> 
> This way, when we add logs or change their location, we only do that on the infrastructure side, and the script doesn't need any changes.
> 
> Same should go for the credentials: these should be part of the terraform output, and this script shouldn't need explicit flags for these.
> 
> Also, I'm not convinced by the location of the script. I think we'll have more and more scripts going forward, so we might as well introduce a `/scripts` directory.

Thanks for the feedback @stefanotorresi ,
This is a draft PR to show you how I vision the tool to gather the logs.
Obviously, it can (and must) have improvements, but honestly I think it's quite ambitous to think that all of them will be in the first version. This script/tool/whatever will have an iterative implementation where we will add new things overtime, or do we want to have a final version already?
The things you exposed don't seem a `non go` things for me.
- We don't want to gather hardcoded file names. Agree. We need to include supportconfig and other things of course
- We can the public key from terrafrom output. Yes. Do we need to do it in the first version?
- Change `support` folder by `scripts`. I don't really mind honestly. Both look fine
- If it gets difficult to maintain because terraform changes the way it outputs (i don't really thing so), or we change the terraform outputs, we might need to find other way, but again, we cannot expect that from the 1st version cover all of this things.

The idea is to validate a tool that gathers the logs from all of the created machines by terraform, and download them to the local machine to have them available. I think we need this tool. More when people have started using the project and we have to ask them to gather the log files from all the machines.

stefanotorresi:
I didn't mean my feedback as a no-go for the idea in general: the idea is fine for me, I just long for a decoupled implementation so that we write this script now in away that _it just works_, and we immediately stop caring about it.
The proposed changes are not that gigantic, are they?

arbulu89:
> I didn't mean my feedback as a no-go for the idea in general: the idea is fine for me, I just long for a decoupled implementation so that we write this script now in away that _it just works_, and we immediately stop caring about it.
> The proposed changes are not that gigantic, are they?

Not gigantic:
- We would need to add a new output in all of the providers to output the used public key. Not difficult at all
- We would need to create a new output to define the files that must be retrieved. We could create a new simple terraform module to do so

About being tightly coupled with the Terraform, well this is the idea. To get the data from terraform (what goes along with the suggestions you did).

Ok, I will work on those changes.
Besides that, we still need to figure it out how to deal with the authorized ssh keys. Most of the log files are created by the `root` user, so to download them, the ssh connection must to be created using this user. For that, we need to authorized this key. This is quite inconvinient. The implementation is really simple, but I don't see a good practice. Maybe we need to have some other options to enable/disable this feature.

stefanotorresi:
> Besides that, we still need to figure it out how to deal with the authorized ssh keys. Most of the log files are created by the `root` user, so to download them, the ssh connection must to be created using this user. For that, we need to authorized this key. This is quite inconvinient. The implementation is really simple, but I don't see a good practice. Maybe we need to have some other options to enable/disable this feature.

Hmm, the only thing I can think of is to configure some dedicated user that has read-only access to the logs via the common salt states that are executed in all the nodes, but that's not trivial.

arbulu89:
Closing as it is stalled

PR #485, opened 2020-05-26
Feature/azure disk

arbulu89:
Implementation of the usage of multiple disks for Azure.
This PRs implements what we have in:

https://documentation.suse.com/sbp/all/html/SBP-SAP-AzureSolutionTemplates/index.html
https://github.com/SUSE/azure-resource-manager-sap-solution-templates

As unique difference this implementation uses `/dev/disk/azure/scsi1/lun` format to create the PVs, In the template scripts it uses `readlink` to find the device. My option is the one that azure docs says to use, but I don't have much idea about if there is any difference...
Besides that, everything else is exactly the same.

@petersatsuse Could you have a look too?

PD: 
- I have added the vm size and network options to the examples
- I have added the network acceleration options to the examples
- I have added a small fix in this PR to add a new load balancer rule for the hanadb_exporter

I will implement a similar thing for Netweaver in a next PR


petersatsuse:
Overall the new way for different disk usage looks good. The usage of the lunX needs to be changed, as they are not really persistent and the creation of the pv needs to have persistent devicenames (by uuid , as its the only way in linux as of today).

As i'm a newbet with TF and salt, is it really necessary to have disk configuration in the file terraform.tfvars or could it be "hidden" further down without the need to show up in tfvars at all?

arbulu89:
> Overall the new way for different disk usage looks good. The usage of the lunX needs to be changed, as they are not really persistent and the creation of the pv needs to have persistent devicenames (by uuid , as its the only way in linux as of today).
> 
> As i'm a newbet with TF and salt, is it really necessary to have disk configuration in the file terraform.tfvars or could it be "hidden" further down without the need to show up in tfvars at all?

Thanks for the review @petersatsuse ,
I will change how the luns and disks ID are managed in general.

About the 2nd question. I think it's preferable to have this way, giving the user the option to decide how they want to configure the disks. Later on, if we implement the wizard, this will only ask for the deployment size, and configure all of this values automatically. This way we have the 2 options, a high level configuration for advanced users, and guided configuration with the wizard. I don't usually like to hardcode this parameters as they tend to be volatile.

arbulu89:
Hi @petersatsuse @diegoakechi 
I have included the implementation to mount the disks and add them to fstab using th UUID.
As discussed internally, there is not a way to create the pv-s using a UUID, as Azure has its own udev rules that map the disks in `/dev/disks/azure/iscsi1/lunx`, so there is no any entry in `/dev/disks/by-uuid` for these disks.
Anyway, for the moment this is what azure docs comment too. If in the future this changes, we can modify this implementation as well

arbulu89:
@diegoakechi @petersatsuse @Simranpal I think this PR is ready to be merged. All the requirements that I commented with Peter about disks usage are done.
Have a look and approve if you think it's OK.

diegoakechi:
> PD: Diego, one thing we need to think about is if this configuration should belong to the `saphanabootstrap-formul` or not.

AFAIK the storage part is quite variable and can be composed in many different ways, so it is quite hard to make it generic enough to be part of the formula. It looks very specific for how Azure configures the disks at the moment. 
@petersatsuse  What do you think?

petersatsuse:
Different disk types and different lvm settings you have on the others too, so this is really generic

The only real Azure specials are as of now the "write accelerator" and the "caching" setting. Everything else is true for the the others too.

Looking at:
AWS - has some additional parameters too, but it seems they need to be set on instance level
GCE - set additional parameters, like caching too, but on within the OS eg. as mount parameter

So overall i think the two columns do not hurt and could additional filled too depending on the cloud providers functionality. 

As HANA has (more or less) fixed names for the different disk areas i'm not sure what the note with the saphanabootstrap belongs to.

arbulu89:
@diegoakechi @petersatsuse "write accelerator" and "caching" are part of the infrastructure, so this wouldn't be involved in the salt formula.
The salt part only creates the lvm units and mount the new partitions. By now, the unique dependency here is the initial disk name `/dev/disk/azure/iscsi1/lun0` that is azure dependant, but besides that everything is just Linux dependant, so for gcp/aws should be doable.
The question should be if we are going to use the same disks configuration for them.

Anyway, by now, we will keep the code here, and we can reopen the discussion in the future. 
As we don't have many users only wanting to use the salt formula it's not really needed to have them in the formula. If we see that for example, users would like to have these configuration in SUMA, it would make sense, as it's almost the most difficult part of the configuration hehe

arbulu89:
> Looks good to me!
> About moving the disk configuration part to hana formula, I think it makes sense to use it in formula, as its HANA specific parameters

I wouldn't say this is HANA specific. HANA doesn't require such a configuration to be running, this is just a better one to improve performance.

diegoakechi:
@arbulu89 I agree in keeping the code here for now. Moving to the formula needs to cover the other providers and also bare metal, that is the part that I am not sure. For example, I know that many customers maintains a AutoYast profile that takes care of the partitions, formating, mounting etc... And it can be integrated on SUMA. It is not possible on Public Cloud since the machines are image based. So with that, we would need some research to evaluate if it would bring value or not.

PR #488, opened 2020-05-28
Feature/cp sap scenarios

arbulu89:
Follow up for: https://github.com/SUSE/ha-sap-terraform-deployments/pull/483

This PR implements the changes started in the commented PR for the cloud providers.

- Add `hana_ha_enabled` and `netweaver_ha_enabled` variables
- Use `netweaver_count` better depending on the deployment is HA or not
- Remove `init_type` variable
- Add `sbd_enabled` option for DRBD in AWS

Next steps in future PRs:
- Create the infrastructure depending on `ha_enabled`. I have not implemented that for not making this PR bigger, as every cloud provider has differences and better to address them independently

### DIFF
--- a/aws/main.tf
+++ b/aws/main.tf
@@ -29,11 +29,12 @@ locals {
 
   # range(4) hardcoded as we always deploy 4 nw machines
   netweaver_ip_start    = 30
-  netweaver_ips         = length(var.netweaver_ips) != 0 ? var.netweaver_ips : [for index in range(4) : cidrhost(element(local.netweaver_subnet_address_range, index % 2), index + local.netweaver_ip_start)]
-  netweaver_virtual_ips = length(var.netweaver_virtual_ips) != 0 ? var.netweaver_virtual_ips : [for ip_index in range(local.netweaver_ip_start, local.netweaver_ip_start + 4) : cidrhost(var.virtual_address_range, ip_index)]
+  netweaver_count       = var.netweaver_enabled ? (var.netweaver_ha_enabled ? 4 : 2) : 0
+  netweaver_ips         = length(var.netweaver_ips) != 0 ? var.netweaver_ips : [for index in range(local.netweaver_count) : cidrhost(element(local.netweaver_subnet_address_range, index % 2), index + local.netweaver_ip_start)]
+  netweaver_virtual_ips = length(var.netweaver_virtual_ips) != 0 ? var.netweaver_virtual_ips : [for ip_index in range(local.netweaver_ip_start, local.netweaver_ip_start + local.netweaver_count) : cidrhost(var.virtual_address_range, ip_index)]
 
   # Check if iscsi server has to be created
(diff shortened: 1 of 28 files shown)


aws/main.tf
```diff
@@ -29,11 +29,12 @@ locals {
 
   # range(4) hardcoded as we always deploy 4 nw machines
   netweaver_ip_start    = 30
-  netweaver_ips         = length(var.netweaver_ips) != 0 ? var.netweaver_ips : [for index in range(4) : cidrhost(element(local.netweaver_subnet_address_range, index % 2), index + local.netweaver_ip_start)]
-  netweaver_virtual_ips = length(var.netweaver_virtual_ips) != 0 ? var.netweaver_virtual_ips : [for ip_index in range(local.netweaver_ip_start, local.netweaver_ip_start + 4) : cidrhost(var.virtual_address_range, ip_index)]
+  netweaver_count       = var.netweaver_enabled ? (var.netweaver_ha_enabled ? 4 : 2) : 0
```
Simranpal, line 32: I am wondering if we should make the count logic 2 or 4 a variable. Maybe in future, we need to consider more/less no of nw machines

aws/terraform.tfvars.example
```diff
@@ -189,10 +188,27 @@ ha_sap_deployment_repo = ""
...
 #netweaver_cluster_sbd_enabled = true
 
 # Netweaver installation required folders
+# This S3 bucket must contain the next software (select the version you want to install of course)
+#SWPM - `IND:SLTOOLSET:2.0:SWPM:*:LINUX_X86_64:*x`
+#Netweaver export - `SAP:NETWEAVER:750:DVD_EXPORT:SAP NetWeaver 750 Installation Export DVD 1/1:D51050829_2`
+#HANA Platform- `HDB:HANA:2.0:LINUX_X86_64:SAP HANA PLATFORM EDITION 2.0::BD51053787`
+#Sapexe folder files:
+#igsexe_23-20007790.sar  igshelper_4-10010245.sar  SAPEXE_501-80002573.SAR  SAPEXEDB_501-80002572.SAR  SAPHOSTAGENT45_45-20009394.SAR
+
```
Simranpal, line 211: I guess we don't need to mention the full version infos, but its fine as its example

aws/main.tf
arbulu89: Yes, this might come most probably in the future. Anyway, if we think about it, a logical deployment only consists on:
- 1 ASCS
- 1 ERS if the deployment is HA, so this is optional
- 1 PAS
- X AAS (these are optional, and each deployment might have a different number)

So here, the logic would we to ask if the deployment is HA or NOT, and how many application servers to deploy (being 1 only the PAS and beyond that PAS + AAS * x)

aws/terraform.tfvars.example
arbulu89: Yeah, the versions are just example, but I think it's good as reference

PR #489, opened 2020-06-01
Feature/azure netweaver multiple aas

arbulu89:
This PR implements improvements in Netweaver deployment;

- Create the correct availability sets
- Create load balancer depending on the deployment is HA or not (this has been added to HANA too)
- Give the option to deploy a configurable number of application servers
- Document the different configuration sizes for netweaver
- Provision properly the netweaver application servers disk (xscs instances don't need an additional disk).

Could you have a look @petersatsuse too ?

Here some context about the types: https://documentation.suse.com/sbp/all/html/SBP-SAP-AzureSolutionTemplates/index.html


Simranpal:
Do we need to enable the `writeAccelerator` feature for hana log disk? as its mentioned in the guide

azure/README.md
```diff
@@ -194,6 +194,83 @@ hana_data_disks_configuration = {
...
+
+#### Large
+
+```
+netweaver_xscs_vm_size = "Standard_D2s_v3"
+netweaver_app_vm_size = "Standard_E64s_v3"
+netweaver_data_disk_type = "Premium_LRS"
+netweaver_data_disk_size = 128
+netweaver_data_disk_caching = "ReadWrite"
+netweaver_xscs_accelerated_networking = false
```
Simranpal, line 269: in the variable description you mentioned this feature is mandatory for nw xSCS machines, but I did not see it set to true

salt/netweaver_node/init.sls
```diff
@@ -1,4 +1,5 @@
 include:
+  - netweaver_node.mount
```
Simranpal, line 2: I do not see a `mount.sls` file in this directory

Simranpal:
@arbulu89  I also see there are other settings in the template document(4.3 Other Settings) like below:
```
All machines use the Azure feature “Accelerated Network (AN)” (only if the VM size supports AN).
All AVSets for the APP machines use a FaultDomainCount of 2 and an UpdateDomainCount of 20.
```

And also for the various instances sizes, we are setting the fixed number of instances, but I cant find the numbers in document.

azure/modules/hana_node/main.tf
```diff
@@ -1,6 +1,11 @@
 # Availability set for the hana VMs
 
+locals {
+  create_ha_infra = var.hana_count > 0 && var.ha_enabled ? 1 : 0
```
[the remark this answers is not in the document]
arbulu89: Ok, I see your point. You mean we should change this to `var.hana_count > 1`, right?
Yes, you are right

azure/modules/hana_node/main.tf
```diff
@@ -34,21 +40,23 @@ resource "azurerm_lb" "hana-load-balancer" {
 # backend pools
 
 resource "azurerm_lb_backend_address_pool" "hana-load-balancer" {
+  count               = local.create_ha_infra
   resource_group_name = var.resource_group_name
-  loadbalancer_id     = azurerm_lb.hana-load-balancer.id
+  loadbalancer_id     = azurerm_lb.hana-load-balancer[0].id
```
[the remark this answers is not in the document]
arbulu89: As long as it has a `count`, it's a list

azure/README.md
arbulu89: Yes, you are right. I just used the values from the template, and there only app instances use this value. I will change the variable description

salt/netweaver_node/init.sls
arbulu89: Thanks! I just missed to upload it in my git commit as it is a new folder

arbulu89:
> @arbulu89 I also see there are other settings in the template document(4.3 Other Settings) like below:
> 
> ```
> All machines use the Azure feature “Accelerated Network (AN)” (only if the VM size supports AN).
> All AVSets for the APP machines use a FaultDomainCount of 2 and an UpdateDomainCount of 20.
> ```
> 
> And also for the various instances sizes, we are setting the fixed number of instances, but I cant find the numbers in document.

Those values are set. Check the azure netweaver module. `netweaver-app-availability-set` and the `enable_accelerated_networking` in `azurerm_network_interface`

arbulu89:
> Do we need to enable the `writeAccelerator` feature for hana log disk? as its mentioned in the guide

It's enabled, but only for the `Large` deployment. Actually this was done in https://github.com/SUSE/ha-sap-terraform-deployments/pull/485

azure/modules/hana_node/main.tf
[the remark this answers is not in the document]
Simranpal: Yes exactly

azure/README.md
```diff
@@ -194,6 +194,83 @@ hana_data_disks_configuration = {
 }
 ```
 
+## Netweaver configuration
+
+There are different Netweaver configurations available depending on if the environment is an HA one or not, and the expected performance. Here some guidelines for the different options:
+
+To set a 3-tier deployment (simple Netweaver, without HA) use the next options:
```
diegoakechi, line 201: Isn't it a 2-tier deployment?

Simranpal: It is 3-tier as we are supporting the deployment of the sap application layer servers (middle layer), along with database. The 2 tier would be just the database server, if I understand the terminilogy correct :)

https://azure.microsoft.com/en-ca/blog/accelerate-your-sap-on-azure-hana-project-with-suse-microsoft-solution-templates/

arbulu89: Exactly. Thanks @Simranpal

PR #490, opened 2020-06-04
Add support to extract nw media archives

Simranpal:
This PR is related to changes in https://github.com/SUSE/sapnwbootstrap-formula/pull/51
There are some additional changes:
-I have removed `swpm_extract_dir` and replaced with `nw_extract_dir` which will now store all extracted NW media
-Also removed the hardcoded `'/sapmedia/NW' ` and replace with `netweaver_inst_folder`
-Updated the salt mount state` installation_files.sls` for nw media with the `netweaver_inst_folder` usage

I have tested it with SLES15SP1 in libvirt provider


aws/terraform.tfvars.example
```diff
@@ -213,13 +213,12 @@ ha_sap_deployment_repo = ""
 # SAP SWPM installation folder, relative to the netweaver_s3_bucket folder
 #netweaver_swpm_folder     =  "your_swpm"
 # Or specify the path to the sapcar executable & SWPM installer sar archive, relative to the netweaver_s3_bucket folder
-# The sar archive will be extracted to path specified at netweaver_swpm_extract_dir (optional, by default /sapmedia/NW/SWPM)
+# The sar archive will be extracted to path specified at netweaver_extract_dir under SWPM directory (optional, by default /sapmedia/NW/SWPM)
```
arbulu89, line 216: Is this default value correct? Shouldn't be `/sapmedia/NW`

Simranpal: Yes. The `SWPM` sub directory is hardcoded. So extraction will be under  `netweaver_extract_dir_path/SWPM/`

PR #497, opened 2020-06-08
Feature/azure bastion

arbulu89:
Implementation of the bastion host. This machine is used as a jumpbox to avoid the creation of public addresses for the rest of machines.
To use it, set bastion_enabled = true (enabled by default). If set to false the bastion machine is not created and all of the machines will have a public address, as we do before this implementation.

The bastion is just a basic Standard_D2s_v3 machine using sle15sp1. This data is hardcoded. It only enables the coming SSH connections.

The machine is created in a new subnet with a custom security group, so now, we will have 2 subnets:
- One for the bastion
- One for the rest of the machines

@petersatsuse Have a look if you want?

### DIFF
--- a/azure/modules/bastion/main.tf
+++ b/azure/modules/bastion/main.tf
@@ -87,7 +87,7 @@ resource "azurerm_virtual_machine" "bastion" {
   location                         = var.az_region
   resource_group_name              = var.resource_group_name
   network_interface_ids            = [azurerm_network_interface.bastion[0].id]
-  vm_size                          = "Standard_D2s_v3"
+  vm_size                          = var.vm_size
   delete_os_disk_on_termination    = true
   delete_data_disks_on_termination = true
 
--- a/azure/modules/bastion/main.tf
+++ b/azure/modules/bastion/main.tf
@@ -1,0 +1,130 @@
+locals {
+  bastion_enabled = var.bastion_enabled ? 1 : 0
+}
+
+
+resource "azurerm_subnet" "bastion" {
+  count                = local.bastion_enabled
+  name                 = "snet-bastion"
+  resource_group_name  = var.resource_group_name
+  virtual_network_name = var.vnet_name
+  address_prefix       = var.snet_address_range
+}
+
+resource "azurerm_network_security_group" "bastion" {
+  count               = local.bastion_enabled
+  name                = "nsg-bastion"
+  location            = var.az_region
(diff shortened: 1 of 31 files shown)


azure/modules/bastion/main.tf
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+
+resource "azurerm_virtual_machine" "bastion" {
+  count                            = local.bastion_enabled
+  name                             = "vmbastion"
+  location                         = var.az_region
+  resource_group_name              = var.resource_group_name
+  network_interface_ids            = [azurerm_network_interface.bastion[0].id]
+  vm_size                          = "Standard_D2s_v3"
```
diegoakechi, line 90: @arbulu89 I think it might be quite big just for a SSH jump host (it has 8GB RAM, for example). I would try something really small like Standard_B1s or Standard_B1ms

generic_modules/salt_provisioner/variables.tf
```diff
@@ -26,6 +26,12 @@ variable "private_key_location" {
   default     = ""
 }
 
+variable "bastion_host" {
+  description = "Address of a bastion host to use in the SSH connections. It uses the same user/private key combo of the normal connection. Let empty to use the normal connection"
```
diegoakechi, line 30: @arbulu89 Not sure if we should use the same SSH key here as the ones used for the Cluster machines configuration.

azure/modules/bastion/main.tf
arbulu89: I will test with this machine type

generic_modules/salt_provisioner/variables.tf
arbulu89: Why not? We could add other keys, but we would ask to the user 2 ssh key pairs, that honestly makes everything more difficult without adding almost any security. Remember that they cannot access the machines externally.
It's obviously doable though.

What do you think if I create 2 additional variables to provide bastion keys, but if they are not provided I use the currently existing options?

diegoakechi: I am fine with we move this way.

azure/modules/bastion/main.tf
diegoakechi: Maybe would make sense to move the `vm_size` as a module variable, instead of having it hardcoded. This way, advanced users could change it via `main.tf` editing. what do you think?

PR #498, opened 2020-06-10
Register the cloud extension in libvirt to get python3-pip

arbulu89:
We need `python3-pip` to install the SAP `dbapi` python library. `python3-pip` is only available in the cloud extension in SLE12. Until now, we were adding the cloud extension only for the cloud providers, but now we need to add it in libvirt too.

**This only affects to SLE12 and libvirt**

Related: https://github.com/SUSE/saphanabootstrap-formula/pull/89

### DIFF
--- a/salt/os_setup/registration.sls
+++ b/salt/os_setup/registration.sls
@@ -23,7 +23,6 @@ default_sle_module_adv_systems_management_registration:
         attempts: 3
         interval: 15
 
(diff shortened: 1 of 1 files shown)


stefanotorresi:
I'm not quite clear: as it is, would the module be registered in SLE15 too?
Shouldn't we have something like an `if grains['osmajorrelease'] == 12` block around the task?

PR #499, opened 2020-06-11
Correct the iscsi target package in SLE12SPx.

nick-wang:
Package `targetcli-fb-common` and `python3-targetcli-fb` not exist in SLE12SP4.

### DIFF
--- a/salt/cluster_node/ha/iscsi_initiator.sls
+++ b/salt/cluster_node/ha/iscsi_initiator.sls
@@ -13,7 +13,7 @@ lsscsi:
 /etc/iscsi/initiatorname.iscsi:
   file.replace:
     - pattern: "^InitiatorName=.*"
-    - repl: "InitiatorName=iqn.{{ grains['server_id'] }}.suse.qa"
+    - repl: "InitiatorName=iqn.{{ grains['hostname'] }}.{{ grains['server_id'] }}.suse"
 
 /etc/iscsi/iscsid.conf:
   file.replace:
--- a/salt/pillar/iscsi_srv.sls
+++ b/salt/pillar/iscsi_srv.sls
@@ -6,8 +6,13 @@ iscsi:
   target:
     pkgs:
       wanted:
+      {%- if grains['osmajorrelease'] == 12 %}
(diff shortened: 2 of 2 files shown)


arbulu89:
@nick-wang Debugging a bit more I have found this log in the failing nodes in `/var/log/messages`:
```
2020-06-11T12:31:21.812536+02:00 linux iscsid[11892]: iscsid: Kernel reported iSCSI connection 1:0 error (1020 - ISCSI_ERR_TCP_CONN_CLOSE: TCP connection closed) state (3)
2020-06-11T12:31:23.814005+02:00 linux iscsid[11892]: iscsid: connection1:0 is operational after recovery (1 attempts)
2020-06-11T12:31:24.926040+02:00 linux kernel: [  857.337894]  connection1:0: detected conn error (1020)
```
This log is there a lot of times

nick-wang:
> My issue happens when we have iscsi server using SLE15, and the hana/drbd machines using SLE15 SP1.

I never tried like that, will try this in the next week.
Thanks for the information

nick-wang:
@arbulu89 I think the new commit of remove the duplicate initiator name could help on the error. It may help to fix the CI failure on SLE12SP4.
The test passed in my test on gollum.

I think we can try this in CI

arbulu89:
@nick-wang I will test this change. Anyway, the issue doesn't only happen in the CI. I can reproduce it quite easily running terraform locally.

nick-wang:
> Anyway, what do you mean by duplicated initiator? `{{ grains['server_id'] }}` is different for each machine, so it shouldn't be duplicated never

In theory, it should be. But the `{{ grains['server_id'] }}` are the same in my testing on gollum...
Refer to the log(from the failed CI task) pasted in rchat (to you), you can find the `server_id` is actually the same on different nodes...

nick-wang:
> Update2: Nick, could you remove the `qa` part from `"InitiatorName=iqn.{{ grains['hostname'] }}.{{ grains['server_id'] }}.suse.qa"` ?

Done in the latest (rebased) commit

PR #502, opened 2020-06-15
Replace rclone by gcloud tools

arbulu89:
Implementation to replace `rclone` by the official GCP tools (gcloud and gsutil). We cannot use non official 3 party tools like `rclone`, so using the GCP provided tools make sense.
The usage for the user will be exactly the same (it must provide the bucket path and the credentials file). 

I have created the implementation in a macro in order to avoid duplicated code, but if it's too confusing I can just copy/paste the states in HANA and NW.

PD: Even though it's not strictly needed, I have added 2 states to configure the `gcloud` usage so the user can use it later. It's a quite standard tool, so it makes sense. I can remove them anyway (I'm talking about `add_gcloud_path` and `restart_shell` states)

Some info about the official docs:
https://cloud.google.com/sdk/docs/downloads-interactive
https://www.ecosia.org/search?q=gsutil+cp&addon=firefox&addonversion=4.0.4

Adding @petersatsuse

### DIFF
--- a/salt/hana_node/download_hana_inst.sls
+++ b/salt/hana_node/download_hana_inst.sls
@@ -36,19 +36,14 @@ hana_inst_directory:
     - require:
       - cmd: hana_inst_partition
 
-install_rclone:
-  cmd.run:
-    - name: "curl https://rclone.org/install.sh | sudo bash"
 
-configure_rclone:
-  file.append:
-    - name: /root/.rclone.conf
-    - source: /root/salt/hana_node/files/rclone/gcp.conf
+{% from 'macros/download_from_google_storage.sls' import download_from_google_storage with context %}
 
-download_files_from_gcp:
-  cmd.run:
-    - name: "rclone copy remote:{{ grains['sap_hana_deployment_bucket'] }}
-      {{ grains['hana_inst_folder'] }}"
+{{ download_from_google_storage(
+  grains['gcp_credentials_file'],
+  grains['sap_hana_deployment_bucket'],
+  grains['hana_inst_folder']) }}
+
 {% endif %}
 
 {{ grains['hana_inst_folder'] }}:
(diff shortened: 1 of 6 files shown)


salt/macros/download_from_google_storage.sls
```diff
@@ -0,0 +1,49 @@
...
+      - install_gcloud
+
+configure_gcloud_credentials:
+  cmd.run:
+    - name: {{ gcloud_bin_dir }}/gcloud auth activate-service-account --key-file {{ credentials_file }}
+    - require:
+      - install_gcloud
+
+# I have not found any option to have something like os.path.join, what would be better than this url transformation
+{% set gs_url = 'gs://'~bucket_path~ ('/' if bucket_path[-1] != '/' else '')~'*' %}
```
Simranpal, line 39: You can try jinja's` join()` or salt's jinja filter `path_join()`

https://jinja.palletsprojects.com/en/2.11.x/templates/#join
https://docs.saltstack.com/en/latest/topics/jinja/index.html

salt/hana_node/download_hana_inst.sls
```diff
@@ -36,19 +36,14 @@ hana_inst_directory:
...
 
-install_rclone:
-  cmd.run:
-    - name: "curl https://rclone.org/install.sh | sudo bash"
 
-configure_rclone:
-  file.append:
-    - name: /root/.rclone.conf
-    - source: /root/salt/hana_node/files/rclone/gcp.conf
+{% from 'macros/download_from_google_storage.sls' import download_from_google_storage with context %}
```
Simranpal, line 40: maybe enclose the statement in the `-` but its fine

salt/macros/download_from_google_storage.sls
arbulu89: Thanks @Simranpal I spent 1 hour looking for this and I didn't find it (and I knew that something should exist hehe).
Changing right away

PR #504, opened 2020-07-10
Use disk id instead of disk name to setup sbd in iscsi setup

arbulu89:
Change to improve the usage of Sbd. Until now we were using the disk name to setup Sbd (`/dev/sda` for example). This is not a safe way, as the disk name might be changed in any moment. 
So, when we use iscsi, the sbd disk id is calculated after saying to each machine which LUN number to use for setup each cluster Sbd setup. For that, I have replaced `sbd_disk_index` by `sbd_lun_index` to be more meaningful, being this number, the LUN number of all of the LUNs provided by our iscsi server (that by default is 3).

@gao-yan Is this implementation good to get the disk id? In my tests it works, but I would appreciate if you have at least a look on that.


arbulu89:
> LGTM, have you checked by any chance if the HA Cluster dashboard reports the new names correctly? just to be sure.

Thanks for reminding this! Yes, it looks it's displayed properly:
image

salt/cluster_node/ha/sbd.sls
```diff
@@ -1,5 +1,7 @@
 # This state is executed at the end of iscsi_initiator.sls state
-{% set sbd_disk_device = salt['cmd.run']('lsscsi | grep "LIO-ORG" | awk "{ if (NR=='~grains['sbd_disk_index']~') print \$NF }"', python_shell=true) %}
+{% set sbd_disk_id_pattern = '/^\[[0-9]\:[0-9]\:[0-9]\:'~grains['sbd_lun_index']~'\].*/' %}
+{% set sbd_disk_device = salt['cmd.run']('lsscsi -i | grep "LIO-ORG" | awk "'~sbd_disk_id_pattern~'{print \$NF }"', python_shell=true) %}
+{% set sbd_disk_device = '/dev/disk/by-id/scsi-'~sbd_disk_device %}
```
Simranpal, line 4: is the device name include `scsi / iscsi`

arbulu89: scsi

PR #507, opened 2020-07-15
Fix lsscsi command usage to manage `-` output

arbulu89:
Patch to fix the last https://github.com/SUSE/ha-sap-terraform-deployments/pull/504 changes.

In some sporadic cases (specially in the cloud instances) the output of `lsscsi -i` doesn't return the disk id until some seconds are passed. This was generating an output like this: 
```
[2:0:0:0]    disk    ATA      SK hynix SC311 S 0P10  /dev/sda   -
```
And the next device creation: `/dev/disk/by-id/scsi--`

To fix this, I have included an active checking system to wait until the disk id is available in the command output to proceed with the correct data

### DIFF
--- a/salt/cluster_node/ha/iscsi_initiator.sls
+++ b/salt/cluster_node/ha/iscsi_initiator.sls
@@ -47,6 +47,17 @@ iscsid:
     - watch:
       - cmd: iscsi_discovery
 
+# Wait until the disk id is displayed in the lsscsi command, as `-` is displayed at the beginning sometimes
+{% set sbd_disk_id_pattern = '/^\[[0-9]\:[0-9]\:[0-9]\:'~grains['sbd_lun_index']~'\].*/' %}
+
+wait_disk_id_available:
+  cmd.run:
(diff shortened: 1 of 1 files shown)


salt/cluster_node/ha/iscsi_initiator.sls
```diff
@@ -47,6 +47,17 @@ iscsid:
     - watch:
       - cmd: iscsi_discovery
 
+# Wait until the disk id is displayed in the lsscsi command, as `-` is displayed at the beginning sometimes
+{% set sbd_disk_id_pattern = '/^\[[0-9]\:[0-9]\:[0-9]\:'~grains['sbd_lun_index']~'\].*/' %}
+
+wait_disk_id_available:
+  cmd.run:
+    - name: until [ "$(lsscsi -i | grep "LIO-ORG" | awk "{{ sbd_disk_id_pattern }}{print \$NF }")" != "-" ];do sleep 3;done
+    - timeout: 120
```
Simranpal, line 56: This timeout should be good, I guess we can also add retries if it takes some more time to get the disk ID

arbulu89: Usually it takes around 5/10 seconds. If in this time it doesn't work, it means that something went wrong. In any case, I prefer the `until` solution rather than the retries solution in this case. It lets the salt logs output easier to understand

PR #508, opened 2020-07-15
Feature/hana active active

arbulu89:
HANA Active/Active implementation.The active/active setup permits the read-only access the the secondary machine, giving the option to split the queries in the 2 machines. The implementation creates all of the required infrastructure for each providers (azure: load balances and rules, gcp/aws: route entries

To use it use the next variables in the `terraform.tfvars`:
```
hana_active_active = true
#hana_cluster_vip_secondary = 192.168.10.10 # This IP will be auto generated if it's not provided
```

I have tested in all the providers with and without this option, and everything worked fine.

Docs about Active/Active setup:
https://documentation.suse.com/sbp/all/html/SLES4SAP-hana-sr-guide-PerfOpt-12/index.html#_activeactive_read_enabled_scenario
https://help.sap.com/viewer/4e9b18c116aa42fc84c7dbfd02111aba/2.0.04/en-US/8231617177f743d1ba46fdfc5a82dcd1.html

Depends on: https://github.com/SUSE/saphanabootstrap-formula/pull/94


stefanotorresi:
so how should the exporter integrate with this? I guess we have to monitor both nodes, right?

arbulu89:
> so how should the exporter integrate with this? I guess we have to monitor both nodes, right?

You are right, we would need to think in that possibility. I will create a follow up ticket in our backlog, so we can plan it for a future addition to this feature

azure/modules/hana_node/main.tf
```diff
@@ -33,6 +45,17 @@ resource "azurerm_lb" "hana-load-balancer" {
     private_ip_address            = var.hana_cluster_vip
   }
 
+  # Create a new frontend for the Active/Active scenario
+  dynamic "frontend_ip_configuration" {
```
Simranpal, line 49: What is the front end configuration used here for azure

azure/modules/netweaver_node/main.tf
```diff
@@ -137,184 +153,30 @@ resource "azurerm_lb_probe" "netweaver-ers-health-probe" {
...
-}
-
-resource "azurerm_lb_rule" "netweaver-lb-ascs-5xx14" {
-  count                          = local.create_ha_infra
+resource "azurerm_lb_rule" "ascs-lb-rules" {
+  for_each                       = local.ascs_lb_rules_ports
   resource_group_name            = var.resource_group_name
   loadbalancer_id                = azurerm_lb.netweaver-load-balancer[0].id
-  name                           = "lbrule-neweaver-ascs-tcp-5${var.ascs_instance_number}14"
+  name                           = "lbrule-neweaver-ascs-tcp-${each.value}"
```
Simranpal, line 160: `neweaver/netweaver`

azure/modules/hana_node/main.tf
arbulu89: I don't really understand the question. This `dynamic` entry only creates an additional entry to the load balancer with the `hana_cluster_vip_secondary` address (only if the active/active option is enabled). This way, we can use a 2nd floating address managed by the cluster. The configuration is the same than the first one, only with a different address

PR #510, opened 2020-07-17
Refactor repo states and remove devel_mode

stefanotorresi:
This PR refactors all the generic repository configuration into the same `repo.sls` file, and removes the `devel_mode` tfvar, which was basically useless since we moved all the packages into the dedicated OBS project: setting the `ha_sap_deployment_repo` will implicitly disable zypper vendor stickiness.


stefanotorresi:
> @stefanotorresi So now, setting the `ha_sap_deployment_repo` means that we are like in the old `devel_mode`, right?
> The difference I see is that now we don't set any priority to the ha repository, so in the case the packages are coming from the official SSC channels, how are we going to be able to enable the development repository? Remember that the last purpose is to get the packages from there (the official SCC channels, and not from our team obs repository)

Hmm, I'm not sure I understand the question. Why would a user set the devel repo if they don't want to use it?

arbulu89:
> > @stefanotorresi So now, setting the `ha_sap_deployment_repo` means that we are like in the old `devel_mode`, right?
> > The difference I see is that now we don't set any priority to the ha repository, so in the case the packages are coming from the official SSC channels, how are we going to be able to enable the development repository? Remember that the last purpose is to get the packages from there (the official SCC channels, and not from our team obs repository)
> 
> Hmm, I'm not sure I understand the question. Why would a user set the devel repo if they don't want to use it?

What I mean is that even setting the devel repo, if the priority is not higher than the official channels repositories, it won't be used never. I guess this is not what we want, right? Most probably we want to define a higher priority to the devel repo, so it is used if we use the variable

stefanotorresi:
> What I mean is that even setting the devel repo, if the priority is not higher than the official channels repositories, it won't be used never. I guess this is not what we want, right? Most probably we want to define a higher priority to the devel repo, so it is used if we use the variable

Ahh, I see what you mean now. My understanding is that we don't need to set the priority explicitly because the version numbers in the devel repo will always be higher, we only need to disable vendor stickiness to allow upgrading from official channels to devel repos. Am I wrong?

For instance, right now this is how `prometheus-ha_cluster_exporter` is resolved with the changes in this PR:
```
hana01:~ # zypper se --details prometheus-ha_cluster_exporter
Refreshing service 'Basesystem_Module_15_SP2_x86_64'.
Refreshing service 'Desktop_Applications_Module_15_SP2_x86_64'.
Refreshing service 'SAP_Applications_Module_15_SP2_x86_64'.
Refreshing service 'SUSE_Linux_Enterprise_High_Availability_Extension_15_SP2_x86_64'.
Refreshing service 'SUSE_Linux_Enterprise_Server_for_SAP_Applications_15_SP2_x86_64'.
Refreshing service 'Server_Applications_Module_15_SP2_x86_64'.
Loading repository data...
Reading installed packages...

S  | Name                           | Type       | Version                           | Arch   | Repository
---+--------------------------------+------------+-----------------------------------+--------+---------------------------------------
i+ | prometheus-ha_cluster_exporter | package    | 1.1.0+git.1594637514.c1eae0c-44.1 | x86_64 | ha_sap_deployments
v  | prometheus-ha_cluster_exporter | package    | 1.0.1+git.1588608085.3a3faf7-1.6  | x86_64 | SLE-Module-SAP-Applications15-SP2-Pool
   | prometheus-ha_cluster_exporter | srcpackage | 1.1.0+git.1594637514.c1eae0c-44.1 | noarch | ha_sap_deployments
```

As you can see, the development version was installed instead of the one provided by `SLE-Module-SAP-Applications15-SP2-Pool`.

stefanotorresi:
okay! I'll keep this in mind!

stefanotorresi:
yes, good call!

PR #512, opened 2020-07-20
Create common_variables module

arbulu89:
I have create a new terraform module `common_variables` to store the variables that are shared among all of the modules, like the registration code, email, ha repo, etc

The idea of the implementation is to reduce the amount of variables we need to pass from the main terraform file to the modules, wrapping them all in a unique module. It reduces a lot the amount of duplicated code.

After this, it will be easier to add/remove new variables that are common to all the modules (like removing the `qa_mode` and adding new variables like `colored_output` or `offline_provisioning`.

Let me know what do you think? If you like the approach, I will expand the implementation to the other cloud providers, and eventually replace the `qa_mode` variable by other more meaningful options

### DIFF
--- a/libvirt/main.tf
+++ b/libvirt/main.tf
@@ -38,30 +38,39 @@ locals {
   iscsi_enabled = var.sbd_storage_type == "iscsi" && (var.hana_count > 1 && var.hana_cluster_sbd_enabled == true || (var.drbd_enabled && var.drbd_cluster_sbd_enabled == true) || (local.netweaver_count > 1 && var.netweaver_cluster_sbd_enabled == true)) ? true : false
 }
 
-module "iscsi_server" {
-  source                 = "./modules/iscsi_server"
-  iscsi_count            = local.iscsi_enabled == true ? 1 : 0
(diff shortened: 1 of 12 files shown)


arbulu89:
> something more
> 
> `global_variables` `global_vars` or `shared_variables` maybe is ok imho.
> 
> I think `common_variables` is ok too,, but maybe less strong. DUnno what the other thinks
> 
> but if we agree on `common_variables` I'm also ok !

Ok, let's think about it. I'm fine changing the name. The purpose of this module is to store variables that are used in all of the modules reducing duplicated code

stefanotorresi:
I don't think `global_variables` would fit, tbh, because they're not actually global, they're scoped in a module. In fact, Terraform doesn't have the concept of global variables at all. :wink:

MalloZup:
@arbulu89  I actually cliked on red button but I wanted to make a comment. I'm full ok if people thinks `common_variables` are ok Is not really urgent.
@stefanotorresi  even if scoped to module they are a global module which used among all others

PR #513, opened 2020-07-21
Feature/ha infra gcp aws

arbulu89:
Create the HA infrastructure only if its enabled.

Besides that, I have removed the git push `hana_disk_device` variable, as I think it's better to use hardcoded values for these kind of parameters, that nobody changes.

### DIFF
--- a/gcp/modules/netweaver_node/main.tf
+++ b/gcp/modules/netweaver_node/main.tf
@@ -1,6 +1,10 @@
 # netweaver deployment in GCP
 # official documentation: https://cloud.google.com/solutions/sap/docs/netweaver-ha-planning-guide
 
+locals {
+  create_ha_infra = var.netweaver_count > 1 && var.ha_enabled ? 1 : 0
+}
+
 resource "google_compute_disk" "netweaver-software" {
   count = var.netweaver_count
   name  = "${terraform.workspace}-nw-installation-sw-${count.index}"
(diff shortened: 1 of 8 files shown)


gcp/modules/netweaver_node/main.tf
```diff
@@ -12,7 +16,7 @@ resource "google_compute_disk" "netweaver-software" {
 # Don't remove the routes! Even though the RA gcp-vpc-move-route creates them, if they are not created here, the terraform destroy cannot work as it will find new route names
 resource "google_compute_route" "nw-ascs-route" {
   name                   = "${terraform.workspace}-nw-ascs-route"
-  count                  = var.netweaver_count > 0 ? 1 : 0
+  count                  = local.create_ha_infra
```
Simranpal, line 19: We need to create the routes in the case of HA cluster scenario right, not in case of without HA (just netweaver)

arbulu89: No, actually, I'm doubting. Virtual addresses are created independently the scenario, HA or not HA. So, I don't know if need to create the routes if we have non HA scenario. I would say we don't need them, as they are used only by the cluster to redirect the data to the correct machine nothing else.

arbulu89: I would say that we need them only in HA. Otherwise, we would have routes as well for the PAS and AAS machines, what we don't have

PR #515, opened 2020-07-23
Provision Loki

stefanotorresi:
This PR adds a basic setup of Loki in the monitoring server and Promtail in all the target nodes.

It contextually renames the `monitoring` role and state to `monitoring_srv`, to make more clear the distinction between server and targets, with the intention to probably introduce a new role in the future called `monitoring_target` (currently all the related salt tasks are inside the `cluster_node` state).

There is no user-facing integration for the time being, other than the provisioning of a Grafana datasource, which can be browsed via the Grafana Explorer.

You can find a temporary example [here](http://10.162.32.253:3000/explore?orgId=1&left=%5B%22now-6h%22,%22now%22,%22Loki%22,%7B%22expr%22:%22%7Bjob%3D%5C%22systemd-journal%5C%22%7D%20%7C~%20%5C%22(%3Fi)error%7Cwarn%5C%22%22,%22hide%22:false,%22maxLines%22:null%7D,%7B%22mode%22:%22Logs%22%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D).


salt/cluster_node/templates/promtail.yaml.j2
```diff
@@ -0,0 +1,25 @@
+server:
+  http_listen_port: 9080
+  grpc_listen_port: 0
+
+positions:
+  filename: /tmp/promtail-positions.yaml
+
+clients:
+  - url: http://{{ grains['monitoring_srv_ip'] }}:3100/loki/api/v1/push
```
arbulu89, line 9: This entry is not available in all of the nodes as far as I see

libvirt/main.tf
```diff
@@ -48,6 +48,7 @@ module "common_variables" {
   provisioner            = var.provisioner
   background             = var.background
   monitoring_enabled     = var.monitoring_enabled
+  monitoring_srv_ip      = var.monitoring_enabled ? module.monitoring.output_data.private_address : ""
```
arbulu89, line 51: Why have you added this new parameter here? 
I think this will make fail the deployment. This variable is not even declared in the `variables.tf` file of `common_variables`.
Maybe you missed to commit some files?

salt/cluster_node/monitoring.sls
```diff
@@ -18,4 +18,35 @@ activate_node_exporter_systemd_collector:
...
+        attempts: 3
+        interval: 15
+
+promtail_config:
+  file.managed:
+    - name: /etc/loki/promtail.yaml
+    - template: jinja
+    - source: salt://cluster_node/templates/promtail.yaml.j2
+
+loki_systemd_journal_member:
```
MalloZup, line 36: maybe we can comment  on this why we need to do so.

I guess a comment stating that we want to parse systemd journal is enough.  Since isn't somehting we do normally by pkg install

salt/cluster_node/monitoring.sls
```diff
@@ -18,4 +18,35 @@ activate_node_exporter_systemd_collector:
     - name: /etc/sysconfig/prometheus-node_exporter
     - makedirs: True
     - contents: |
-        ARGS="--collector.systemd"
+        ARGS="--collector.systemd --no-collector.mdadm"
```
MalloZup, line 21: is there a particular reason we are adding 
```--no-collector.mdadm"```?

File: libvirt/main.tf (hunk 2 above)
Comment (stefanotorresi) [reply libvirt/main.tf:51]: hmm ahh yes, I missed the variable declaration, which was present in each module before rebasing against the `common_variables` merge.

File: salt/cluster_node/monitoring.sls (hunk 4 above)
Comment (stefanotorresi) [reply salt/cluster_node/monitoring.sls:21]: yes :grin: :grin::grin: :
Screenshot from 2020-07-23 16-23-52

File: salt/cluster_node/templates/promtail.yaml.j2 (hunk 1 above)
Comment (stefanotorresi) [reply salt/cluster_node/templates/promtail.yaml.j2:9]: it's in the common variables now :)

File: salt/cluster_node/monitoring.sls (hunk 3 above)
Comment (stefanotorresi) [reply salt/cluster_node/monitoring.sls:37]: done!

**Comment by stefanotorresi** [comment]
@MalloZup I don't know what to add at the moment, other than we are provisioning Loki along with Prometheus and Grafana. I'm open to suggestions.

**Comment by MalloZup** [comment]
@stefanotorresi to me we can just simply add 2 lines somewhere stating about loki.

Like we did for monitoting

we can add it on the main readme.md, `logging` index and put 2/3 lines.
like

```
Logging
Overview

The logging feature is complementary and to the monitoring feature.

Right now we install the required component needed for capturing logs file in each node
````

@stefanotorresi  It can be a placeholder content.. It is just to don't forget to create doc in future.

salt/cluster_node/monitoring.sls
MalloZup: ok thx

PR #516, opened 2020-07-23
Feature/create common variables

arbulu89:
I have extended the recently created `common_variables` usage to the other providers.

Besides that, I have included the `provider_type` (that is later added in the grains file) and the `public_key_location/private_key_location`. 

The use shouldn't notice any change here (except I have fixed some ssh keys usage, as they were not used properly)

I want to test them all, so, even if the PR is finished, I will create it as draft.


stefanotorresi:
@arbulu89 maybe defer until #515 is merged so that you can add the monitoring server IP grain across the board here.

PR #521, opened 2020-08-24
WIP: Active directory support V1

MalloZup:
# description:

add a V1 initial version of active directory.  As you can see the branch is ```SUSE:develop-active-directory``` so we will merge it later on `develop` branch once we get feedback.

Unfortunately,  I couldn't test the azure part


MalloZup:
image

MalloZup:
why this is closed?

stefanotorresi:
my mistake! sorry!

libvirt/variables.tf
```diff
@@ -47,6 +47,30 @@ variable "public_key_location" {
...
+  type        = string
+  default     = ""
+}
+
+variable "ad_server_domain" {
+  description = "name of the AD domain we want to join"
+  type        = string
+  default     = ""
+}
+
```
arbulu89, line 69: Remove all of these new lines please.

libvirt/variables.tf
```diff
@@ -47,6 +47,30 @@ variable "public_key_location" {
...
+# active directory server variables
+
+variable "ad_server" {
+  description = "IP adress of active directory server"
+  type        = string
+  default     = ""
+}
+
+variable "ad_adm_pwd" {
+  description = "password for admin of domain"
```
arbulu89, line 59: The rest of the descriptions are started with capital letter. This should follow the rule

arbulu89 (changes requested):
@MalloZup In general it looks a good implementation. I added some small comments. And besides them:

We will need to figure it out how to deal with different role nodes (hana, netweaver, monitoring). I would assume that someone who is using AD, would like to have all the machines associated to this AD server, but obviously, not all the users should belong to all machines. Do we know how to do so?

MalloZup:
I think if the admin wants to have granularity and fine acess control, 
for netweaver/etc, the solution can be that the AD admin create subdomain

`foo.test.hana`, `foo.test.netweaver` `foo.test.monitoring`. etc

i think we could create like a list of map in terraform where the user specify host/role to AD

in pseudo code
`hana: foo.test.hana`, `netweaver: foo.test.netweaver` etc
`hana_pdw` 
and by default if there is no map but a global domain, just join all machine to this

somehting in this direction

salt/active_directory/setup.sls
```diff
@@ -0,0 +1,114 @@
+install_sssd_packages:
+  pkg.latest:
+    - pkgs:
+      - realmd
+      - adcli
+      - sssd
+      - sssd-tools
+      - sssd-ad
+      - samba-client
+      - tcsh
```
MalloZup, line 10: we don't need this will remove

libvirt/variables.tf
MalloZup: ok

MalloZup:
```module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): [ERROR   ] Command '['groupmod', '-g', 2002, 'sapsys']' failed with return code: 10
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): [ERROR   ] stdout: groupmod: group 'sapsys' does not exist in /etc/group
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): [ERROR   ] retcode: 10
``` 
seems in netweaver we hardcoded groups with `groupmod`

stefanotorresi:
> seems in netweaver we hardcoded groups with `groupmod`

what salt state was this in?

MalloZup:
@stefanotorresi  it is an upstream call group.present which we call on netweaver formula.

However due the early stage of the `netweaver` AD implementation/integration I need to check if it is the culprit or is an `error` we have just because of that.

MalloZup:
Closing this since we will just offer some documentation which I'm working with thomas. is ok to close

PR #523, opened 2020-08-25
Increase cluster join timeout

arbulu89:
Our jenkins CI tests are failing many times due the joining cluster operations timeout. This happens because the 1st node needs more time than the 2nd, depending how the HANA installation is done (and mostly the HA enablement actions).

Increasing this timeout should fix the issues (most of the times it fails for 30 seconds or so, let's put more time just in case hehe)


MalloZup (approved):
sehr gut

PR #525, opened 2020-09-01
Run iscsi-init service to create initiatorname.iscsi file

arbulu89:
After open-iscsi has been upgraded to 2.1.2 version, the initiatorname.iscsi file is not shipped by default,and it's created with iscsi-init service.

More information in: https://build.opensuse.org/package/view_file/network/open-iscsi/open-iscsi.changes?expand=1

The new version has only been shipped to SLE15SP2, that's why the filter is set.

**PD: This change should be backported to master branch at some point if we don't plan a new release soon, as it has the same impact there**

### DIFF
--- a/salt/cluster_node/ha/iscsi_initiator.sls
+++ b/salt/cluster_node/ha/iscsi_initiator.sls
@@ -10,6 +10,15 @@ lsscsi:
       attempts: 3
       interval: 15
 
+{% if grains['osrelease_info'][0] > 15 or (grains['osrelease_info'][0] == 15 and grains['osrelease_info']|length > 1 and grains['osrelease_info'][1] >= 2) %}
+# We cannot use service.running as this systemd unit will stop after being executed
+# It is used only to create the initiatorname.iscsi file
+start_iscsi_init:
+  module.run:
+    - service.start:
+      - name: iscsi-init
+{% endif %}
+
 /etc/iscsi/initiatorname.iscsi:
   file.replace:
     - pattern: "^InitiatorName=.*"


salt/cluster_node/ha/iscsi_initiator.sls
```diff
@@ -10,6 +10,15 @@ lsscsi:
       attempts: 3
       interval: 15
 
+{% if grains['osrelease_info'][0] == 15 and grains['osrelease_info'][1] == 2 %}
```
stefanotorresi, line 13: what happens with future versions?

stefanotorresi (changes requested):
@arbulu89 I think we should target this change to master and then forward-port it to develop.

salt/cluster_node/ha/iscsi_initiator.sls
arbulu89: > @arbulu89 I think we should target this change to master and then forward-port it to develop.

The things is that it wouldn't be a direct port. The applied change is even in different files, so the commit must be different I guess.

arbulu89: > what happens with future versions?

Frankly, I don't even know what will happen in older versions... What I know now is that the new version only has been shipped in SLE15SP2, and not in the other versions.

MalloZup: an alternative approach would be

```
start_iscsi_init:
  module.run:
    - service.start:
      - name: iscsi-i
    - - check_cmd:
         '/usr/bin/true` (which always be true)

```
It might seems a bit ugly but on the other side we spare the `15sp3` and later checks maintenance.
```

Comment (arbulu89) [reply salt/cluster_node/ha/iscsi_initiator.sls:13]: @MalloZup I don't understand what that command would help on. If it's always true, what is the benefit of using it?

Comment (MalloZup) [reply salt/cluster_node/ha/iscsi_initiator.sls:13]: the benefits is that it will run always and execute, nevertheless the distro. 

it is like `mkdir -p`            no error if existing, make parent directories as needed

Comment (stefanotorresi) [reply salt/cluster_node/ha/iscsi_initiator.sls:13]: @MalloZup as you said, it's a bit ugly: it would hide actual failures of the service, and our future selves would have a harder time understanding what's that for.

File: salt/cluster_node/ha/iscsi_initiator.sls [hunk 2]
```diff
@@ -10,6 +10,15 @@ lsscsi:
       attempts: 3
       interval: 15
 
+{% if grains['osrelease_info'][0] == 15 and (grains['osrelease_info']|length > 1 and grains['osrelease_info'][1] >= 2) %}
```
stefanotorresi, line 13: We still need to cover for version above 15, right?

stefanotorresi (changes requested):

salt/cluster_node/ha/iscsi_initiator.sls
arbulu89: Do we really need that? SLE16 (if this is the next release) won't come in year.

arbulu89: Change pushed in any case

stefanotorresi: better safe than sorry :wink:

salt/cluster_node/ha/iscsi_initiator.sls
```diff
@@ -10,6 +10,15 @@ lsscsi:
       attempts: 3
       interval: 15
 
+{% if grains['osrelease_info'][0] >= 15 and (grains['osrelease_info']|length > 1 and grains['osrelease_info'][1] >= 2) %}
```
stefanotorresi, line 13: Sorry xabi, I hate to bother you further with this :grimacing: but the whole condition now skips SLE16 specifically, because the second segment of the version is always compared to 1.

The final version should be:
```suggestion
{% if grains['osrelease_info'][0] > 15 or (grains['osrelease_info'][0] == 15 and grains['osrelease_info']|length > 1 and grains['osrelease_info'][1] >= 2) %}
```

stefanotorresi (changes requested):

stefanotorresi (approved):
we made it! :smile:

PR #526, opened 2020-09-01
Handle case where srv/salt already exists

MalloZup:
In azure deployment with sle15sp2 image we have failure otherwise

### DIFF
--- a/salt/provision.sh
+++ b/salt/provision.sh
@@ -62,8 +62,13 @@ repeat_command () {
 }
 
 bootstrap_salt () {
-    mv /tmp/salt /srv || true
-    mv /tmp/pillar /srv || true
+    # handle case where something exists already in /srv/salt
+    mkdir -p /srv/salt
+    cp -R /tmp/salt/* /srv/salt || true
+    rm -rf /tmp/salt
+    mkdir -p /srv/pillar
+    cp -R /tmp/pillar/* /srv/pillar || true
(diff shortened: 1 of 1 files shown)


arbulu89 (changes requested):
@MalloZup I have checked this issue, and I have found out something quite interesting.
It looks like that not only salt is installed by default in `SLE15 SP2`, even our `salt-shaptools` and formulas are installed. That's why the `mv` operations files, becauese `_modules` and `_states` folders are already there in `/srv/salt`.

For this reason, it looks quite risky to remove `/srv/salt` folder, because it will remove folder from one of our packages.

I think we need something better

arbulu89:
Something like this should work: 
```
mkdir -p /srv/salt
    cp -R /tmp/salt/* /srv/salt || true
    rm -rf /tmp/salt
    mkdir -p /srv/pillar
    cp -R /tmp/pillar/* /srv/pillar || true
    rm -rf /tmp/pillar
```

MalloZup:
ok adapted. I will test the patch on azure
do we know why shaptools are already installed out of curiosity?

MalloZup:
ok last change suggested works on azure @arbulu89

arbulu89:
> ok adapted. I will test the patch on azure
> do we know why shaptools are already installed out of curiosity?

To have the monitoring packages running apparently

arbulu89 (approved):

Issue #528, opened 2020-09-07
post-run "hook" on terraform


MalloZup:
We should have something like this:

```
--- a/generic_modules/salt_provisioner/main.tf
+++ b/generic_modules/salt_provisioner/main.tf
```

```
 resource "null_resource" "provision_background" {
   count = var.background ? var.node_count : 0
   triggers = {
@@ -80,4 +82,11 @@ resource "null_resource" "provision" {
       "sudo sh /srv/salt/provision.sh -pdql /var/log/provisioning.log",
     ]
   }
+
+  provisioner "local-exec" {
+    command     = <<EOT
+      terraform show | grep tainted
+      ls -l
+    EOT
+  }
 }
```

This is on its primitive status.
I imagine  that we have a custom script, which can collect logs, print the needed errors and show which node is tainted.

I think a python script in this case is needed so we achieve portability on OS systems

MalloZup:
So i researched about this feature and It is not easy to add such hook if we don't have a cli. For the moment I think we can live without this. closing

PR #529, opened 2020-09-07
Improve logging and debugging in context of logs format

MalloZup:
# description

This PR:

1)  add `log_ok` and `log_error` helper function and adapt them in script. 

2) Print during execution only salt `error` and a resume of state at the end. 

3) create a `phase` logger a first v1 of logger, with colors which says the status of phase  with timestamp, hostname and msg

4) add back the colored output

We do have also a `debug` file for the full verbose mode in case needed .

# Screenshot:

image

image

### DIFF
--- a/salt/provision.sh
+++ b/salt/provision.sh
@@ -16,6 +16,18 @@ get_grain () {
     fi
 }
 
+log_ok () {
+  NODE=`hostname`
+  TIMESTAMP=`date -u`
+  GREEN='\033[0;32m'
+  RED='\033[0;31m'
+  NC='\033[0m' # No Color
+  printf "${GREEN}$TIMESTAMP::$NODE [INFO]  $1 ${NC}\n"
+}
+
+
+
+
 salt_output_colored () {
     if [[ "$(get_grain qa_mode)" == "true" ]]; then
         echo "--no-color"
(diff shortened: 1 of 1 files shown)


MalloZup:
I could do also https://github.com/SUSE/ha-sap-terraform-deployments/issues/528 on same PR but the `post-handling` script need to be done on other PRs but it is on the same context

salt/provision.sh
```diff
@@ -16,6 +16,18 @@ get_grain () {
...
+  TIMESTAMP=`date -u`
+  GREEN='\033[0;32m'
+  RED='\033[0;31m'
+  NC='\033[0m' # No Color
+  printf "${GREEN}$TIMESTAMP::$NODE [INFO]  $1 ${NC}\n"
+}
+
+
+
+
```
stefanotorresi, line 30: I think we can get rid of all these empty new lines :)

stefanotorresi (approved):
This looks like a very much needed change! GJ!

Only a those empty newlines to remove.

MalloZup:
@stefanotorresi  thx for review. I want to push during the day more  to screenshot and some other minimal changes so will ask you re-review. and remove the wip

salt/provision.sh
MalloZup: yop

stefanotorresi (approved):
GLORIOUS! :grin: :star_struck: :grin:

MalloZup:
thx @stefanotorresi .. :pray:  

@Simranpal @yandroskaos if from your perspective is ok? let me know your thoughts since it is a global change is always good to have different povs

yandroskaos (approved):
LGTM

PR #531, opened 2020-09-08
Make output quite for command who check only

MalloZup:
this PR fix `false` failure we print as error. Is better to use quiet so we don't print errors which aren't

### DIFF
--- a/salt/cluster_node/ha/iscsi_initiator.sls
+++ b/salt/cluster_node/ha/iscsi_initiator.sls
@@ -62,6 +62,7 @@ iscsid:
 wait_disk_id_available:
   cmd.run:
     - name: until [ "$(lsscsi -i | grep "LIO-ORG" | awk "{{ sbd_disk_id_pattern }}{print \$NF }")" != "-" ];do sleep 3;done
+    - output_loglevel: quiet
     - timeout: 120
     - require:
       - lsscsi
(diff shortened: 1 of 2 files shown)


MalloZup:
that is a valid question.

we can add `check_cmd` so we are fine.
https://docs.saltstack.com/en/latest/ref/states/requisites.html#check-cmd

MalloZup:
@arbulu89  so I tested locally a bit.
It is safe enough to add this option, since if there is an error on the state it will be marked as failed.
```
foo:
  cmd.run:
    - name: until brakezd ;do sleep 1;done
    - output_loglevel: quiet
    - timeout: 10
```              


```
# bash provision.sh -p
[ERROR   ] {'stdout': 'until brakezd ;do sleep 1;done : Timed out after 10 seconds', 'stderr': '', 'retcode': 1, 'pid': 26011}
local:
----------
          ID: foo
    Function: cmd.run
        Name: until brakezd ;do sleep 1;done
      Result: False
     Comment: Command "until brakezd ;do sleep 1;done" run
     Started: 15:25:53.343380
    Duration: 10008.521 ms
     Changes:   
              ----------
              pid:
                  26011
              retcode:
                  1
              stderr:
              stdout:
                  until brakezd ;do sleep 1;done : Timed out after 10 seconds

Summary for local
------------
Succeeded: 0 (changed=1)
Failed:    1
------------
Total states run:     1
Total run time:  10.009 s
```

arbulu89 (approved):
Thanks for checking @MalloZup

PR #533, opened 2020-09-09
WIP: Failhard feature

MalloZup:
# Description
this PR add `failhard` global option enabled by default to true.

What does this feature? IN a cloud native environment, failing fast is saving money.

So we exit the salt process at **first failure**. Note: we don't do this right now, and the behavious is we are continuing executing states, which is quite confusing when users have issue/debug.

I think we don't need to add yet another variable for disabling this.

Thoughts?

The only possible alternative, would be that we have a configurable variable which can set this to false, if user want default .

But I personally think, in sake for minimalism and variable overflow we shouldn't add it now. If we have user request  we can still `expose` that variable later

More info here: https://docs.saltstack.com/en/latest/ref/states/failhard.html

### DIFF
--- a/salt/os_setup/minion_configuration.sls
+++ b/salt/os_setup/minion_configuration.sls
@@ -7,6 +7,7 @@ configure_file_roots:
   file.append:
     - name: /etc/salt/minion
     - text: |
+        failhard: True
         file_roots:
           base:
             - /srv/salt


arbulu89:
@MalloZup Some comments:
- I think we should identify and define the states that are really critical and set the failhard there. If in some of our formulas, we have states that if they fail, everything else will fail, we might need to set them with the failhard option.
- This change is quite aggressive, I don't know if we want to stop the execution if any single state fails. This needs more research.
- I think that we don't need to think about `cloud native` environments. Users might need to have the environments running for days/weeks/months. I don't think having them running for 1 hour more is a big deal.
- Making it optional as a variable might be a better option.

MalloZup:
ok fair enough

arbulu89:
@MalloZup I have found more reasons why having a generic `failhard` is dangerous.
For example, now we have implemented a `lock` mechanism in `habootstrap-formula`: https://github.com/SUSE/habootstrap-formula/pull/56

With this change, the formula will acquire a lock before the joining process and release after that. If for some reason, the join process fails, with the `failhard`, the whole execution would be stopped, without running the release of the lock.

With this, what I mean that we need to really conscious about this feature. I wouldn't implement it in any case, as we might have in the future some `cleaning up` states that must be executed even in the failure of other processes.

MalloZup:
@arbulu89 ok make sense. I think we can still implement it for individual states

arbulu89:
> @arbulu89 ok make sense. I think we can still implement it for individual states

Yes, and it would be really needed!

MalloZup:
closing as stale issue. I will not have time to do this

PR #534, opened 2020-09-10
Hotfix/aws fixes

arbulu89:
2 fixes for AWS deployment:
- Remove quotes from aws command. They are causing an error (even though it was not happening before...). I don't really know the route cause, but having the quotes didn't make much sense neither, as it was changing the bash call
- Update folder permissions to the hana software folder. It might cause errors depending on the used HANA version

### DIFF
--- a/salt/hana_node/download_hana_inst.sls
+++ b/salt/hana_node/download_hana_inst.sls
@@ -50,7 +50,7 @@ hana_inst_directory:
   file.directory:
     - user: root
     - group: root
-    - dir_mode: "0754"
+    - dir_mode: "0755"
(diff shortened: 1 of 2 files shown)


MalloZup (approved):
thx!

would be a `failhard: True` command afaik :grin:

ab-mohamed:
Thanks, @arbulu89,
Is the hotfix that will be included in version 5?

arbulu89:
> Thanks, @arbulu89,
> Is the hotfix that will be included in version 5?

Hi @ab-mohamed ,
No, it will be released in `v6` that will come this week

PR #535, opened 2020-09-10
Add provisioning_log_level variable

arbulu89:
`provisioning_log_level` variable added to the deployment. Now, we can decide from a user perspective which log level we want to use in our deployments.

I have put `error` by default. The old output was `info`, and we log as `debug` in the log files.

PD: I have indented some lines in `provision.sh` too


MalloZup (approved):
thx ok

PR #536, opened 2020-09-11
Export logs to log file

MalloZup:
so this PR fix the case where we were not printing to logs to files

### DIFF
--- a/salt/provision.sh
+++ b/salt/provision.sh
@@ -116,7 +116,6 @@ bootstrap_salt () {
 
 os_setup () {
     LOG_FILE=$1
-
     # Execute the states within /srv/salt/os_setup
     # This first execution is done to configure the salt minion and install the iscsi formula
     salt-call --local \
@@ -125,7 +124,7 @@ os_setup () {
         --log-file-level=debug \
         --retcode-passthrough \
         $(salt_output_colored) \
-        state.apply os_setup | tee $LOG_FILE || log_error "os setup failed"
+        state.apply os_setup | tee -a $LOG_FILE || log_error "os setup failed"
     log_ok "os setup done"
 }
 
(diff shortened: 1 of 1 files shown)


MalloZup:
currently testing it

stefanotorresi (changes requested):
@MalloZup  instead of passing the `$LOG_FILE` variable around into the functions, maybe we could append `| tee -a $log_to_file || log_error "foo"` to each function invocation? This would result in less jumping around the code and all the logging things being in one place.

MalloZup:
@stefanotorresi  can you be more explicit? I didn't understand what you mean tia

salt/provision.sh
```diff
@@ -225,20 +229,20 @@ done
...
 fi
 
+
 if [ $argument_number -eq 0 ]; then
-    bootstrap_salt
-    os_setup
-    predeploy
-    deploy
-    run_tests
+    bootstrap_salt $log_to_file
```
MalloZup, line 236: @stefanotorresi 

`  bootstrap_salt  | tee   $log_to_file || log_error "tests failed"`

MalloZup: I will check if this work in bash

stefanotorresi:
also, since we are now catching errors from a pipe with `tee`, we need to add the `-o pipefail` option to the script interpreter, so the shebang would become `#!/bin/bash -xeo pipefail` .

salt/provision.sh
stefanotorresi: yes, exactly, so no logging happens inside the `boostrap_salt` function, all of it happens in the "main" of the script.

MalloZup:
I'm ok to close this in favor of @arbulu89  PR. Allthough I find this PR somehow less `bashism` and more understanble , I think for the `provision` script we can live with the other pr if we have comment why we put that command.

PR #537, opened 2020-09-11
Fix salt result output logging issues

arbulu89:
Log properly the salt execution result summary to a file (besides the standard output)

Besides that, I have changed the `provisioning.log` file name to `salt-result.log`.

**PD: Now the `provision.sh` is executed with `bash` what makes sense as actually it was defined in the shebang. I don't really know why we put `sh` honestly...**


MalloZup (changes requested):
looks good only 2 things:

-` #!/bin/bash -xeo pipefail` as @stefanotorresi 
- comment the command 

a part of that ok is more minimalistic then mine :robot:

arbulu89:
@MalloZup @stefanotorresi Changes applied

MalloZup (approved):

MalloZup:
thx!

PR #539, opened 2020-09-18
Correct drbd pillar file to use the splitbrain handler correctly

arbulu89:
Fix drbd pillar file to use the splitbrain handler properly


nick-wang (approved):
Looks good to me. It should be found eariler...

nick-wang (approved):
Now I am pretty sure this function didn't use after merged.
handlers not working == no split brain happen ...

PR #545, opened 2020-09-22
update readme with more tfvars.example reference

MalloZup:
this will make the tfvars.example more visible as entry point for features details

https://github.com/MalloZup/ha-sap-terraform-deployments/tree/doc-link#major-features

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ ___
 
 ## Major features
 
+For details, examples about features, read the `terraform.tfvars.example` of each provider.
+
 - HA Clusters and HANA deployment
 - [Monitoring of cluster](doc/monitoring.md)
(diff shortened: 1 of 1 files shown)


arbulu89 (changes requested):
@MalloZup I agree that we could add something like this in the main README file. But I don't think this is the best place or text.

We already have such a link for each of the providers: https://github.com/SUSE/ha-sap-terraform-deployments/tree/master/azure#specifications

I would just simply improve the https://github.com/SUSE/ha-sap-terraform-deployments/tree/master/azure#specifications entry to say that you have to edit the `terraform.tfvars` and you have some example here

MalloZup:
ok let close this then for moment

Issue #542, opened 2020-09-21
DOC Add example for `zip/rar/exe/` media file


MalloZup:
Add support for zip,rar,exe,sar HANA media files. Q: there are no good examples for the usage.

@Simranpal cc @arbulu89

arbulu89:
We have this in the `terraform.tfvars.example`.

```
# Or specify the path to the HANA installation archive file in either of SAR, RAR, ZIP, EXE formats, relative to the 'hana_inst_media' mounting point
# For multipart RAR archives, you only need to provide the first part EXE file name.
# If using hana sar archive, please also provide the compatible version of sapcar executable to extract the sar archive
# HANA installation archives be extracted to path specified at hana_extract_dir (optional, by default /sapmedia/HANA)
hana_archive_file = "IMDB_SERVER.SAR"
hana_sapcar_exe = "SAPCAR"
hana_extract_dir = "/sapmedia/HDBSERVER"
```

Is not enough?

MalloZup:
this issue was reported by Peter schnagel. Ido think maybe we could add this reference to release note.

petersatsuse:
regarding place of the SAP media: If a customer downloads e.g. the HANA files from SAP he will get this:
51053381_part1.exe
51053381_part2.rar
51053381_part3.rar
51053381_part4.rar

The examples show a) the fully unpacked directories b) something with a number - which is not clear how to use c) with a .sar archive and sapcar, and a extract dir - which does also not fit, as sapcar is normally not needed for HANA

arbulu89:
@petersatsuse I will update the description, but remember that this is just an example. I'm afraid that we cannot cover all the use cases (and the `SAR` option with `SAPCAR` executable is one of them, even though you say that doesn't fit. Other users have requested this specifically).

petersatsuse:
So how to use the above showed downloads, as this is what a customer downloads from SAP. With all other ways the customer had done something with downloaded files...or is using some former extracted parts

petersatsuse:
1)
# Azure storage account path where HANA installation master is located
# hana_inst_master 

This mean to me is a path pointing to the *extracted* archive files 

2) 
# Or you can combine the `hana_inst_master with` `hana_platform_folder` variable.

This mean to me, i use the path above and in addition point to a directory

3) # Or specify the path to the HANA installation archive file in either of SAR, RAR, ZIP, EXE formats, relative to the 'hana_inst_media' mounting point.
[...]

PR #546, opened 2020-09-22
Improve HANA installation description in the examples

arbulu89:
An improvement for the documentation where we explain how to setup the terraform configuration file to use properly the HANA installation files.

In fact, I have noticed that we use a different variable name for each of the providers....
We should unify it. Should we use `hana_inst_media` ?

FYI: @petersatsuse
#542


petersatsuse:
looks much clearer as before to me

azure/terraform.tfvars.example
```diff
@@ -92,27 +92,44 @@ public_key_location = "/path/to/your/public/ssh/key"
...
+
+# 2. Combine the `hana_inst_master` with `hana_platform_folder` variable.
+#hana_inst_master = "//YOUR_STORAGE_ACCOUNT_NAME.file.core.windows.net/sapdata/sap_inst_media"
 # Specify the path to already extracted HANA platform installation media, relative to hana_inst_master mounting point.
 # This will have preference over hana archive installation media
 #hana_platform_folder = "51053381"
 
-# Or specify the path to the HANA installation archive file in either of SAR, RAR, ZIP, EXE formats, relative to the 'hana_inst_media' mounting point
+# 3. Specify the path to the HANA installation archive file in either of SAR, RAR, ZIP, EXE formats, relative to the 'hana_inst_master' mounting point
 # For multipart RAR archives, you only need to provide the first part EXE file name.
```
MalloZup, line 122: `you only` we could remove it

azure/terraform.tfvars.example
```diff
@@ -92,27 +92,44 @@ public_key_location = "/path/to/your/public/ssh/key"
...
-# If using hana sar archive, please also provide the compatible version of sapcar executable to extract the sar archive
+#hana_archive_file = "51053381_part1.exe"
+
+# 4. If using HANA SAR archive, please also provide the compatible version of sapcar executable to extract the SAR archive
```
MalloZup, line 125: remove `please` .
we can use imperative way for docs

MalloZup (approved):
looks good to me. I would remove the `you` and `please` and using imperative impersonal doc

libvirt/terraform.tfvars.example
```diff
@@ -19,20 +19,41 @@ iprange = "192.168.XXX.Y/24"
...
+#hana_archive_file = "IMDB_SERVER.SAR"
+#hana_sapcar_exe = "SAPCAR"
+
+# For option 3 and 4, HANA installation archives are extracted to the path specified
+# at hana_extract_dir (optional, by default /sapmedia/HANA)
+#hana_extract_dir = "/sapmedia/HDBSERVER"
+
+
+# More configuration about the HANA machines
+# Set the IP addresses for the HANA machines. Let this commented to get an autogenerated addresses
```
Simranpal, line 56: ` Let this commented / Leave this commented` to get autogenerated addresses

Simranpal (approved):
Thanks @arbulu89 for the improvements. Yes, I agree we can unify and use  `hana_inst_media` for all providers

petersatsuse:
I did yesterday additional checks as i got errors during the deployment, so i looked at the usage of the variables at all.

a) If you look at the past where the "install_master" was a DVD, it's clear that this can for sure not be the mountpoint to extract the media. 
b) Looking into the code, there is *always" hana_extract_dir used. 

The hana_extract_dir should *never* point to the _"readonly DVD"_ mountpoint of "install_master", instead there should be a normal **writeable** mountpoint and disk be used.

I have mostly on baremetal installs seen that /hana/data/<something> or /hana/backup/<something> is been used

So the result would be for example:

```
 # readonly mount point of the SAP media
  hana_inst_master = "sapfilestore.file.core.windows.net/sapbits/HANA"
 
 # HANA downloaded archive 
  hana_archive_file="51053381_part1.exe"
 
 # Extracted HANA media
  hana_extract_dir = "/hana/shared/HDBSERVER"
```

arbulu89:
> I did yesterday additional checks as i got errors during the deployment, so i looked at the usage of the variables at all.
> 
> a) If you look at the past where the "install_master" was a DVD, it's clear that this can for sure not be the mountpoint to extract the media.
> b) Looking into the code, there is *always" hana_extract_dir used.
> 
> The hana_extract_dir should _never_ point to the _"readonly DVD"_ mountpoint of "install_master", instead there should be a normal **writeable** mountpoint and disk be used.
> 
> I have mostly on baremetal installs seen that /hana/data/ or /hana/backup/ is been used
> 
> So the result would be for example:
> 
> ```
>  # readonly mount point of the SAP media
>   hana_inst_master = "sapfilestore.file.core.windows.net/sapbits/HANA"
>  
>  # HANA downloaded archive 
>   hana_archive_file="51053381_part1.exe"
>  
>  # Extracted HANA media
>   hana_extract_dir = "/hana/shared/HDBSERVER"
> ```

Fantastic @petersatsuse ! Thank you for the research.
So, we will need to change the default `hana_extract_dir` value to something like you exposed. We will do the changes and test them in Azure.
This will probably come in a new PR though, to keep the things in order.
I will update the `terraform.tfvars.example` to show your example.

PR #551, opened 2020-09-28
Feature/redo os image usage

arbulu89:
Some new implementations to the mechanism used to select the OS image for the machines.

It brings:

- Use PAYG images by default. We receive some issues because the users use default values (BYOS) and they don't have (or don't use) the registration code. For newcomers (or the most standard user), PAYG are better.
- Create new os_image variable as default OS image. Right now, we need to set the OS image for all of the nodes (hana, iscsi, etc). As most of the users will use the same OS for all the machines, having this option makes things easier, as you only need to change 1 variable. Specific customization is still possible of course.
- Unify naming. Now, all the variables for all providers follow the same syntax (`*_os_image).
- Change how Azure images are selected. I think that is easier to only have unique variable, instead of 4, that can be obtained from CLI commands output
- Reformat the terraform.tfvars.example to be more informative (I think I will split this file in terraform.tfvars.example.basic and terraform.tfvars.example.advanced in the future)

FYI @peteratsuse

### DIFF
--- a/aws/modules/monitoring/variables.tf
+++ b/aws/modules/monitoring/variables.tf
@@ -11,7 +11,7 @@ variable "monitoring_enabled" {
 variable "instance_type" {
   description = "The instance type of monitoring node."
   type        = string
-  default     = "t2.micro"
+  default     = "t3.micro"
 }
 
 variable "key_name" {
--- a/aws/terraform.tfvars.example
+++ b/aws/terraform.tfvars.example
@@ -28,7 +28,7 @@ aws_secret_access_key = my-secret-access-key
 # aws-cli credentials file. Located on ~/.aws/credentials on Linux, MacOS or Unix or at C:\Users\USERNAME\.aws\credentials on Windows
 aws_credentials = "~/.aws/credentials"
 
-# If BYOS images are used in the deployment, SCC registration code is required
+# If BYOS images are used in the deployment, SCC registration code is required. Set `reg_code` and `reg_email` variables below
 # By default, all the images are PAYG, so these next parameters are not needed
 #reg_code = "<<REG_CODE>>"
 #reg_email = "<<your email>>"
@@ -40,7 +40,7 @@ aws_credentials = "~/.aws/credentials"
 #    "sle-ha-geo/12.4/x86_64" = "<<REG_CODE>>"
 #}
 
-# Default os_image and os_owner. These values are not used if the specific values are not set (e.g.: hana_os_image)
+# Default os_image and os_owner. These values are not used if the specific values are set (e.g.: hana_os_image)
 # BYOS example with sles4sap 15 sp1 (this value is a pattern, it will select the latest version that matches this name)
 #os_image = "suse-sles-sap-15-sp1-byos"
 #os_owner = "amazon"
@@ -76,7 +76,7 @@ cluster_ssh_key = "salt://sshkeys/cluster.id_rsa"
 # Provisioning log level (error by default)
 #provisioning_log_level = "info"
 
-# Enable all some pre deployment steps (disabled by default)
+# Enable pre deployment steps (disabled by default)
 pre_deployment = true
 
 # To disable the provisioning process
@@ -87,7 +87,7 @@ pre_deployment = true
 
 # QA variables
 
-# Define if the deployment is using for testing purpose
+# Define if the deployment is used for testing purpose
 # Disable all extra packages that do not come from the image
 # Except salt-minion (for the moment) and salt formulas
 # true or false (default)
@@ -181,15 +181,11 @@ hana_inst_master = "s3://sapdata/sap_inst_media/51053381"
 # SBD related variables
 #######################
 
-# In order to enable SBD, an ISCSI server is needed as right now is the unique option
+# In order to enable SBD, an ISCSI server is needed as right now is the only option
 # All the clusters will use the same mechanism
 # In order to enable the iscsi machine creation sbd_enabled must be set to true for any of the clusters
 
-# iSCSI OS image
-#iscsi_os_image = "ami-xxxxxxxxxxxxxxxxx"
-#iscsi_os_owner = "self"
-
-# iSCSI server image. By default, PAYG image is used. The usage is the same than the HANA images
+# iSCSI server image. By default, PAYG image is used. The usage is the same as the HANA images
 #iscsi_os_image = "suse-sles-sap-15-sp1-byos"
 #iscsi_os_owner = "amazon"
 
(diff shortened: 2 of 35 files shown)


azure/modules/drbd_node/main.tf
```diff
@@ -187,10 +192,10 @@ resource "azurerm_virtual_machine" "drbd" {
...
   storage_image_reference {
     id        = var.drbd_image_uri != "" ? join(",", azurerm_image.drbd-image.*.id) : ""
-    publisher = var.drbd_image_uri != "" ? "" : var.drbd_public_publisher
-    offer     = var.drbd_image_uri != "" ? "" : var.drbd_public_offer
-    sku       = var.drbd_image_uri != "" ? "" : var.drbd_public_sku
-    version   = var.drbd_image_uri != "" ? "" : var.drbd_public_version
+    publisher = var.drbd_image_uri != "" ? "" : module.os_image.publisher
+    offer     = var.drbd_image_uri != "" ? "" : module.os_image.offer
+    sku       = var.drbd_image_uri != "" ? "" : module.os_image.sku
+    version   = var.drbd_image_uri != "" ? "" : module.os_image.version
```
Simranpal, line 198: Do these properties of `os_image` need to be defined as well in the terraform tfvars? If so should we provide an example there?

azure/modules/drbd_node/variables.tf
```diff
@@ -45,24 +45,9 @@ variable "drbd_image_uri" {
...
-variable "drbd_public_sku" {
-  type    = string
-  default = "15"
-}
-
-variable "drbd_public_version" {
-  type    = string
-  default = "latest"
+variable "os_image" {
+  description = "sles4sap image used to create the this module machines. Composed by 'Publisher:Offer:Sku:Version' syntax. Example: SUSE:sles-sap-15-sp2:gen2:latest"
```
Simranpal, line 49: create the this /  create this

azure/modules/hana_node/main.tf
```diff
@@ -202,6 +202,11 @@ resource "azurerm_image" "sles4sap" {
 
 # hana instances
 
+module "os_image" {
+  source   = "../../modules/os_image_reference"
```
Simranpal, line 206: I guess you can use "`../os_image_reference`"

azure/modules/drbd_node/variables.tf
Simranpal: same for other places defining variable "os_image"

azure/modules/drbd_node/main.tf
Simranpal: I can see the properties are defined in os_image module, but should this be taken as input from user

gcp/terraform.tfvars.example
```diff
@@ -101,79 +170,51 @@ sap_hana_deployment_bucket = "MyHanaBucket/sapdata/sap_inst_media/51053381"
...
 
-# To disable the provisioning process
-#provisioner = ""
+#######################
+# SBD related variables
+#######################
 
-# Run provisioner execution in background
-#background = true
+# In order to enable SBD, an ISCSI server is needed as right now is the unique option
```
Simranpal, line 183: `is needed as right now it is the only option`

Simranpal (changes requested):
@arbulu89 I have doubt about the `os_image` if we should add the example usage  of setting sku, version etc. Other than that some typos and corrections in variable description and terraform.tfvars.example. I made the suggestions in `aws/terraform.tfvars.example` but it should apply to other modules

MalloZup (approved):
+1

aws/modules/monitoring/variables.tf
```diff
@@ -11,7 +11,7 @@ variable "monitoring_enabled" {
 variable "instance_type" {
   description = "The instance type of monitoring node."
   type        = string
-  default     = "t2.micro"
+  default     = "t3.micro"
```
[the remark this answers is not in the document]
arbulu89: There is not any documentation/suggestion. I changed this just after trial and error.
Depending on the nodes (`iscsi`), if we use `micro/small` types, the `xvdd` disk is not mounted. So we need to use large.
For the `monitoring` it doesn't really matter, and I think the cost is the same

aws/terraform.tfvars.example
```diff
@@ -17,38 +17,110 @@ aws_region = "eu-central-1"
...
+#reg_email = "<<your email>>"
+
+# To add additional modules from SCC. None of them is needed by default
+#reg_additional_modules = {
+#    "sle-module-adv-systems-management/12/x86_64" = ""
+#    "sle-module-containers/12/x86_64" = ""
+#    "sle-ha-geo/12.4/x86_64" = "<<REG_CODE>>"
+#}
+
+# Default os_image and os_owner. These values are not used if the specific values are not set (e.g.: hana_os_image)
```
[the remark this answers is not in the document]
arbulu89: Yes! Thanks for having a deep look XD

aws/terraform.tfvars.example
```diff
@@ -17,38 +17,110 @@ aws_region = "eu-central-1"
...
+# To auto detect the SLE version
+#ha_sap_deployment_repo = "http://download.opensuse.org/repositories/network:/ha-clustering:/Factory/"
+# Specific SLE version used in all the created machines
+#ha_sap_deployment_repo = "http://download.opensuse.org/repositories/network:/ha-clustering:/Factory/SLE_15/"
+#ha_sap_deployment_repo = ""
+
+# Provisioning log level (error by default)
+#provisioning_log_level = "info"
+
+# Enable all some pre deployment steps (disabled by default)
```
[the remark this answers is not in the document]
arbulu89: Yes, this needs a correction. I just copy/pasted without rewriting

azure/modules/drbd_node/main.tf
arbulu89: No. I changed the concept. Before, we asked the user for these values individually, and we had 4 variables. Now, I changed to a unique variable (`os_image`) that is composed by all the values separated by `:`.
Basically, the `os_image` module takes `os_image = "SUSE:sles-sap-15-sp2:gen2:latest"` and converts to `publisher = "SUSE"`, `offer = "sles-sap-15-sp2"`, etc.
I did this to reduce the amount of changes required to update a sles version

azure/modules/drbd_node/variables.tf
arbulu89: Thanks, I will have a look

aws/terraform.tfvars.example
```diff
@@ -17,38 +17,110 @@ aws_region = "eu-central-1"
...
+
+# QA variables
+
+# Define if the deployment is using for testing purpose
+# Disable all extra packages that do not come from the image
+# Except salt-minion (for the moment) and salt formulas
+# true or false (default)
+#qa_mode = false
+
+# Execute HANA Hardware Configuration Check Tool to bench filesystems
```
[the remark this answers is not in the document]
arbulu89: No idea. This was set by the QA team. In any case, most probably we will remove the `qa_mode` in the future, so I wouldn't be so worried

arbulu89:
@Simranpal Thank you so much for your nice review! I have addressed all the comments and applied your suggestions.
Have a new review when you have some time.

Simranpal (approved):
Thanks for the changes @arbulu89 looks good!

PR #552, opened 2020-09-29
Make iscsi initiator state idempotent

arbulu89:
The re-provisioning of `predeployment` fails if we use `iscsi` as the `iscsiadm` command already finds an existing session.
This `unless` command will check this, and ignore the state if a session is on place


MalloZup (commented):
thx

Simranpal (approved):
looks good!

Issue #548, opened 2020-09-24
hard coded values like /dev/sdc break install sometimes


petersatsuse:
in many places the hard coded disk device names are replaced by the respective dynamic look for the attached device to get the real devicename. But this is not true everywhere.

The need to use persistent device naming should be nothing new as the kernel device name like /dev/sda etc could be change on the fly.

It hit me for the azure drbd disk, as without a working drbd, there is no nfs and without nfs there is the netweaver install failing
but its still there on more places

PR #554, opened 2020-09-30
Use disk name instead of device in azure

arbulu89:
Fix for https://github.com/SUSE/ha-sap-terraform-deployments/issues/548

Azure always use the same disk names, so the name for the storage disk 1st lun always is: `/dev/disk/azure/scsi1/lun0`

Some info here:https://docs.microsoft.com/en-us/azure/virtual-machines/troubleshooting/troubleshoot-device-names-problems
I guess there is more information in Azure docks, but I was not able to find it.

FYI: @peteratsuse

### DIFF
--- a/azure/modules/drbd_node/salt_provisioner.tf
+++ b/azure/modules/drbd_node/salt_provisioner.tf
@@ -27,7 +27,7 @@ host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]
 host_ip: ${element(var.host_ips, count.index)}
 cluster_ssh_pub:  ${var.cluster_ssh_pub}
 cluster_ssh_key: ${var.cluster_ssh_key}
-drbd_disk_device: /dev/sdc
+drbd_disk_device: /dev/disk/azure/scsi1/lun0
 drbd_cluster_vip: ${var.drbd_cluster_vip}
 sbd_enabled: ${var.sbd_enabled}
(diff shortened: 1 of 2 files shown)


Simranpal (commented):
https://github.com/MicrosoftDocs/azure-docs/issues/13958

arbulu89:
> [MicrosoftDocs/azure-docs#13958](https://github.com/MicrosoftDocs/azure-docs/issues/13958)

@Simranpal Honestly, I don't know if what they say in this doc is true...
The disk name is based in `udev` rules, and the booting process, that might be different every boot. Actually, we have seen times, when the data disk is mounted in `/dev/sda` and the OS disk `/dev/sdc`. That's why @peteratsuse opened the issue, because time to time, the deployment fails when this scenario happens.
I think we should ignore the comments on that link.

arbulu89:
@peteratsuse I'm merging this. I have tested and I have not seen the DRBD issue anymore

petersatsuse:
> @peteratsuse I'm merging this. I have tested and I have not seen the DRBD issue anymore

Yes, i also did not see it anymore in my deployments. Thx

arbulu89:
@Simranpal @MalloZup Could any of you approve this change?

Simranpal (approved):
Thanks!

MalloZup (approved):

PR #556, opened 2020-10-06
Update the ssh command to connect to hana ips

Simranpal:
fixed typo to now use the hana instance ip instead of bastion host ip


Simranpal:
closing this as it should go into Develop branch first

PR #557, opened 2020-10-06
Update the ssh command to connect to hana ips

Simranpal:
fixed typo to now use the hana instance ip instead of bastion host ip


MalloZup (approved):

arbulu89 (approved):

arbulu89:
@Simranpal By the way, I have noticed that you are creating branches in the main fork, and not in yours. Please, don't do that.
For this time it doesn't matter. Remove the branch when the code is merged.

petersatsuse:
Sorry for this late note.
i think the command itself is a way too complicated, as the actual ssh command provide a flag "-J" which deals with the jumphost.
example: ssh -J <user@jumphost> <user@targethost>
So this is much simpler to read and understand, even if the actual cmd is not wrong.

arbulu89:
> Sorry for this late note.
> i think the command itself is a way too complicated, as the actual ssh command provide a flag "-J" which deals with the jumphost.
> example: ssh -J [user@jumphost](mailto:user@jumphost) [user@targethost](mailto:user@targethost)
> So this is much simpler to read and understand, even if the actual cmd is not wrong.

@Simranpal Could you test first this command? And update the docs if it works. Thanks

MalloZup:
thx Didn't know of that `-J`

Simranpal:
I tested the command but it did not work `ssh -J user@jumphost user@targethost` for me
I tried providing the SSH Key and other options but get this same error below

```
cloudadmin@51.141.174.162: Permission denied (publickey).
ssh_exchange_identification: Connection closed by remote host
```
@petersatsuse  Do I need to specify other options?

@MalloZup https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Proxies_and_Jump_Hosts#Jump_Hosts_--_Passing_Through_a_Gateway_or_Two

petersatsuse:
> I tested the command but it did not work `ssh -J user@jumphost user@targethost` for me
> I tried providing the SSH Key and other options but get this same error below
> 
> ```
> cloudadmin@51.141.174.162: Permission denied (publickey).
> ssh_exchange_identification: Connection closed by remote host
> ```
> 
> @petersatsuse Do I need to specify other options?
> 
> @MalloZup https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Proxies_and_Jump_Hosts#Jump_Hosts_--_Passing_Through_a_Gateway_or_Two

no. but maybe you need still the -i if they keys are not in the default path, as the -J only does the part of the proxy and -W

Simranpal:
@petersatsuse thanks, I tried with the `-i ` path to keys, for the bastion host, but got the same error, maybe we missing some other parameter

arbulu89:
@Simranpal @petersatsuse Keep in mind that we might have 2 ssh key pairs. One to connect to the bastion, and other to connecto the nodes. I think we need to provide both (this is what we do in the original command at least). Both key pairs being the same can be just a coincidence.

Simranpal:
Thanks @arbulu89 so the command with `-J` option does not work with key location other than default one you mentioned. 
@petersatsuse which explains the error I was getting  with ssh keys, due with key location `-i /home/simran/.ssh/id_rsa_cloud`
I will merge the PR with existing command as by default we use the custom key location.

PR #559, opened 2020-10-07
Replace how fencing mechanism is managed

arbulu89:
Replace how fencing mechanism is managed.
Before this change, we had the option to disable fencing mechanism completely.
We agreed that this is risky, and we shouldn't provide such a configuration.

This way, instead of enabling/disabling sbd, we give the option to select between `sbd` or `native` fence mechanism (in the CSPs that provide this option, `aws` and `gcp` by now. We still need to implement the Azure fence agent usage).

Now we have the next options:
`hana_cluster_fencing_mechanism = "sbd"`
or
`hana_cluster_fencing_mechanims = "native"`

The same for DRBD and Netweaver.
For Azure and Libvirt, the option is implemented, but it's not written in the `terraform.tfvars.example` as there are not more options.

It depends on the next PRs to work:
https://github.com/SUSE/saphanabootstrap-formula/pull/106
https://github.com/SUSE/sapnwbootstrap-formula/pull/65

FYI: @petersatsuse


MalloZup (approved):
looks good to me.

Only 2 points:

I find the variable `native` fencing somehow deviating from SBD documentation so I personally already have some difficulties what it means.

the 2nd point is, how we document this variable.

If `native` means like the default the cloud provider is offering etc

aws/terraform.tfvars.example
```diff
@@ -183,7 +183,7 @@ hana_inst_master = "s3://sapdata/sap_inst_media/51053381"
 
 # In order to enable SBD, an ISCSI server is needed as right now is the only option
 # All the clusters will use the same mechanism
-# In order to enable the iscsi machine creation sbd_enabled must be set to true for any of the clusters
+# In order to enable the iscsi machine creation _fencing_mechanism must be set to 'sbd' for any of the clusters
```
Simranpal, line 186: ..`_fencing_mechanism` / `fencing_mechanism` parameter must be set

aws/terraform.tfvars.example
```diff
@@ -264,8 +267,8 @@ hana_inst_master = "s3://sapdata/sap_inst_media/51053381"
 # Fon non HA
 #netweaver_product_id = NW750.HDB.ABAP
 
-# Enable SBD for the netweaver cluster
-#netweaver_cluster_sbd_enabled = true
+# Select Netweaver cluster fencing mechanism. 'native' by default
+#netweaver_cluster_fencing_mechanism = "sbd"
```
Simranpal, line 271: for all of these variables commented out, should this be set to `native`, as the line above says.. 
but this should not matter, as long as user understand the default

azure/modules/drbd_node/variables.tf
```diff
@@ -83,10 +83,9 @@ variable "bastion_private_key" {
   default     = ""
 }
 
-variable "sbd_enabled" {
-  description = "Enable sbd usage in the HA cluster"
-  type        = bool
-  default     = true
+variable "fencing_mechanism" {
+  description = "Choose the fencing mechanism for the cluster. Options: sbd"
```
Simranpal, line 87: maybe add .`Options: sbd for now`

libvirt/main.tf
```diff
@@ -35,7 +35,8 @@ locals {
   netweaver_virtual_ips = length(var.netweaver_virtual_ips) != 0 ? var.netweaver_virtual_ips : [for ip_index in range(local.netweaver_ip_start, local.netweaver_ip_start + local.netweaver_count) : cidrhost(local.iprange, ip_index + local.netweaver_count)]
 
   # Check if iscsi server has to be created
```
Simranpal, line 37: does this comment need to be moved down? next to `iscsi_enabled`

libvirt/terraform.tfvars.example
```diff
@@ -59,16 +59,16 @@ hana_inst_media = "url-to-your-nfs-share:/sapdata/sap_inst_media/51053381"
...
+#hana_cluster_fencing_mechanism = "sbd"
 
 # Enable Active/Active HANA setup (read-only access in the secondary instance)
 #hana_active_active = true
 
 # SBD related variables
 # In order to enable SBD, an ISCSI server is needed as right now is the unique option
 # All the clusters will use the same mechanism
-# In order to enable the iscsi machine creation sbd_enabled must be set to true for any of the clusters
+# In order to enable the iscsi machine creation _fencing_mechanism must be set to 'sbd' for any of the clusters
```
Simranpal, line 71: `_fencing_mechanism / fencing_mechanism`

libvirt/terraform.tfvars.example
```diff
@@ -161,17 +161,17 @@ ha_sap_deployment_repo = ""
 # Netweaver system identifier
 #netweaver_sid = "HA1"
 
-# Enable SBD for the netweaver cluster
-#netweaver_cluster_sbd_enabled = true
+# Select SBD as fencing mechanism for the Netweaver cluster
```
Simranpal, line 164: should we add comment that the sbd is only option, or there are no other options

Simranpal (approved):
Looks good, just some minor comments below about documentation. The '`native`' name sounds fine to me.

aws/terraform.tfvars.example
arbulu89: This is on purpose as the variables are like `hana_cluster_fencing_mechanism`

libvirt/main.tf
arbulu89: `use_sbd` is only used there, to reduce the complexity, so both line are for the same purpose

libvirt/terraform.tfvars.example
arbulu89: There are not other options

arbulu89:
> looks good to me.
> 
> Only 2 points:
> 
> I find the variable `native` fencing somehow deviating from SBD documentation so I personally already have some difficulties what it means.
> 
> the 2nd point is, how we document this variable.
> 
> If `native` means like the default the cloud provider is offering etc

Hi @MalloZup ,
I have update the fencing documentation. Could you have a look to see if the changes make the usage simpler?

MalloZup (approved):
thx Xabi for documentation. 
Looks good to me. THx for understanding. It not trivial to known all this details when you don't work that closely on that detail! +1

PR #564, opened 2020-10-19
Add new option to authorize additional ssh keys

arbulu89:
Implement the option to authorize additional ssh keys to access the machines providing a list of ssh public key list.

This will be required for the 1st blue-horizon version, as we don't want to ask the private key to the user, as we do now.
So, the project will ask only a public key, and the private/public key to run the salt execution will be hidden to the user.


MalloZup (commented):
I'm ok with this change but I think we need to provide a user documentation hint to .

Isn't this also somehow conflicting with the `public_key_location`?

arbulu89:
> I'm ok with this change but I think we need to provide a user documentation hint to .
> 
> Isn't this also somehow conflicting with the `public_key_location`?

It doesn't conflict. The `public_key_location` is the the primary key used to run the salt provisioning too.
The new variable just authorized new public keys to give access.

About the documentation, what do you mean? The description in the variables.tf is not good enough.
Or would you prefer to have it in the `terraform.tfvars.example` ?

MalloZup:
I'm just reasoning from the UX perspective right now we have following approach

## Before this pr
main.tf
```
public_key_location = "~/Documents/cloud.pub"
# Private SSH Key location
private_key_location = "~/Documents/cloud"
 ```
## user 

```
ssh -o ProxyCommand="ssh -W %h:%p {admin_user}@{bastion_ip} -i {private_key_location} -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" {admin_user}@{bastion_ip} -i {private_key_location} -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no
```

# After this PR:

```
public_key_location = "~/Documents/cloud.pub"
# Private SSH Key location
private_key_location = "~/Documents/cloud"

authorized_keys = "[~/Documents/cloud.pub,  "foo2"]"
```

IMHO this conflicts somehow with the `public_key_location`. Note we reference this way also in other documentation here.


It think we should hide/remove `private_key_location and public_key_location`

**Comment by MalloZup** [comment]
the same discourse is also with the `variable "bastion_private_key" ` . I think  we should aim like if we are going to use 

```authorized_keys = "[~/Documents/cloud.pub,  "foo2"]" why such keys can't be the only entry-point?

**Comment by arbulu89** [comment]
@MalloZup I think it's not that simple.
First of all, we have a hard requirement where we need to supply a public/private ssh keys combo to run the salt execution.
That's why we have the `puclic_key_location` and `private_key_location`. If we don't have the `private_key_location`, there might be the case where the ssh keys are not correctly configured in the local machine. If we remove them, and only use the `authorized_keys`, we would need to trust that the users have always the correct setup.

The other options as you suggested, is to create the sshkeys by the code to run `salt` by terraform, and hide this logic, but we would be managing sshkeys silently (but some users don't like), and even having the private key content in the terraform `tfstate`.

These are architectural decisions that we can obviously do, but all of the have good and bad consequences. Terraform, already has a native option to create and manage SSH keys: https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key

About the bastion, we were explicitily asked to include the option where the key to connect to the bastion, and from there to the machines must be different. Later on, the user might decide to use the same key. So, we are not authorizing the access to the machines, we are authorizing the access to the bastion. For that, we need at least 2 variables.

**Review by MalloZup — APPROVED** [review]
I think we need some documentation explaning all the various ssh keys so the user is more informed about this and the different combination.  *with examples. 
I think maybe a markdown section could be helpful enough
Since it is already right now imho quite not clear what is doing what.

File: aws/variables.tf [hunk 1]
```diff
@@ -63,6 +63,12 @@ variable "private_key_location" {
   type        = string
 }
 
+variable "authorized_keys" {
+  description = "List of authorized SSH public keys path to access the created machines with the used admin user (root in this case)"
```
Simranpal, line 67: Will it be` path/paths`

`List of authorized SSH public keys paths, to access the created machines with the used admin user (root in this case)`

arbulu89:
@MalloZup I have added more documentation.
Please, be aware that the usage on libvirt and the cloud providers is different. In libvirt, the ssh connection to run `salt` is done using `user/password` while in the cloud only keys are used to avoid security holes.

Simranpal (approved):
Should the description about `authorized_keys` not contain the admin user relevant for each cloud providers? like `ec2user`, `cloudadmin` etc. Right now it says only `root`

aws/variables.tf
arbulu89: Yes, I was just changing that hehe

MalloZup (commented):
ok conflicts need to be solved

PR #567, opened 2020-10-19
[BLUE-Horizon] Move the code into our repo 

MalloZup:
The main goal of this pr is to move the rails application into `blue-horizon` directory. (forked) to our `blue-horizon` directory

This is needed because we will need to modify fonts, we need to track the TF files we need to change Icons, fonts, color and other changes for devel

It is a move from commit https://github.com/SUSE-Enceladus/blue-horizon/commit/33d3743a2d82d8f70d39f42330e953b2db0a05b1

* since the vendor are specific, TF it make sense to track them *remove the gitignore 
* same discourse apply for customisation we do, we need to track it in * remove the gitignore

If you think we do have another way, feel free to propose but I think that is the most "sane" way to keep all the change unified without have any other coupling/dep issue

# The workflow:

the new workflow would be that we work completely on this branch and add needed customisation/layout and terraform changes etc.


arbulu89:
@MalloZup I don't get this. Why would we want this?
Why we don't just work in one fork of the initial project?

MalloZup:
@arbulu89  I moved this here because:

- how we will change the color of the buttons.
- how we will add `monitoring` link
- how we will add `saptune` webpage and so on?

I mean this changes, where we do add them if they are deviating from upstream? To me looks like a fork also on the ruby webapp.

Or how we will change all this upstream? we could create a branch upstream

MalloZup:
ok for moment closing

PR #569, opened 2020-10-19
WIP: Set to default some variables and remove user input

MalloZup:
# desc

Simplify variables.

How it looks with this PR

image

### DIFF
--- a/blue_horizon/azure/sources/main.tf
+++ b/blue_horizon/azure/sources/main.tf
@@ -99,6 +99,7 @@ locals {
       app_server_count            = 10
     }
   }
+  sles4sap_version = "SLES15SP2"
   sles_version = {
     SLES15 = {
       publisher = "SUSE"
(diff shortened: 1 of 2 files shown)


arbulu89:
@MalloZup As far as I see, this are the unique required parameters that the user should add in my opinion and we cannot hardcoded:
- Deployment name (this could even be a random string that we can later output)
- System identifier
- System Instance number
- System master password

We need to discuss:
- Authorized key. I think that users like to use their own keys. Otherwise, we need to output the public and private keys content
- Being HA or not. For me this is important, as the non HA deployments are really common
- SLES4SAP version. I would use the latest version (SLE15SP2 now)
- Admin user. We can just hardcode and output at the end


Other variables could be based on `deployment_name` (like primary site, secondary site)

The last 3 will be configurable after merging: https://github.com/SUSE/ha-sap-terraform-deployments/pull/561

MalloZup:
@arbulu89 I think we can review this.
For the variable I didn't addressed on this PR I created a follow-up.


 https://github.com/SUSE/ha-sap-terraform-deployments/issues/570 followup

arbulu89:
@MalloZup I guess that as an initial work we could merge it. It won't work in any case, that's true.
Feel free to merge if you want, we will do more changes for sure, but for the current state it's fine

MalloZup:
ok thx lets merge it and follow-up.

PR #572, opened 2020-10-20
[Blue-horizon]: Add groups  for better visualisation

MalloZup:
# Description:

This pr add possibility to group variable in groups which are visualized in different frames.

I prefer the `hana and azure` combination. I think having 2 categories is most ordered and simple approach

# Possible grouping:

##  1)  Hana and Azure:

image

## 2)  Hana and Storage:

image

## 3) Additional data, Hana, Azure 

image

### DIFF
--- a/blue_horizon/azure/sources/variables.tf.json
+++ b/blue_horizon/azure/sources/variables.tf.json
@@ -1,10 +1,5 @@
 {
   "variable": {
-    "azure_region": {
-      "type": "string",
-      "default": "westeurope",
-      "description": "Azure region where the deployment machines will be created <!--[group:azure]-->"
-    },
     "ssh_authorized_keys": {
       "type": "list",
       "description": "List of additional authorized SSH public keys paths to access the created machines with the used admin user <!--[group:hana]-->"
@@ -26,7 +21,12 @@
     "deployment_type": {
       "type": "string",
       "default": "2-tier",
-      "description": "Select the desired deployment type. Options=[\"2-tier\",\"2-tier-HA\"]. More information in: https://documentation.suse.com/sbp/all/html/SBP-SAP-AzureSolutionTemplates/index.html <!--[group:azure]-->"
+      "description": "Select the desired deployment type. Options=[\"2-tier\",\"2-tier-HA\"]. More information in: https://documentation.suse.com/sbp/all/html/SBP-SAP-AzureSolutionTemplates/index.html <!--[group:hana]-->"
+    },
+    "azure_region": {
+      "type": "string",
+      "default": "westeurope",
+      "description": "Azure region where the deployment machines will be created <!--[group:azure]-->"
     },
     "deployment_size": {
       "type": "string",
--- a/blue_horizon/azure/sources/variables.tf.json
+++ b/blue_horizon/azure/sources/variables.tf.json
@@ -3,7 +3,7 @@
     "azure_region": {
       "type": "string",
       "default": "westeurope",
-      "description": "Azure region where the deployment machines will be created <!--[group:hana]-->"
+      "description": "Azure region where the deployment machines will be created <!--[group:azure]-->"
     },
     "ssh_authorized_keys": {
       "type": "list",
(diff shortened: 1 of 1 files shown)


MalloZup:
I prefer the 1) approach

Simranpal (commented):
I think we should not reduce the user inputs, in order to make UI look better. Some of the parameters will be required by user to customize.
Or atleast we should tell the defaults in case we do not allow them to customize.

@MalloZup is there an option to show/hide group or make certain group collapse, or have a checkbox. So the additional data is there to configure if needed

arbulu89:
@MalloZup Let's postpone or delegate this design to @petersatsuse . He has more SAP domain knowledge

arbulu89:
> I think we should not reduce the user inputs, in order to make UI look better. Some of the parameters will be required by user to customize.
> Or atleast we should tell the defaults in case we do not allow them to customize.
> 
> @MalloZup is there an option to show/hide group or make certain group collapse, or have a checkbox. So the additional data is there to configure if needed

@Simranpal There is not such a feature right now. All the variables are showed always and we cannot create dependencies between them

petersatsuse:
I would start first with showing some per-requisits, like the fileshare for the sap media (and some example), ssh-key

Next would be the Azure infrastructure needs, like the users need to do in the Azure Portal
- Region, instance username, sshkeys, fileshare details for SAP Media

Next i would go to the SAP parts
- Deployment type: NW,HANA,NW+HANA,S/4,...
- Deployment size (demo,small.medium,large) 
  and this should showcase the SAPS and number of systems behind the chosen size
- SAP SID , instance#, etc dependent on the chosen type
- SAP masterpw
- HA install yes/no

The other details should be hidden in additional vars-files and should be loaded dependent on the chosen selection. Not sure if its possible, but showing all variables should not be the way. If customers/partner need to tweak everything they should use the cmdline/file based option for deployments.
Not sure if its doable, but i remember that Xabi showed something similar like above at some of the sprint reviews as reseach idea

arbulu89:
> I would start first with showing some per-requisits, like the fileshare for the sap media (and some example), ssh-key
> 
> Next would be the Azure infrastructure needs, like the users need to do in the Azure Portal
> 
>     * Region, instance username, sshkeys, fileshare details for SAP Media
> 
> 
> Next i would go to the SAP parts
> 
>     * Deployment type: NW,HANA,NW+HANA,S/4,...
> 
>     * Deployment size (demo,small.medium,large)
>       and this should showcase the SAPS and number of systems behind the chosen size
> 
>     * SAP SID , instance#, etc dependent on the chosen type
> 
>     * SAP masterpw
> 
>     * HA install yes/no
> 
> 
> The other details should be hidden in additional vars-files and should be loaded dependent on the chosen selection. Not sure if its possible, but showing all variables should not be the way. If customers/partner need to tweak everything they should use the cmdline/file based option for deployments.
> Not sure if its doable, but i remember that Xabi showed something similar like above at some of the sprint reviews as reseach idea

@petersatsuse 
All of this is totally doable, but we need to decide how to show the information.

From your suggestions, what I understand is:
- In the `welcome` page, we do an introduction of the page and we explain how to setup the SAP HANA media share and create the SSH keys
- In the cluster sizing panel we show the `deployment_size` (the text still pending to be defined)
- In the variables page, we have to groups:
  - `Infrastructure configuration` with
    - `Azure region` (it can have a default value in the input)
    - `Cloud administrator name` (it can have a default value in the input or we can even hardcode)
    - `Deployment name` (it can have a default value in the input or we can even hardcode)
    - `Authorized SSH public key`  (mandatory value)
  - `HANA configuration group`
    - `System identifier` (it can have a default value in the input)
    - `System instance number` (it can have a default value in the input)
    - `System master password` (mandatory value)
    - `HANA installation media data` I'm setting this inside HANA, as in the future, if we want to add Netweaver, the media might be different.

We don't need more variables by now. What do you think @petersatsuse @MalloZup

petersatsuse:
The media place should be always the same place with sub-directories. Idea behind it that you have a library of SAP medias which can be used in more deployments. Therefor the main url for the share is always the same, but the media may be in a different sub-directory.

The second part you limit only to HANA, guess as a initial start.
I would like to have it prepared / think through how we need it for all types of deployment,
so we do not have blockers during the next steps.

Otherwise i'm fine.. :)

MalloZup:
ok i'm fine thx for feedback.

My only point is the category name: `Infrastructure configuration`. Maybe it resonate better to `Cloud Infrastructure` 

I personally think the word `configuration` is implicit, here. What do you think?

meanwhile I will prepare a pr so we can see an initial state/screenshot of this.

MalloZup:
closing as discussed

PR #574, opened 2020-10-22
Improve blue-horizon variables usage

arbulu89:
Some changes to make Blue-horizon work properly with the current `develop` branch.

After this, I won't do more changes. 
@petersatsuse will take care of tuning the variables, the text and the size cluster panel

### DIFF
--- a/blue_horizon/azure/locales/custom-en.yml
+++ b/blue_horizon/azure/locales/custom-en.yml
@@ -10,7 +10,40 @@ en:
   lead: "Deploy your SAP deployments using SLES4SAP in Azure"
   # Long-form description of the application
   # Markdown syntax is accepted - see https://daringfireball.net/projects/markdown/syntax
-  description: "Here the solution you were looking for to deploy your SAP workspaces using SLES4SAP in some few clicks!<br>This tool will create your SAP infrastructure and provision the machines to have an up and running environment in blink of an eye!<br>It just provides a basic set of options that we think are the most used ones.<br>For a more advanced customization you will need to get your hands dirty and play with the source project: https://github.com/SUSE/ha-sap-terraform-deployments"
+  description: |
+    Here the solution you were looking for to deploy your SAP workspaces using SLES4SAP in some few clicks!
+
+    This tool will create your SAP infrastructure and provision the machines to have an up and running environment in blink of an eye!
+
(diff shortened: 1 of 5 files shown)


blue_horizon/azure/sources/main.tf
```diff
@@ -100,72 +100,72 @@ locals {
...
-}
-
-output "bastion_ip" {
-  value = module.bluehorizon.bastion_public_ip
-}
-
-output "hana_ips" {
-  value = join(",", module.bluehorizon.cluster_nodes_ip)
+  #ha_sap_deployment_repo             = "https://download.opensuse.org/repositories/network:ha-clustering:sap-deployments:devel"
+  netweaver_master_password          = "not used"
```
MalloZup, line 170: why do we have netewaver here?

MalloZup (approved):
look good! 
can you attach a screenshot so we have a history how it looked like at this time?

blue_horizon/azure/locales/custom-en.yml
```diff
@@ -10,7 +10,40 @@ en:
   lead: "Deploy your SAP deployments using SLES4SAP in Azure"
   # Long-form description of the application
   # Markdown syntax is accepted - see https://daringfireball.net/projects/markdown/syntax
-  description: "Here the solution you were looking for to deploy your SAP workspaces using SLES4SAP in some few clicks!<br>This tool will create your SAP infrastructure and provision the machines to have an up and running environment in blink of an eye!<br>It just provides a basic set of options that we think are the most used ones.<br>For a more advanced customization you will need to get your hands dirty and play with the source project: https://github.com/SUSE/ha-sap-terraform-deployments"
+  description: |
```
MalloZup, line 13: I guess this is something peter will need to have look this description

arbulu89: Yes

blue_horizon/azure/sources/main.tf
arbulu89: Now the netweaver sidadm password doesn't have a password. This is something we will need to think about

MalloZup: @arbulu89  ok open then a follow-up issue we can work and forget about here in this pr :dog:

PR #575, opened 2020-10-22
Increate wait_for_initialization to 120 to avoid sporadic errors

arbulu89:
Time to time, the cluster joining process fails because the 1st node is not totally initialized (specially in Azure).
To avoid that, I have set a really conservative time that won't make a big difference in the overall deployment time, but it will avoid these errors.


MalloZup (approved):
lgtm

PR #576, opened 2020-10-23
Feature/bastion proxy server

arbulu89:
This PR implements a Nginx proxy server in the Azure bastion machine. This helps to access the monitoring machine through the bastion, so we don't need to have an additional public address.

In order to achieve that:
- Bastion machine uses now the `common_variables` variables. This is important to be able to apply some of the configuration we apply during salt provisioning, like authorizing additional SSH keys. This is required in blue-horizon
- The bastion runs the salt provisioner to install and configure nginx
- As the bastion machine must not be rebooted (this would cut the connection to the other machines), implement a new variable to `enable/disable` the reboot
- Set the `bastion_host` in the modules to avoid cyclic imports
- Bastion image OS is configurable.

**INFO: It only opens `http` connection by now**

FYI: @petersatsuse

### DIFF
--- a/azure/modules/bastion/main.tf
+++ b/azure/modules/bastion/main.tf
@@ -1,6 +1,5 @@
 locals {
-  bastion_enabled = var.bastion_enabled ? 1 : 0
-  public_key      = fileexists(var.public_key) ? file(var.public_key) : var.public_key
+  bastion_enabled = var.common_variables["bastion_enabled"] ? 1 : 0
 }
 
 
@@ -65,7 +64,7 @@ resource "azurerm_network_interface" "bastion" {
   }
 
   tags = {
-    workspace = var.deployment_name
+    workspace = var.common_variables["deployment_name"]
   }
 }
 
@@ -78,7 +77,7 @@ resource "azurerm_public_ip" "bastion" {
   idle_timeout_in_minutes = 30
 
   tags = {
-    workspace = var.deployment_name
+    workspace = var.common_variables["deployment_name"]
   }
 }
 
@@ -108,15 +107,15 @@ resource "azurerm_virtual_machine" "bastion" {
 
   os_profile {
     computer_name  = "vmbastion"
-    admin_username = var.admin_user
+    admin_username = var.common_variables["authorized_user"]
   }
 
   os_profile_linux_config {
     disable_password_authentication = true
 
     ssh_keys {
-      path     = "/home/${var.admin_user}/.ssh/authorized_keys"
-      key_data = local.public_key
+      path     = "/home/${var.common_variables["authorized_user"]}/.ssh/authorized_keys"
+      key_data = var.common_variables["bastion_public_key"]
     }
   }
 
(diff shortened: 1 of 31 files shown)


MalloZup (commented):
thx Xabi for this.  I will review it on monday more carefullu. but sofar seems good.


My thoughts were, should we expose also  9090 prometheus webui server port? :thinking:  
Maybe peter could also help.

I think personally we shouldn't but it might be useful to sap admins

arbulu89:
> thx Xabi for this. I will review it on monday more carefullu. but sofar seems good.
> 
> My thoughts were, should we expose also 9090 prometheus webui server port? thinking
> Maybe peter could also help.
> 
> I think personally we shouldn't but it might be useful to sap admins

Yes, we can obviously export the prometheus and the hawk webs (maybe this second is not that simple as it only works with https). Personally, I don't see much value exposing prometheus web, as it doesn't provide a lot of value for SAP/HA users.

MalloZup:
yep I guess the prometheus ui for the normal users can be hidden where the HAWK ui could exposed. yop. 

Looks good. I think the bastion is evolving to a feature set which we might need to document more consistently, and I dunno if we could put in a longterm roadmap to other clouds .

salt/bastion/templates/nginx.conf.j2
```diff
@@ -0,0 +1,16 @@
...
+events {
+    worker_connections  1024;
+    use epoll;
+}
+
+http {
+  server {
+    listen 3000;
+    listen [::]:3000;
+    location / {
```
Simranpal, line 12: is the location `/` correct

Simranpal (approved):
looks good, I did not get why `bastion_host` was needed to be defined in other modules. What did you mean by cyclic imports?

azure/modules/bastion/main.tf
```diff
@@ -43,6 +43,21 @@ resource "azurerm_network_security_group" "bastion" {
   }
 }
 
+resource "azurerm_network_security_rule" "grafana" {
```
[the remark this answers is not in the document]
arbulu89: Not by now, let's only provide grafana.
If we are asked, we will change this in the future

salt/bastion/templates/nginx.conf.j2
arbulu89: Yes, it is

arbulu89:
> looks good, I did not get why `bastion_host` was needed to be defined in other modules. What did you mean by cyclic imports?

In order to use `common_variables` in the bastion module, we must remove the `bastion_host` as this is created in the bastion module. So each of the modules would require a variable from the other, creating the cyclic import

Simranpal:
> > looks good, I did not get why `bastion_host` was needed to be defined in other modules. What did you mean by cyclic imports?
> 
> In order to use `common_variables` in the bastion module, we must remove the `bastion_host` as this is created in the bastion module. So each of the modules would require a variable from the other, creating the cyclic import

Thanks got it.

PR #577, opened 2020-10-26
Sap blue horizon text changes

petersatsuse:
Text changes to provide more context
adaption of the instance size to new types
fix large instance size as it had complete wrong settings

### DIFF
--- a/blue_horizon/azure/locales/custom-en.yml
+++ b/blue_horizon/azure/locales/custom-en.yml
@@ -7,42 +7,78 @@ en:
   # Abbreviated title for use in navigation
   short_title: "BlueHorizon"
   # Brief tag line shown below the title
-  lead: "Deploy your SAP deployments using SLES4SAP in Azure"
+  lead: "Deploy your SAP Landscapes using SLES4SAP in Azure"
   # Long-form description of the application
   # Markdown syntax is accepted - see https://daringfireball.net/projects/markdown/syntax
+  # 
+  ## Note: to minimize the text on the page, i would like to the details with the software
+  ## out to a webpage/document instead of showing it here
+  #
   description: |
-    Here the solution you were looking for to deploy your SAP workspaces using SLES4SAP in some few clicks!
+    The solution will deploy the needed Azure infrastructure for SAP Netweaver and SAP HANA 
+    and provision and configure the virtual machines and install the SAP application for you.
+    
+    Doing SAP sizing right is not easy, therefor a simplified sizing approach is used. 
+    With the well known concept of T-Shirt sizes S,M,L we provide a simple way of using different settings. 
+    
+    Behind such T-Shirt sizes, we use the right Azure instance and the needed number of disks with the right size.
+    The Demo and S size are thought for non-production, the M and L will use SAP certified instance types and can be used for production scenarios.
+    Below is a rought estimation about the SAPS behind the three sizes:
 
-    This tool will create your SAP infrastructure and provision the machines to have an up and running environment in blink of an eye!
-
-    It just provides a basic set of options that we think are the most used ones.
-
-    For a more advanced customization you will need to get your hands dirty and play with the source project: https://github.com/SUSE/ha-sap-terraform-deployments
-
+    * Small  <  30.000 SAPS
+    * Medium <  70.000 SAPS
+    * Large  < 180.000 SAPS
+    
+    **Please be aware that it may require a high amount of resources, therefor an increase of your quota could be necessary**.
+    
+    For a more advanced customization you will need to get your hands dirty and work directly with the source project: https://github.com/SUSE/ha-sap-terraform-deployments
 
     ## Pre-requirements
+    1. You need to have ssh-keys to be able to access the deployed machines.
 
-    In order to use the project the SAP installation software must be available for the installer.
-    The software can be downloaded from https://launchpad.support.sap.com/#/softwarecenter
+    2. The SAP software must be available in Azure, in order to be installed by the solution.
 
-    In order to use the software, create a new Storage account in your Azure subscription and upload the downloaded format to a file share.
+        The SAP software can be downloaded from https://launchpad.support.sap.com/#/softwarecenter
 
-    **This Storage account has an account name and key. Store them as they are needed in the next steps.** They can be collected from the Azure portal.
+    3. In order to provide the SAP software in Azure, please create a storage account in your Azure subscription and upload the downloaded format to a file share.
+       See Microsoft documentation [Howto create a storage account][]
+       [Howto create a storage account]:https://docs.microsoft.com/en-us/azure/storage/files/storage-how-to-create-file-share?tabs=azure-portal#create-a-storage-account
 
-    ### SAP HANA
+        **This storage account has an account name and key. Store them as they are needed in the next steps.** They can be collected from the Azure portal.
 
-    This software is available in different formats.
+    ### SAP Software
+    The SAP software is available in different formats from the SAP Software Download Center.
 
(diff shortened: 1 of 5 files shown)


blue_horizon/azure/locales/custom-en.yml
```diff
@@ -7,42 +7,78 @@ en:
...
+  # 
+  ## Note: to minimize the text on the page, i would like to the details with the software
+  ## out to a webpage/document instead of showing it here
+  #
   description: |
-    Here the solution you were looking for to deploy your SAP workspaces using SLES4SAP in some few clicks!
+    The solution will deploy the needed Azure infrastructure for SAP Netweaver and SAP HANA 
+    and provision and configure the virtual machines and install the SAP application for you.
+    
+    Doing SAP sizing right is not easy, therefor a simplified sizing approach is used.
```
arbulu89, line 21: Typo `therefor/therefore`

blue_horizon/azure/locales/custom-en.yml
```diff
@@ -7,42 +7,78 @@ en:
...
+    1. You need to have ssh-keys to be able to access the deployed machines.
 
-    It just provides a basic set of options that we think are the most used ones.
+    2. The SAP software must be available in Azure, in order to be installed by the solution.
 
-    For a more advanced customization you will need to get your hands dirty and play with the source project: https://github.com/SUSE/ha-sap-terraform-deployments
+        The SAP software can be downloaded from https://launchpad.support.sap.com/#/softwarecenter
 
+    3. In order to provide the SAP software in Azure, please create a storage account in your Azure subscription and upload the downloaded format to a file share.
+       See Microsoft documentation [Howto create a storage account][]
```
arbulu89, line 44: `[Howto create a storage account][]` looks like duplicated

blue_horizon/azure/locales/custom-en.yml
```diff
@@ -7,42 +7,78 @@ en:
...
+       [Howto create a storage account]:https://docs.microsoft.com/en-us/azure/storage/files/storage-how-to-create-file-share?tabs=azure-portal#create-a-storage-account
```
arbulu89, line 45: To create a link follow should be [Howto create a storage account](https://docs.microsoft.com/en-us/azure/storage/files/storage-how-to-create-file-share?tabs=azure-portal#create-a-storage-account)

blue_horizon/azure/locales/en.yml
```diff
@@ -53,7 +53,7 @@ en:
   sidebar:
     welcome: "Welcome"
     sources: "Edit sources"
-    cluster: "Size cluster"
+    cluster: "Sizes"
```
arbulu89, line 56: This change should go in `custom-en.yml` rather than `en.yml`. Just add in `en`:
```
sidebarL
  cluster: "Sizes"
```

blue_horizon/azure/locales/en.yml
```diff
@@ -77,7 +77,7 @@ en:
 
   page_title:
     plan: "Plan your deployment"
-    deploy: "Deploy your cluster"
+    deploy: "Deploy your machines"
```
arbulu89, line 80: The same. It should go in `custom-en.yml`

blue_horizon/azure/sources/variables.tf.json
```diff
@@ -3,41 +3,49 @@
     "azure_region": {
       "type": "string",
       "default": "westeurope",
-      "description": "Azure region where the deployment machines will be created<!-- [group:general_configuration] -->"
+      "description": "Azure region where the machines will be created<!-- [group:Azure_configuration] -->"
+    },
+    "ssh_authorized_keys": {
```
arbulu89, line 8: I don't know if this should be a list or an individual box. If we finally merge the option to upload files, we won't be able to use lists.

arbulu89: Actually, i see later `ssh_authorized_key`, so this one might not be needed

blue_horizon/azure/sources/variables.tf.json
```diff
@@ -3,41 +3,49 @@
...
+      "description": "Azure storage account access key<!-- password_key [group:Azure_configuration] -->"
     },
     "instance_type": {
       "type": "string"
     },
     "system_identifier": {
       "type": "string",
       "default": "prd",
-      "description": "HANA system identifier. System identifier must be 3 letters string"
+      "description": "HANA SID. The System identifier (SID) must be 3 letters"
```
arbulu89, line 39: The SID accepts numbers to (not as first character). We will work on the limitations in the short future applying regular expressions

blue_horizon/azure/sources/variables.tf.json
```diff
@@ -3,41 +3,49 @@
...
     },
     "storage_account_key": {
       "type": "string",
-      "description": "Azure storage account access key<!-- password_key [group:general_configuration] -->"
+      "description": "Azure storage account access key<!-- password_key [group:Azure_configuration] -->"
     },
     "instance_type": {
       "type": "string"
     },
     "system_identifier": {
```
arbulu89, line 36: I think we would need to add the HANA entries in a group to (`HANA configuration` maybe). Otherwise they will be grouped as `Additional data`

blue_horizon/azure/sources/variables.tf.json
```diff
@@ -51,6 +59,14 @@
     "hana_installation_software_path": {
       "type": "string",
       "description": "Path to the folder where the SAP HANA installation software is uploaded in the Storage account. It is composed of the file share name and the uploaded HANA folder or archive file. For example: myshare/51053381_part1.exe or myshare/51053381"
+    },
+    "storage_account_name": {
```
arbulu89, line 63: This last 2 entries look duplicated, right? 
They are set in the `Azure configuration`

arbulu89 (changes requested):
@petersatsuse I like the new text. Good job!
I suggested some changes in the variables page.

blue_horizon/azure/locales/custom-en.yml
petersatsuse: no this is the text for the link - markdown syntax to show a text instead of a link

petersatsuse: yes

blue_horizon/azure/sources/variables.tf.json
petersatsuse: i was also wondering about that double entry. i only changed the group, and i think the some more variables should go to groups

blue_horizon/azure/sources/variables.tf.json
petersatsuse: There is a whole regex needed, as SAP also has some values not allowed. We have it in the yast install wizard ... i need to search for it.

blue_horizon/azure/sources/variables.tf.json
petersatsuse: yes - think so too

blue_horizon/azure/sources/variables.tf.json
petersatsuse: hmm did not add them - but yes you are right

blue_horizon/azure/sources/variables.tf.json
arbulu89: This is the expression: `/^(?!ADD|ALL|AND|ANY|ASC|COM|DBA|END|EPS|FOR|GID|IBM|INT|KEY|LOG|MON|NIX|NOT|OFF|OMS|RAW|ROW|SAP|SET|SGA|SHG|SID|SQL|SYS|TMP|UID|USR|VAR)[A-Z][A-Z0-9]{2}$`.
Anyway, this will come hidden somehow in the new feature to add client side validation

MalloZup:
can we have a screenshot to visualize this? Tia :sunflower:

blue_horizon/azure/locales/custom-en.yml
arbulu89: This is still wrong. The `[text]:link` doesn't work. It must be `[text](link)`. Otherwise nothing is visualized

arbulu89 (changes requested):
@petersatsuse I still see some needed improvements.
1. The `[Howto create a storage account]` link is still not created properly
2. `deployment_type` should go in `hana_configuration` group, or some other group. Having it in `Additional data` looks strage. 
3. The `ssh_authorized_keys` value and description is invalid. We say `keys paths`, but we cannot use this way. For that we should upload the file. Right now, the options is to paste the content of the public key there.

arbulu89 (approved):
Thanks @petersatsuse 
Nice work

PR #578, opened 2020-10-26
Disable loki and promtail configuration

Simranpal:
With this PR, the `loki` and `promtail` configuration is disabled. It is because the package is not available from official SUSE repositories and package.

Basically, commented the parts related to `loki` and `promtail`. I have tested with deployment.


salt/monitoring_srv/init.sls
```diff
@@ -1,4 +1,3 @@
 include:
   - .prometheus
   - .grafana
-  - .loki
```
arbulu89, line 4: I would comment this instead of removing

salt/monitoring_srv/loki.sls
```diff
@@ -1,3 +1,4 @@
+{# loki package is currently disabled. To use it, add the .loki sls entry in init.sls and enable loki as datasource in grafana/datasources template #}
```
arbulu89, line 1: If you comment/remove the `loki` entry in the init.sls file, we don't really need to comment this

arbulu89 (approved):
It looks good. Just some minor comments

salt/monitoring_srv/init.sls
MalloZup: yes

MalloZup (changes requested):
I'm not sure it will work this because we are missing the client part.

https://github.com/SUSE/ha-sap-terraform-deployments/blob/16b27530f03552448d3610b598a8602e8b04f747/salt/cluster_node/monitoring.sls#L30

Simranpal:
closing this as too many conflict issues, I have created https://github.com/SUSE/ha-sap-terraform-deployments/pull/579 for `sap-blue-horizon` branch as discussed

PR #580, opened 2020-10-27
Feature/add common variables

Simranpal:
We have added a lot of new variables, that  are mainly used to render the salt pillar files. With this PR, we move most of the HANA/Netweaver variables from all the modules of all providers in the "common_variables" module. 
With this we will get:

- `common_variables` module contain most of the hana/nw variables for all providers and one place to configure
- The grains will be the same for this topic, avoiding issues like ("I have forgotten to add this variable to this cloud provider")
- We reduce the duplicated code in the `salt_provisioner.tf` files
- We wouldn't have any change for the user

In addition to common variables. I have renamed some variables to be consistent in all providers:
- libvirt variable `hana_inst_media` to `hana_inst_master`
- gcp variable `sap_hana_deployment_bucket` to `hana_inst_master`

Below is variables added to common variables:
`HANA:`
  hana_sid              
  hana_instance_number             
  hana_cost_optimized_sid           
  hana_cost_optimized_instance_number 
  hana_master_password              
  hana_cost_optimized_master_password 
  hana_primary_site                  
  hana_secondary_site                
  hana_inst_master                   
  hana_inst_folder                  
  hana_fstype                       
  hana_platform_folder            
  hana_sapcar_exe                  
  hana_archive_file                
  hana_extract_dir                 
  scenario_type                      
  hana_cluster_vip                   
  hana_cluster_vip_secondary         
  hana_ha_enabled                    
  hana_cluster_fencing_mechanism    
  sbd_storage_type

`Netweaver:`

netweaver_sid                      
  ascs_instance_number                
  ers_instance_number                 
  pas_instance_number                 
  netweaver_master_password          
  netweaver_product_id                
  netweaver_inst_folder               
  netweaver_extract_dir              
  netweaver_swpm_folder              
  netweaver_sapcar_exe               
  netweaver_swpm_sar                 
  netweaver_sapexe_folder           
  netweaver_additional_dvds          
  netweaver_nfs_share               
  hana_ip                          
  netweaver_hana_sid               
  netweaver_hana_instance_number   
  netweaver_hana_master_password   
  netweaver_ha_enabled            
  netweaver_cluster_fencing_mechanism
  netweaver_sbd_storage_type


MalloZup (commented):
I'm not sure if this logically is correct.

e.g `common_variables` are imho `host` releated variables.

If we add `hana/nw` the category becomes more ambigous

aws/main.tf
```diff
@@ -188,51 +203,36 @@ module "netweaver_node" {
...
+  key_name                   = aws_key_pair.key-pair.key_name
+  security_group_id          = local.security_group_id
+  route_table_id             = aws_route_table.route-table.id
+  aws_credentials            = var.aws_credentials
+  aws_access_key_id          = var.aws_access_key_id
+  aws_secret_access_key      = var.aws_secret_access_key
+  host_ips                   = local.hana_ips
+  hana_data_disk_type        = var.hana_data_disk_type
+  hana_data_disk_size        = var.hana_data_disk_size
+  hana_inst_master           = var.hana_inst_master
```
arbulu89, line 226: I still see some variable that I think should go to common_variables.
Remember, common variables are the variables that are used in all cloud providers.
- `hana_inst_master`
- `hana_cluster_vip`
- `hana_cluster_vip_secondary`
- `fencing_mechanism` (that in the common variables would be `hana_cluster_fencing_mechanism`
- `sbd_storage_type`
- `iscsi_srv_ip`
- `cluster_ssh_pub`
- `cluster_ssh_key`
`hwcct` should be removed as it is already in `common_variables`

arbulu89 (commented):
@Simranpal It looks good. There are still variables that I would set in the `common_variables` though.
Remember, common variables are the variables that are used in all cloud providers.
Just, maybe some of them cannot be added there, because we would create cyclic imports (like in the bastion PR). So they must be excluded by force

generic_modules/common_variables/hana_variables.tf
```diff
@@ -0,0 +1,81 @@
...
+variable "hana_archive_file" {
+  description = "Path to the HANA database server installation SAR archive or HANA platform archive file in zip or rar format, relative to the 'hana_inst_master' mounting point. Use this parameter if the hana media archive is not already extracted"
+  type        = string
+  default     = ""
+}
+
+variable "hana_extract_dir" {
+  description = "Absolute path to folder where SAP HANA archive will be extracted"
+  type        = string
+  default     = "/sapmedia/HANA"
```
arbulu89, line 75: This should be `/sapmedia_extract/HANA`.
In any case, we should consider removing all the `default` as they are already set in the main `variables.tf` files.

MalloZup (commented):
we should rename `common_variable` to `common_cloud_variables` or `sap_common_variables`. or related

Right now the `common_variable` is empty like `utils`, it is a bad naming

aws/main.tf
Simranpal: @arbulu89 regarding this variables:

- `hana_inst_master` I added now, replacing `hana_inst_media` in libvirt with `hana_inst_master` so its common on all providers
- `hana_cluster_vip` this variable is set differently inside Azure grains, so we cannot have it same for all providers
- `hana_cluster_vip_secondary` is used in the terraform hana module code, so if we want to use it in `common_variables` module, it also need to assigned in module "`hana_node`" in main.tf.  I am not sure if this will cause confusion with 2 assignments in main.tf.
- `sbd_storage_type` is used in the terraform hana module code for libvirt, again same issue as 2 assignments in main.tf. One in common variable and other in hana module. Besides that it used in multiple modules like hana, nw, drbd with the same value. And other grains entry depend on it like `sbd_disk_device`. So defining in two places causes confusion which one is being used.
- `ha_enabled` same issue as 2 assignments
- `cluster_ssh_pub`, `cluster_ssh_key` are not available in libvirt
- `iscsi_srv_ip` and `hwcct` was not in common_variables but I added now

Simranpal: `iscsi_srv_ip` in common variables caused cycle issue, so I reverted back

Simranpal: Similarly `hana_inst_master` did not exist for GCP, so I will revert to existing usage

Simranpal:
> we should rename `common_variable` to `common_cloud_variables` or `sap_common_variables`. or related
> 
> Right now the `common_variable` is empty like `utils`, it is a bad naming

I am not sure about a good naming here. As it will contain variables common to all providers and various modules, the name of module itself `common_variables` is reasonable to me.

arbulu89:
> > we should rename `common_variable` to `common_cloud_variables` or `sap_common_variables`. or related
> > Right now the `common_variable` is empty like `utils`, it is a bad naming
> 
> I am not sure about a good naming here. As it will contain variables common to all providers and various modules, the name of module itself `common_variables` is reasonable to me.

I like `common_variables` too. Adding cloud is redundant in my opinion, the project is already based on cloud platforms, why would we like to add it again here?

aws/main.tf
```diff
@@ -55,24 +55,41 @@ locals {
...
+  hana_secondary_site                 = var.hana_secondary_site
+  hana_inst_folder                    = var.hana_inst_folder
+  hana_fstype                         = var.hana_fstype
+  hana_inst_master                    = var.hana_inst_master
+  hana_platform_folder                = var.hana_platform_folder
+  hana_sapcar_exe                     = var.hana_sapcar_exe
+  hana_archive_file                   = var.hana_archive_file
+  hana_extract_dir                    = var.hana_extract_dir
+  scenario_type                       = var.scenario_type
+  hwcct                               = var.hwcct
```
arbulu89, line 92: I would put `hwcct` just after `qa_mode`, to have things in order

generic_modules/common_variables/outputs.tf
```diff
@@ -48,6 +48,25 @@ monitoring_enabled: ${var.monitoring_enabled}
 monitoring_srv_ip: ${var.monitoring_srv_ip}
 qa_mode: ${var.qa_mode}
 provisioning_log_level: ${var.provisioning_log_level}
+EOF
```
arbulu89, line 51: Here, besides adding the data to the grains, I would add all the variables as we do with the other. Maybe, we could create a map where using `hana` as key we can include all the elements there so they might be used later. I mean something like:

```
output "configuration" {
  value = {
    provider_type = var.provider_type
    ...
    hana = {
      sid = var.hana_sid
      instance_number = var.hana_instance_number
    }
    grains_output = <<EOF
...
```

This way everything is available to be used even by terraform

arbulu89 (commented):
@Simranpal I still think that we can add more variables to the `common_variables`, like
- `hana_cluster_vip`
- `hana_cluster_vip_secondary`
- `ha_enabled`
- `fencing_mechanism`
- `sbd_storage_type`

We can use these variables in the modules again, as we do with the current common_variables.
At the end, the goal is to unify the variables among the providers to avoid duplicated code and issues maintaining this.

generic_modules/common_variables/outputs.tf
```diff
@@ -35,6 +35,27 @@ output "configuration" {
     monitoring_enabled     = var.monitoring_enabled
     monitoring_srv_ip      = var.monitoring_srv_ip
     qa_mode                = var.qa_mode
+    hana                   = {
+      sid = var.hana_sid
```
arbulu89, line 39: This format doesn't look like follows HCL format rules. The `=` character should start in the same place in all lines to improve readability

generic_modules/common_variables/outputs.tf
```diff
@@ -47,7 +68,77 @@ additional_packages: [${join(", ", formatlist("'%s'", var.additional_packages))}
...
+hana_archive_file: ${var.hana_archive_file}
+hana_extract_dir: ${var.hana_extract_dir}
+hana_cluster_vip: ${var.hana_cluster_vip}
+hana_cluster_vip_secondary: ${var.hana_cluster_vip_secondary}
+scenario_type: ${var.scenario_type}
+ha_enabled: ${var.hana_ha_enabled}
+fencing_mechanism: ${var.hana_cluster_fencing_mechanism}
+sbd_storage_type: ${var.sbd_storage_type}
+EOF
+    netweaver = {
```
arbulu89, line 97: I would put this before all grains, so all grains are set in sequence

generic_modules/common_variables/outputs.tf
```diff
@@ -47,7 +68,77 @@ additional_packages: [${join(", ", formatlist("'%s'", var.additional_packages))}
 monitoring_enabled: ${var.monitoring_enabled}
 monitoring_srv_ip: ${var.monitoring_srv_ip}
 qa_mode: ${var.qa_mode}
+hwcct: ${var.hwcct}
```
arbulu89, line 71: `hwcct` should go in `hana` grains, as this tool is just a tool to test HANA performance

generic_modules/common_variables/variables.tf
```diff
@@ -112,3 +112,9 @@ variable "qa_mode" {
   type        = bool
   default     = false
 }
+
+variable "hwcct" {
```
arbulu89, line 116: This should go in `hana_variables.tf` file

libvirt/main.tf
```diff
@@ -43,22 +43,65 @@ locals {
...
+  hana_archive_file                   = var.hana_archive_file
+  hana_extract_dir                    = var.hana_extract_dir
+  scenario_type                       = var.scenario_type
+  hana_cluster_vip                    = local.hana_cluster_vip
+  hana_cluster_vip_secondary          = var.hana_active_active ? local.hana_cluster_vip_secondary : ""
+  hana_ha_enabled                     = var.hana_ha_enabled
+  hana_cluster_fencing_mechanism      = var.hana_cluster_fencing_mechanism
+  sbd_storage_type                    = var.sbd_storage_type
+  netweaver_sid                       = var.netweaver_sid
+  ascs_instance_number                = var.netweaver_ascs_instance_number
```
arbulu89, line 85: Maybe, we should take advantage and add `netweaver` prefix to these variables

libvirt/main.tf
```diff
@@ -43,22 +43,65 @@ locals {
...
+  hana_primary_site                   = var.hana_primary_site
+  hana_secondary_site                 = var.hana_secondary_site
+  hana_inst_master                    = var.hana_inst_master
+  hana_inst_folder                    = var.hana_inst_folder
+  hana_fstype                         = var.hana_fstype
+  hana_platform_folder                = var.hana_platform_folder
+  hana_sapcar_exe                     = var.hana_sapcar_exe
+  hana_archive_file                   = var.hana_archive_file
+  hana_extract_dir                    = var.hana_extract_dir
+  scenario_type                       = var.scenario_type
```
arbulu89, line 78: I would add `hana` prefix to the variable, only to the one in the module, so we have a better understanding

aws/main.tf
```diff
@@ -188,51 +211,32 @@ module "netweaver_node" {
...
+  key_name                   = aws_key_pair.key-pair.key_name
+  security_group_id          = local.security_group_id
+  route_table_id             = aws_route_table.route-table.id
+  aws_credentials            = var.aws_credentials
+  aws_access_key_id          = var.aws_access_key_id
+  aws_secret_access_key      = var.aws_secret_access_key
+  host_ips                   = local.hana_ips
+  hana_data_disk_type        = var.hana_data_disk_type
+  hana_data_disk_size        = var.hana_data_disk_size
+  hana_cluster_vip           = local.hana_cluster_vip
```
arbulu89, line 234: I don't still understand why don't you put `hana_cluster_vip` in the common variables. Is there any reason?
You know that you can get this data in the module like `var.common_variables["hana"]["cluster_vip"]`, right?

arbulu89 (changes requested):
@Simranpal In general looks fine, but I have added some additional comments.

aws/main.tf
Simranpal: I already added in common variables. `hana_cluster_vip` is still used in the hana module terraform code. So I had to pass it to module. 
Eg: https://github.com/SUSE/ha-sap-terraform-deployments/blob/86d8ec8671a680596323c89bbd5b6c407daf2560/aws/modules/hana_node/main.tf#L30

generic_modules/common_variables/outputs.tf
Simranpal: The terraform `fmt` did not format it. But I can do it manually

aws/main.tf
arbulu89: That's not a limitation. You don't need to pass to the module if it is in the common variables.
Please, read my last comment.
You can use `var.common_variables['hana']['cluster_vip']`.
Everything you pass to the common variables is available in module, so you don't need to pass again.

Simranpal: Yes I understand, I was also thinking to keep the variable name short in the terraform code. But it makes sense to use the values from common variables, as is being used elsewhere too

aws/main.tf
```diff
@@ -188,51 +211,30 @@ module "netweaver_node" {
...
+  key_name              = aws_key_pair.key-pair.key_name
+  security_group_id     = local.security_group_id
+  route_table_id        = aws_route_table.route-table.id
+  aws_credentials       = var.aws_credentials
+  aws_access_key_id     = var.aws_access_key_id
+  aws_secret_access_key = var.aws_secret_access_key
+  host_ips              = local.hana_ips
+  hana_data_disk_type   = var.hana_data_disk_type
+  hana_data_disk_size   = var.hana_data_disk_size
+  ha_enabled            = var.hana_ha_enabled
```
arbulu89, line 234: `ha_enabled` is already in the common variables, right?

aws/main.tf
```diff
@@ -133,53 +176,33 @@ module "iscsi_server" {
...
+  route_table_id        = aws_route_table.route-table.id
+  efs_enable_mount      = var.netweaver_enabled == true && var.drbd_enabled == false ? true : false
+  efs_file_system_id    = join("", aws_efs_file_system.netweaver-efs.*.id)
+  aws_credentials       = var.aws_credentials
+  aws_access_key_id     = var.aws_access_key_id
+  aws_secret_access_key = var.aws_secret_access_key
+  s3_bucket             = var.netweaver_s3_bucket
+  host_ips              = local.netweaver_ips
+  virtual_host_ips      = local.netweaver_virtual_ips
+  ha_enabled            = var.netweaver_ha_enabled
```
arbulu89, line 202: `ha_enabled` is already in the common variables, right?

arbulu89 (approved):
@Simranpal Good job!
It looks great!
Please, could you add in the PR main description the changed variables names (`hana_inst_master` for example). Just in the PR description to have it at hand in the future

aws/main.tf
Simranpal: yes I have removed it from the modules now

Simranpal:
@arbulu89 thanks, i have updated the description and done some changes to fix minor issues. I have now tested successfully with deployment in `libvirt` and `azure`

arbulu89:
@Simranpal It looks good! Merge when you finish testing everything

generic_modules/common_variables/outputs.tf
```diff
@@ -35,7 +35,53 @@ output "configuration" {
...
+      cluster_vip_secondary          = var.hana_cluster_vip_secondary
+      ha_enabled                     = var.hana_ha_enabled
+      fencing_mechanism              = var.hana_cluster_fencing_mechanism
+      sbd_storage_type               = var.hana_sbd_storage_type
+    }
+    netweaver = {
+      ha_enabled                = var.netweaver_ha_enabled
+      fencing_mechanism         = var.netweaver_cluster_fencing_mechanism
+      sbd_storage_type          = var.netweaver_sbd_storage_type
+      netweaver_sid             = var.netweaver_sid
```
arbulu89, line 65: I think we could remove `netweaver` prefix from this variables

Simranpal: yes, I have removed from it all the ones with prefix inside this map

PR #584, opened 2020-10-30
Add PackageHub usage for the monitoring server

arbulu89:
Workaround for the `sap-blue-horizon` branch to install the monitoring server packages from PackageHub.
Until we have some better official solution to get the supported packages, PackageHub is the best option.

### DIFF
--- a/salt/os_setup/registration.sls
+++ b/salt/os_setup/registration.sls
@@ -1,8 +1,8 @@
 {% if grains['os_family'] == 'Suse' %}
+{% set arch = grains['osarch'] %}
 
 {% if grains['reg_code'] and (not grains.get('qa_mode') or '_node' not in grains.get('role')) %}
 {% set reg_code = grains['reg_code'] %}
-{% set arch = grains['osarch'] %}
 register_system:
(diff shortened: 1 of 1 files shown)


MalloZup (approved):

salt/os_setup/registration.sls
```diff
@@ -59,6 +59,18 @@ default_sle_module_public_cloud_registration:
...
+# Works for BYOS and PAYG images
+# These packages are only available for SLE15SP1 and SLE15SP2
+packagehub_registration_monitoring:
+  cmd.run:
+    - name: /usr/bin/SUSEConnect -p PackageHub/{{ grains['osrelease'] }}/{{ arch }}
+    - retry:
+        attempts: 3
+        interval: 15
+{% endif %}
+
```
stefanotorresi, line 73: wouldn't it make more sense to put this task in the `monitoring_srv` state, which is already selected for the role?

Simranpal (approved):
Thanks looks good. Maybe we can add a note in monitoring documentation as well, that these packages are installed from `PackageHub`

PR #588, opened 2020-11-03
add shellcheck to CI and apply fixes to provision.sh

stefanotorresi:
https://github.com/koalaman/shellcheck is awesome, but wtf, we're still using shell. :rofl:

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,1 @@
+disable=SC1091
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: focal
 addons:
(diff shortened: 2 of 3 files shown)


arbulu89 (approved):
@stefanotorresi Thanks

I guess that it only adds the check and the recommended changes, right?
Without any change in the behavour

stefanotorresi:
exactly, they were mostly mitigation for common shell scripting traps like variable quoting and such.

MalloZup (approved):
thx ! yeah we should substitute bash to golang code imho but ok another story in another context.

stefanotorresi:
@MalloZup I was just joking, man!! :grimacing:

PR #590, opened 2020-11-07
Patch for netweaver_nfs_share variable usage 

Simranpal:
Update the `netweaver_nfs_share` variable usage in grains entry. 
Reason because in `aws` the grains entry needs quotes  around the value: `"${var.netweaver_nfs_share}"` in order for the yaml parsing correctly.


arbulu89 (changes requested):
@Simranpal I see all of this not needed.
Why don't we just `"` in the `netweaver_nfs_share` output in common variables?
This should just work without so many changes

arbulu89 (approved):
Thanks @Simranpal , This should work

PR #591, opened 2020-11-09
Fix some provision.sh bash malfunctions

arbulu89:
Fix some issues introduced in: https://github.com/SUSE/ha-sap-terraform-deployments/pull/588
With these changes the functionality is reverted following the bash best practices applied before.

- Remove `-x` option from `set` to avoid debugging more. Some info at: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

- Add `-e` to colored messages to escape the characters. https://stackoverflow.com/questions/5947742/how-to-change-the-output-color-of-echo-in-linux

- Remove quoted parameters from some points (VERSION_ID and `set_colored_output`) to have the correct behaviour. These quotes were introducing incorrect usage of variables

- Fix message in QA mode salt execution

### DIFF
--- a/salt/provision.sh
+++ b/salt/provision.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -xeo pipefail
+set -eo pipefail
 # Script to provision the machines using salt. It provides different stages to install and
 # configure salt and run different salt executions. Find more information in print_help method
 # or running `sh provision.sh -h`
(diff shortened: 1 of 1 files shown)


MalloZup (approved):
:headphones:

stefanotorresi (approved):
I'm sorry shellcheck introduced so many regressions!

PR #592, opened 2020-11-10
Run bastion provision without monitoring

arbulu89:
The bastion provisioning was not done if the monitoring machine was not created. This creates issues as some other important states (update machine, authorize ssh keys, etc) are not executed.

Now, `os_setup` is always executed, and states in `bastion` will be only executed if `monitoring` is enabled as we only setup nginx there to proxy grafana dashboards to the bastion machine.

I took advantage and renamed the internal variable `bastion_enabled` to `bastion_count` as it is more meaningful

### DIFF
--- a/azure/modules/bastion/main.tf
+++ b/azure/modules/bastion/main.tf
@@ -1,11 +1,11 @@
 locals {
-  bastion_enabled    = var.common_variables["bastion_enabled"] ? 1 : 0
+  bastion_count      = var.common_variables["bastion_enabled"] ? 1 : 0
   private_ip_address = cidrhost(var.snet_address_range, 5)
 }
 
 
 resource "azurerm_subnet" "bastion" {
-  count                = local.bastion_enabled
+  count                = local.bastion_count
   name                 = "snet-bastion"
   resource_group_name  = var.resource_group_name
   virtual_network_name = var.vnet_name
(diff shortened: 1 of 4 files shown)


MalloZup (commented):
I think that the logic could have somehow conflicts.

E.g we could have customers who wants to have `bastion` without monitoring.

Imho we should have like bastion standalone, and bastion + monitoring.

Imho the logic should be delegated to the monitoring variable, if it is enabled and if the bastion is enabled we do the nginx logic.

what do you think?

arbulu89:
> I think that the logic could have somehow conflicts.
> 
> E.g we could have customers who wants to have `bastion` without monitoring.
> 
> Imho we should have like bastion standalone, and bastion + monitoring.
> 
> Imho the logic should be delegated to the monitoring variable, if it is enabled and if the bastion is enabled we do the nginx logic.
> 
> what do you think?

The change is doing actually what you say here.
Now, the `bastion` machine will run the `os_setup` salt states to configure and initialize the machine (as all of them). Later on, depending if the `monitoring` is enabled or not, it will run what we have in `bastion` that basically is nginx configuration for the monitoring.
This must be the functionality.
The unique thing I can figure it out is to make this distinction in the `bastion` salt code itself, instead of the `top.sls` file, but by now it won't change the behaviour.
Maybe it would be more explicit.
What do you say @MalloZup ?

MalloZup (approved):
ok then thx!

PR #593, opened 2020-11-11
Add provisioning_output_colored variable

arbulu89:
Implementation of a non colored provisioning output. 
It adds the `provisioning_output_colored` variable to decide whether the provisioning scripts output is colored or not.

Until now, this was done using the `qa_mode` option.

This feature is a good go have thing if the project is used in environments such as blue-horizon.
To use ansii output colors in html is not trivial, as you need to translate all the characters by html tags and span entries, and besides that, salt/terraform combo doesn't just put colors in individual lines, so it makes things really difficult. I think going colorless is the best by now.

**INFO: Unfortunately the `SUSEConnect` tool doesn't have the option to disable colors, so this command output always be colored...*

### DIFF
--- a/salt/provision.sh
+++ b/salt/provision.sh
@@ -18,27 +18,37 @@ get_grain () {
 }
 
 log_ok () {
-    NODE=$(hostname)
-    TIMESTAMP=$(date -u)
-    GREEN='\033[0;32m'
-    NC='\033[0m' # No Color
-    echo -e "${GREEN}$TIMESTAMP::$NODE::[INFO] $1 ${NC}"
+    node=$(hostname)
+    timestamp=$(date -u)
+    message="$timestamp::$node::[INFO] $1"
+    if [[ "$(get_grain provisioning_output_colored)" == "true" ]]; then
+      green="\033[0;32m"
+      nc="\033[0m" # No Color
+      echo -e "$green$message $nc"
+    else
+      echo -e "$message"
+    fi
 }
 
 log_error () {
-    NODE=$(hostname)
-    TIMESTAMP=$(date -u)
-    RED='\033[0;31m'
-    NC='\033[0m' # No Color
-    echo -e "${RED}$TIMESTAMP::$NODE::[ERROR] $1 ${NC}"
+    node=$(hostname)
+    timestamp=$(date -u)
+    message="$timestamp::$node::[ERROR] $1"
+    if [[ "$(get_grain provisioning_output_colored)" == "true" ]]; then
+      red="\033[0;31m"
+      nc="\033[0m" # No Color
+      echo -e "$red$message $nc"
+    else
+      echo -e "$message"
+    fi
     exit 1
 }
 
-salt_output_colored () {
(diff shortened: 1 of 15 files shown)


salt/provision.sh
```diff
@@ -18,27 +18,37 @@ get_grain () {
...
+      echo -e "$message"
+    fi
     exit 1
 }
 
-salt_output_colored () {
-    if [[ "$(get_grain qa_mode)" == "true" ]]; then
-        echo "--no-color"
-    else
+provisioning_output_colored () {
```
stefanotorresi, line 47: I guess this is not just the colored one, it's `provisioning_output` in general.

salt/provision.sh
```diff
@@ -18,27 +18,37 @@ get_grain () {
...
 log_ok () {
-    NODE=$(hostname)
-    TIMESTAMP=$(date -u)
-    GREEN='\033[0;32m'
-    NC='\033[0m' # No Color
-    echo -e "${GREEN}$TIMESTAMP::$NODE::[INFO] $1 ${NC}"
+    node=$(hostname)
+    timestamp=$(date -u)
+    message="$timestamp::$node::[INFO] $1"
+    if [[ "$(get_grain provisioning_output_colored)" == "true" ]]; then
```
stefanotorresi, line 24: maybe we should consider extracting `$(get_grain provisioning_output_colored)` in a global variable instead of calling it repeatedly inside many functions.

stefanotorresi (changes requested):
LGTM other than a couple of minor changes to the shell script!

MalloZup (approved):
ok to me thx!

salt/provision.sh
arbulu89: I don't get this.
This function is used to return the salt options about coloring, `--force-color` or `--no-color`. Why do you think that it's not about colored?
Maybe it could be renamed to `salt_output_colored` or `salt_color_flag` as it's exclusive for salt

salt/provision.sh
arbulu89: Having global variables is not that simple here. Have in mind that this script is based in different flags and execution. This means that we don't always execute the same code.
For example, we execute it 2 times, during the regular deployment. First to initialize, and after the reboot to finish the configuration.
During this process, the `grains` file is moved from `/tmp/grains` to `/etc/salt/grains`.
This means that we cannot just initialize the variables at the beginning of the script.
To handle all this complexity, we would eventually need to control more things making the code more complex.
Do you think that it's worthy?

stefanotorresi: uhm, no, absolutely not worth it, let's leave it repeated! :grimacing:

salt/provision.sh
stefanotorresi: Ohh, that wasn't clear at all to me! I thought this function would actually print the output, colored or not.
Then yes, I would appreciate going with `salt_color_flag` to make this clearer.

arbulu89: Updated to `salt_color_flag`.
Thanks for the review!

stefanotorresi (approved):
all good!

PR #595, opened 2020-11-13
Fix Grafana configuration

stefanotorresi:
The latest version of Grafana changes the configuration slightly so the salt state would not match it anymore. I made it more generic so that it shouldn't fail in the future.


MalloZup (approved):
ok to me .

Alternatively we could have a data-plain file managed and then copied from salt but i'm ok with this approach thx!

stefanotorresi:
yeah, I thought about it, but at the end of the day we just need that one line, we don't change anything else, so it's less work this way, to keep it in sync with upstream changes.

PR #598, opened 2020-11-18
blue horizon tuning dashboard and exporter setup

MalloZup:
# Note for reviewers:

review commit by commit so you have less files

# Description:

This pull-request add:

- the new `tuning.json` dashboard which is the saptune_exporter dashboard
- an option for grafana.ini to allow dashboards/panels to be embedded (by default this is disabled by grafana)
- add the saptune_exporter systemd-service activation on hana nodes

### DIFF
--- a/salt/hana_node/saptune_exporter.sls
+++ b/salt/hana_node/saptune_exporter.sls
@@ -1,0 +1,11 @@
+prometheus_ha_cluster_exporter:
+  pkg.installed:
+    - name: prometheus-saptune_exporter
+
+saptune_exporter_service:
+  service.running:
+    - name: prometheus-saptune_exporter
+    - enable: True
+    - restart: True
+    - require:
+      - pkg: prometheus-saptune_exporter
--- a/salt/hana_node/saptune_exporter.sls
+++ b/salt/hana_node/saptune_exporter.sls
@@ -6,6 +6,5 @@ saptune_exporter_service:
   service.running:
     - name: prometheus-saptune_exporter
     - enable: True
-    - restart: True
     - require:
       - pkg: prometheus-saptune_exporter
--- a/salt/hana_node/saptune_exporter.sls
+++ b/salt/hana_node/saptune_exporter.sls
@@ -1,4 +1,4 @@
-prometheus_ha_cluster_exporter:
+prometheus_saptune_exporter:
   pkg.installed:
     - name: prometheus-saptune_exporter
 
--- a/salt/monitoring_srv/grafana.sls
+++ b/salt/monitoring_srv/grafana.sls
@@ -25,7 +25,7 @@ grafana_allow_embedding:
     - require:
       - pkg: grafana
 
-# change default_them to light for better matching blue-horizon colors ;default_theme = dark
+# change theme to light for better matching blue-horizon colors ;default_theme = dark
 grafana_default_color:
   file.line:
     - name: /etc/grafana/grafana.ini
--- a/salt/monitoring_srv/grafana.sls
+++ b/salt/monitoring_srv/grafana.sls
@@ -14,6 +14,17 @@ grafana_anonymous_login_configuration:
     - require:
       - pkg: grafana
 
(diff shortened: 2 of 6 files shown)


salt/hana_node/saptune_exporter.sls
```diff
@@ -0,0 +1,11 @@
+prometheus_ha_cluster_exporter:
```
stefanotorresi, line 1: copy-paste leftover? ;)

stefanotorresi (changes requested):
I'm not sure if we should pursue proper vendoring of the web console assets here.
Committing a bazillion files from the eos-design system when that's already packaged with a node module feels like quite a liability.

@arbulu89 what do you think here? Do we want to go quick and dirty or do we go slow and smooth?

stefanotorresi:
> Committing a bazillion files from the eos-design system when that's already packaged with a node module feels like quite a liability.

to clarify, a cleaner way to go about this would be to include just a package.json file instead of the EOS files, then, via Salt states, install nodejs and npm from an opensource repo, and run `npm install` from the website dir.

salt/monitoring_srv/grafana.sls
```diff
@@ -14,6 +14,27 @@ grafana_anonymous_login_configuration:
     - require:
       - pkg: grafana
 
+# by default grafana will instruct browsers to not allow rendering Grafana in a <iframe>, set value to true
+# this is needed in the blue-horizon static pages
+grafana_allow_embedding:
+  file.line:
+    - name: /etc/grafana/grafana.ini
+    - mode: ensure
+    - after: \[security\]
```
arbulu89, line 23: Is this `after` secure?
Shouldn't we find the current `allow_embedding` entry and replace it.
If there is not this value, I guess we can go with this option

arbulu89 (changes requested):
@MalloZup @stefanotorresi I agree with your approach. Better try to have a neat project installing the styles from a package (even if it's npm).
This way it is much easier to know our current html file.

Besides that, we need to open the new saptune exporter port in the security rules, and maybe the load balancer. Have a look no how the other exporters work

salt/monitoring_srv/grafana.sls
MalloZup: this follow the same pattern we used on already merged state. 

The value is already there but uncommented so is not existing there.

MalloZup: https://github.com/SUSE/ha-sap-terraform-deployments/blob/develop/salt/monitoring_srv/grafana.sls#L12

salt/hana_node/saptune_exporter.sls
MalloZup: done thx

salt/hana_node/saptune_exporter.sls
```diff
@@ -0,0 +1,11 @@
+prometheus_saptune_exporter:
+  pkg.installed:
+    - name: prometheus-saptune_exporter
+
+saptune_exporter_service:
+  service.running:
+    - name: prometheus-saptune_exporter
+    - enable: True
+    - restart: True
```
arbulu89, line 9: @MalloZup I think this `restart` option doesn't exist. Let's remove it.
If you need a restart, there are other alternatives I think

arbulu89 (changes requested):
@MalloZup I think that this won't work yet for Azure. There are thing that are missing about the new exporter:
- Open security ports
- Forward ports in the load balancer (this might not be needed, as the exporter is installed in all nodes, but have a look anyway).
Have a look in how the other exporters work
Please, work based in Azure

salt/hana_node/saptune_exporter.sls
MalloZup: done

arbulu89 (approved):
Thanks @MalloZup

MalloZup:
@stefanotorresi  are you ok with this or do you wants more changes? let me know tia

MalloZup:
follow-up: https://github.com/SUSE/ha-sap-terraform-deployments/issues/601

PR #600, opened 2020-11-19
Fix network restart on SLE15-SP3

ldevulder:
Current code tries to enable the network service which lead to an error on SLE15-SP3 because service is already enabled.
In fact, it's not really needed to enable this service, so better to remove this `Enable:` option.

This PR also force-adds the `CLOUD_NETCONFIG_MANAGE` option, to ensure that there will be no issues if `cloud-init` is used.

Tested on 15-SP2 and 15-SP3 images.

### DIFF
--- a/salt/cluster_node/ha/network.sls
+++ b/salt/cluster_node/ha/network.sls
@@ -18,9 +18,11 @@ cloud-netconfig-azure:
   file.replace:
     - pattern: '^CLOUD_NETCONFIG_MANAGE.*'
     - repl: "CLOUD_NETCONFIG_MANAGE='no'"
+    {% if grains['provider'] != 'libvirt' %}
+    - append_if_not_found: True
+    {% endif %}
 
 network:
   service.running:
-    - enable: True
     - watch:
       - file: /etc/sysconfig/network/ifcfg-eth0
--- a/salt/bastion/nginx.sls
+++ b/salt/bastion/nginx.sls
@@ -16,7 +16,6 @@ nginx_service:
   service.running:
     - name: nginx
     - enable: True
-    - restart: True
     - require:
       - pkg: nginx
       - file: nginx_config_file
--- a/salt/cluster_node/monitoring.sls
+++ b/salt/cluster_node/monitoring.sls
@@ -6,7 +6,6 @@ node_exporter_service:
   service.running:
     - name: prometheus-node_exporter
     - enable: True
-    - restart: True
     - require:
       - pkg: prometheus_node_exporter
       - file: activate_node_exporter_systemd_collector
@@ -46,7 +45,6 @@ promtail_service:
   service.running:
     - name: promtail
     - enable: True
-    - restart: True
     - require:
       - pkg: loki
       - file: promtail_config
--- a/salt/monitoring_srv/grafana.sls
+++ b/salt/monitoring_srv/grafana.sls
@@ -36,7 +36,6 @@ grafana_service:
   service.running:
     - name: grafana-server
     - enable: True
-    - restart: True
     - require:
       - pkg: grafana
       - pkg: grafana_dashboards
--- a/salt/monitoring_srv/loki.sls
+++ b/salt/monitoring_srv/loki.sls
@@ -9,6 +9,5 @@ loki_service:
   service.running:
     - name: loki
     - enable: True
-    - restart: True
     - require:
       - pkg: loki


ldevulder:
During my investigations and tests I found that some sls files have `restart: True` set in `service.running`, but this option doesn't exist (see https://docs.saltstack.com/en/latest/ref/states/all/salt.states.service.html#starting-or-restarting-of-services-and-daemons), and I checked the salt code to be sure.

In fact, as long as the syntax is correct you can add whatever you want, I tested by setting `bidule: True` and it "works" (it does nothing but no error).

So I think that `restart: True` could be removed from these files:
- `/salt/cluster_node/monitoring.sls`
- `/salt/monitoring_srv/grafana.sls`
- `/salt/monitoring_srv/loki.sls'

WDYT? I could do it in this PR or in another.

salt/cluster_node/ha/network.sls
```diff
@@ -18,9 +18,9 @@ cloud-netconfig-azure:
   file.replace:
     - pattern: '^CLOUD_NETCONFIG_MANAGE.*'
     - repl: "CLOUD_NETCONFIG_MANAGE='no'"
+    - append_if_not_found: True
```
MalloZup, line 21: I would add a comment  here why this option exists. namely what you wrote 
`
 ensure that there will be no issues if cloud-init is used.`

MalloZup (commented):
looks good only I would add the comment why we add the `append_if not_found`

MalloZup:
@ldevulder  good catch about the restart true, :+1: sofar I could also not find evidence why we need it.
I would not mind if you remove it here with a separete commits and just do 1 test to see if everything is ok !

salt/cluster_node/ha/network.sls
arbulu89: The is already a comment on top of the file. Have a look to the entire file

arbulu89:
@ldevulder If this happens, it looks like a bug in salt core code. Why should it complain if it's enable and you want to enable? It should do nothing and that's it.
Could you paste the error code here?

salt/cluster_node/ha/network.sls
MalloZup: ok

ldevulder:
> @ldevulder If this happens, it looks like a bug in salt core code. Why should it complain if it's enable and you want to enable? It should do nothing and that's it.
> Could you paste the error code here?

The error is not in salt but in systemd, it's because we try to enable a service (network) that is in fact a link to another service (wicked) and that's not allowed. Should be a change in systemd I think, but I checked and that's normal, not a bug in systemd.

Here the error I have: `Failed to enable unit: Refusing to operate on alias name or linked unit file: network.service`

> Besides that, we cannot include this append_if_not_found: True always. I guess that in libvirt it is not used. It should have a
> conditional to wrap it.
It could be added as it's used only by `cloud-init` and nothing else. I can remove it yes if you really want, but `cloud-init` could also be used in libvirt VM. The idea of force to have this line in the code is also to ensure that if `cloud-init` is activated later (don't know why but technically it could be) the HA cluster will not have the issue documented in the KB note.

salt/cluster_node/ha/network.sls
ldevulder: Yes, I think that the KB note is enough.

ldevulder:
> @ldevulder good catch about the restart true,  sofar I could also not find evidence why we need it.

In any case if a restart is needed it will be done, as it's the default behavior of the `watch` command.

> I would not mind if you remove it here with a separete commits and just do 1 test to see if everything is ok !

Ok, will do it.

salt/cluster_node/ha/network.sls
arbulu89, line 21: @ldevulder I think we need to make this `append_if_not_found` conditional.
This value is only for the cloud providers, not for `libvirt` so I think we don't need to add it if it doesn' exist

arbulu89 (changes requested):

salt/cluster_node/ha/network.sls
ldevulder: Added, but as I explained previously `cloud-init` could be used in `libvirt` also, and added an unused variable in this file doesn't hurt.

arbulu89 (approved):
Thanks @ldevulder

ldevulder:
Last tests shown no issue on `master` branch.

PR #607, opened 2020-11-25
Output hana virtual ip

arbulu89:
This value is consumed by blue-horizon to format the correct grafana iframe urls.

Here an example:
http://%{bastion_ip}:3000/d/EcC4JDFWz2/sap-hana?orgId=1&var-DS_PROMETHEUS=Prometheus&var-node_name=hana01&var-node_ip=%{hana_ip}&var-sid=%{sid}&var-instance_number=00&var-database_name=%{sid}&kiosk

**bastion_ip and sid can be already retrieved**

### DIFF
--- a/azure/outputs.tf
+++ b/azure/outputs.tf
@@ -58,6 +58,10 @@ output "cluster_nodes_public_name" {
   value = module.hana_node.cluster_nodes_public_name
 }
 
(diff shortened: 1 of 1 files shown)


MalloZup (approved):
:black_joker:

PR #609, opened 2020-11-27
Feature/count states

arbulu89:
This PR implements some new logic to be able to parse the current deployment progress by blue-horizon.

Main implementations:
- `count_states.py` creation to count the planned salt states. It runs some `lowstate` options to get the planned states recursevely, and it prints to count
- Change where the salt formulas are installed. Now, they will be installed at the first salt execution (`os_setup`). Otherwise, we are not able to count the states, as the formulas must be present to render all of them.
- Add some meaningful and unique `prints` that are consumed later by blue-horizon to know the state of the progress.


salt/count_states.py
```diff
@@ -0,0 +1,86 @@
...
+    Main method
+    """
+    state_count = 0
+    for state in LOW_STATES:
+        state_count += run_lowstate(state_path=state)
+    for saltenv in SALTENVS:
+        cmd = LOWSTATE_CMD.format(saltenv=saltenv)
+        state_data = execute_command(cmd)
+        current_state_count = len(state_data["local"])
+        LOGGER.debug("States count: %d", current_state_count)
```
Simranpal, line 76: Should this not be "Executed states count:"

salt/count_states.py
```diff
@@ -0,0 +1,86 @@
...
+    cmd = LOWSTATE_SLS_CMD.format(state_path=state_path, saltenv=saltenv, pillar=pillar)
+    state_data = execute_command(cmd)
+    state_count += len(state_data["local"])
+    LOGGER.debug("States count: %d", state_count)
+    state_count += count_inner_states(saltenv, state_data)
+    return state_count
+
+def count_inner_states(saltenv, states):
+    """
+    Count inner states show_low_sls
```
Simranpal, line 50: This will count the states recursively, right? for each state which calls further states. I did not understand the check for `module`

salt/provision.sh
```diff
@@ -139,10 +139,18 @@ os_setup () {
     log_ok "os setup done"
 }
 
+count_states () {
+    # Run count_state.py to print the planned states count
```
Simranpal, line 143: how often/ where is this run from

salt/provision.sh
```diff
@@ -216,6 +225,9 @@ while getopts ":hsopdql:" opt; do
         s)
             excute_bootstrap_salt=1
             ;;
+        c)
+            excute_count_states=1
```
Simranpal, line 229: `excute/execute`

salt/provision.sh
```diff
@@ -248,12 +260,14 @@ fi
...
     bootstrap_salt
     os_setup
+    count_states
     predeploy
     deploy
     run_tests
 else
     [[ -n $excute_bootstrap_salt ]] && bootstrap_salt
     [[ -n $excute_os_setup ]] && os_setup
+    [[ -n $excute_count_states ]] && count_states
```
Simranpal, line 270: `excute/execute`

Simranpal (commented):
@arbulu89
The logic to count states is run periodically, to check the executed and planned states? and  is the state count from `prints `interpreted by blue-horizon somewhere else?
And it doesnt count the `os_setup` states?

I have some below comments:

salt/count_states.py
```diff
@@ -0,0 +1,86 @@
...
+    proc = subprocess.Popen(
+        shlex.split(cmd),
+        stdout=subprocess.PIPE, stdin=subprocess.PIPE, stderr=subprocess.PIPE)
+
+    out, err = proc.communicate()
+    return json.loads(out)
+
+def run_lowstate(state_path, saltenv="base", pillar={}):
+    """
+    Run low state
```
MalloZup, line 34: would be helpful here to add hint what is a "low" state because is not that something clear, is this a term coming from upstream?

salt/count_states.py
```diff
@@ -0,0 +1,86 @@
...
+            if len(state["state.sls"]) > 1:
+                pillar = state["state.sls"][1].get("pillar", {})
+            for mod in mods:
+                state_count += run_lowstate(state_path=mod, saltenv=saltenv, pillar=pillar)
+    return state_count
+
+
+def main():
+    """
+    Main method
```
MalloZup, line 67: I would appreciate here maybe a description on the main algorithm we use for counting the states and rationale.

like 
"""  using LOW_states ( description what is low_state) we count the state ... XXX blabla. describing technically
 This was used mostly initially by blue-horizon webui
"""

generic_modules/salt_provisioner/main.tf
```diff
@@ -64,7 +64,7 @@ resource "null_resource" "provision" {
 
   provisioner "remote-exec" {
     inline = [
-      "sudo bash /tmp/salt/provision.sh -sol /var/log/salt-result.log",
+      "sudo bash /tmp/salt/provision.sh -socl /var/log/salt-result.log",
```
MalloZup, line 67: this command line is becoming quite complex to understand. We could improve readability with a comment like:

something in this direction
```
# execute bootstrap salt installation and os configuration states, then count states 
 "sudo bash /tmp/salt/provision.sh -socl /var/log/salt-result.log",
```

MalloZup (commented):
Hi Xabi, I think I miss the algorithm  in code how do we count the states. I think if we could somehow document this in code with comments , it can improve how we will maintain in future.

A part of that, rest of change are ok. thx

salt/os_setup/packages.sls
```diff
@@ -1,5 +1,20 @@
-iscsi-formula:
+{%- if grains['role'] in ['iscsi_srv', 'drbd_node', 'hana_node', 'netweaver_node'] %}
+install-salt-formulas:
   pkg.installed:
+    - pkgs:
+      {%- if grains['role'] == 'iscsi_srv' %}
+      - iscsi-formula
+      {%- elif grains['role'] == 'drbd_node' %}
+      - drbd-formula`
```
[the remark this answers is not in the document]
arbulu89: Yes, you are right. As we don't deploy drbd in this particular branch I didn't have the error

salt/count_states.py
arbulu89: What's the difference? 
They are not actually executed. This is a salt `lowstate` command, that prints the planned states. There is not any execution

salt/count_states.py
arbulu89: Yes, it does

salt/provision.sh
arbulu89: Once, after `os_setup` is finished

salt/count_states.py
arbulu89: This is salt terminology. A low state is the same as a high state, but only applies to one particular file or salt module.

salt/count_states.py
arbulu89: Anyway, this is are debug logs. I think they do the work I would expect from them

arbulu89:
@MalloZup @Simranpal I fixed the typos and added some additional documentation to some methods

stefanotorresi:
LGTM!
I tried really hard to understand the Salt API we're consuming here, but it gave me a headache. Kudos @arbulu89 for being able to make sense of it.

Only one remark for the future: the provisioning script is starting to become cumbersome, especially now that it's mixed bash/python (which suggests we've reached the limits of what Bash can do for us).
We need to keep this complexity in check, so if you folks have ideas on this regard, let's talk.

arbulu89:
> LGTM!
> I tried really hard to understand the Salt API we're consuming here, but it gave me a headache. Kudos @arbulu89 for being able to make sense of it.
> 
> Only one remark for the future: the provisioning script is starting to become cumbersome, especially now that it's mixed bash/python (which suggests we've reached the limits of what Bash can do for us).
> We need to keep this complexity in check, so if you folks have ideas on this regard, let's talk.

@stefanotorresi Yes, it's true that the provisioning scripts are going out of control...
I just wanted to create a temporary solution that fits our use case for Blue-horizon. Something easy and consumable.
I don't expect to merge this in the `master` branch.
For future designs, we need to re-evaluate this needs and do something much better.

MalloZup (approved):
looks better now with some doc! thx

stefanotorresi:
> I don't expect to merge this in the `master` branch.

Oh, okay then! I admit I was thinking under that assumption.

Simranpal (approved):
Thanks!

PR #610, opened 2020-12-03
Tuning dashboard new version

MalloZup:
Add new dashboard according expected layout

image

### DIFF
--- a/salt/monitoring_srv/grafana/dashboards/tuning.json
+++ b/salt/monitoring_srv/grafana/dashboards/tuning.json
@@ -16,7 +16,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": 6,
-  "iteration": 1606995630740,
+  "iteration": 1607348370168,
   "links": [],
   "panels": [
     {
@@ -64,24 +64,21 @@
         "x": 0,
         "y": 0
       },
-      "id": 4,
+      "id": 7,
       "options": {
         "frameIndex": 1,
         "showHeader": true,
-        "sortBy": []
+        "sortBy": [
+          {
+            "desc": true,
+            "displayName": "CURRENT PROFILE"
+          }
+        ]
       },
       "pluginVersion": "7.3.1",
       "targets": [
         {
-          "expr": "node_uname_info",
-          "format": "table",
-          "instant": true,
-          "interval": "",
-          "legendFormat": "",
-          "refId": "A"
-        },
-        {
-          "expr": "saptune_solution_enabled",
+          "expr": "label_replace(saptune_solution_enabled, \"instance\", \"$1\", \"instance\", \"([^:]*):.*\")\n* on (instance) group_left(nodename) label_replace(node_uname_info, \"instance\", \"$1\", \"instance\", \"([^:]*):.*\")",
           "format": "table",
           "instant": true,
           "interval": "",
@@ -94,28 +91,18 @@
       "title": "Saptune solution enabled",
       "transformations": [
         {
-          "id": "filterFieldsByName",
-          "options": {
-            "include": {
-              "names": [
-                "nodename",
-                "solution_name"
-              ]
-            }
-          }
-        },
-        {
-          "id": "concatenate",
-          "options": {}
-        },
-        {
           "id": "organize",
           "options": {
-            "excludeByName": {},
+            "excludeByName": {
+              "Time": true,
+              "Value": true,
+              "instance": true,
+              "job": true
+            },
             "indexByName": {},
             "renameByName": {
               "nodename": "HOST",
-              "solution_name": "CURRENT PROFILE"
+              "solution_name": "Current Profile"
             }
           }
         }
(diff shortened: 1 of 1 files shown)


arbulu89 (commented):
@MalloZup It looks much much better. Well done.

But... Some personal opinions (as always hehe):
- I would use the next columns order. `host|sap note|description|status`
- I would just put the colored box in the status
- The column names don't have the same structure. Some have just capital letters, others not (check the 1st table). And I would put complete text. `s/SAP NOTE DESC/SAP Note description`

Finally, did you check what we have when the solution/note is not applied? To check the colors and text. Just in case testing hehe

stefanotorresi:
> * I would use the next columns order. `host|sap note|description|status`

IIRC Grafana doesn't allow for customizing the column order, it goes alphabetically by default.

MalloZup:
there are some points we can't address, mostly because as users we are tempted to consider grafana as normal UI.

1) the colors:  grafana colors the field by threshold values. So if 2 or more  metrics have value `1` you can say, colors all 1 to green, 2 to yellow , 3 red. Is not coloring by field   Is in this case not a UI.

2) the solution NOT applied/note is an old problem which is related on how an exporter is build.
  In case of the saptune solution, this could be achieved (https://github.com/SUSE/saptune_exporter/issues/4)
  In case of the saptune note, that could require much more work * or we need to investigate.

The question is about 2, since mostly everything need to be redesigned, I think is not even worth the effort

stefanotorresi:
@MalloZup try to address everything that can be addressed before the end of the sprint and open a followup issue for the remaining point.
We'll most likely need to fix that issue in the MVP, we cannot just postpone this to the superseding project: the MVP should show which SAP notes and solutions are _not applied.

MalloZup:
yes, I will adress the comments one by one and follow-up other things which require more work in follow-up issues, as usual.

thx :+1:

arbulu89:
> The question is about 2, since mostly everything need to be redesigned, I think is not even worth the effort

@MalloZup I didn't understand the answer from `question 2`. Do you mean that we cannot show the notes that `are not` applied?
That would be quite bad... But if don't have any alternative. Right now, the exporter is only sending metrics for applied notes? And not for the `not applied`?

MalloZup:
@arbulu89  regarding the note `not` enabled I think we need to search bit.

```
saptune note list

All notes (+ denotes manually enabled notes, * denotes notes enabled by solutions, - denotes notes enabled by solutions but reverted manually later, O denotes override file exists for note):
	1410736		TCP/IP: setting keepalive interval
			Version 6 from 13.01.2020 
	1557506		Linux paging improvements
			Version 14 from 10.08.2015 
	1680803		Sybase - SAP Adaptive Server Enterprise
			Version 24 from 20.11.2017 
 *	1771258		Linux: User and system resource limits
			Version 5 from 18.06.2018 
	1805750		SYB: Usage of HugePages on Linux Systems with Sybase ASE
			Version 6 from 14.11.2017 
 *	1980196		Setting Linux Kernel Parameter /proc/sys/vm/max_map_count on SAP HANA Systems
			Version 7 from 18.10.2017 
 *	1984787		SUSE LINUX Enterprise Server 12: Installation notes
			Version 35 from 29.11.2019 
	2161991		VMware vSphere configuration guidelines
			Version 26 from 02.12.2019 
 *	2205917		SAP HANA DB: Recommended OS settings for SLES 12 / SLES for SAP Applications 12
			Version 57 from 23.04.2019 
 *	2382421		Optimizing the Network Configuration on HANA- and OS-Level
			Version 36 from 16.01.2020 
 *	2534844		Indexserver Crash During Startup due to Insufficient Shared Memory Segment
			Version 12 from 15.11.2017 
 *	941735		SAP memory management system for 64-bit Linux systems
			Version 11 from 04.05.2018 
	SAP_BOBJ	operating system tuning for SAP Business OBJects (BOBJ)
			Version 0 from 10.03.2017 
```
@angelabriel  is this ok if I would assume that the notes without a `*` are the `not` enabled?

In that case we can have the `not` enabled  note metric

angelabriel:
> @angelabriel is this ok if I would assume that the notes without a `*` are the `not` enabled?
> 
> In that case we can have the `not` enabled note metric, but we need to change a bit the code who is parsing that.

yes, that assumption is true.
There a 2 possibilities:
1. use 'saptune note list' and parse the output
or
2. use the function 'note.GetTuningOptions(NoteTuningSheets, ExtraTuningSheets)'. Here you get a structure, which contains all available note definitions (from saptune and the custom specific). With the knowledge of the 'enabled' notes, you can extract the 'non enabled' notes. 
But as discussed some days ago, the saptune functions were not designed as an external API and may be changed in future releases. So this can be a short term solution for a 'demo', but for a solid solution we should work on the definition of the new saptune API based on 'json' files. Sören is currently working on some definitions and we should setup a meeting/workshop early in January next year to define together all the needed requirements to make it easy to use and stable

stefanotorresi:
> Do you mean that we cannot show the notes that `are not` applied?
> That would be quite bad... But if don't have any alternative. Right now, the exporter is only sending metrics for applied notes? And not for the `not applied`?

@arbulu89 yes, precisely. That's due to the design issue in the exporter mentioned above.

MalloZup:
http://10.162.32.100:3000/d/HzFepKtMz/tuning

MalloZup:
@stefanotorresi @arbulu89 please review this.


Follow-up issue are:

on the exporter side.
- Improve note metric to display not enabled notes SUSE/saptune_exporter#7
- SUSE/saptune_exporter#4

After that we can release the exporter to packagehub. updating jira

arbulu89 (approved):
It looks nice @MalloZup 

I would just rename the table headers to:
`s/Saptune solution enabled/Applied profile`
`s/Current profile/Profile`
`s/SAP Note/Applied SAP notes/`

But the current ones can work too. We will review them in the next group review.
Thanks again

MalloZup:
@arbulu89 I took some names by the presentation we did in past I don't mind to change.

I applied the change on latest commit.

image

MalloZup:
everything is adressed on latest commit a part of the exporter tasks

PR #614, opened 2021-01-15
Use azure service principal variables to authenticate

arbulu89:
Use Azure service principal. More info in: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/service_principal_client_secret

### DIFF
--- a/azure/variables.tf
+++ b/azure/variables.tf
@@ -51,6 +51,28 @@ variable "storage_account_key" {
   type        = string
 }
 
+# The next 4 variables are only introduced to enable the Service Principal usage
+
+variable "subscription_id" {
+  description = "Azure subscription ID"
+  type        = string
+}
+
+variable "client_id" {
+  description = "Azure Service Principal 'appId'"
+  type        = string
+}
+
+variable "client_secret" {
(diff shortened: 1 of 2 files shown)


MalloZup (approved):

azure/variables.tf
```diff
@@ -51,6 +51,28 @@ variable "storage_account_key" {
...
+
+variable "client_secret" {
+  description = "Azure Service Principal 'password'"
+  type        = string
+}
+
+variable "tenant_id" {
+  description = "Azure Service Principal 'tenant'"
+  type        = string
+}
```
stefanotorresi, line 74: hmm, I see some discrepancy with the nomenclature of these fields, could you elaborate on this? Is this an oversight of the upstream azure provider?

arbulu89: Do you mean the description? The names are the same ones used by terraform:
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/service_principal_client_secret

stefanotorresi: ah, I see. Nothing we can do about it then. Carry on!

PR #620, opened 2021-01-28
Fix how the lvm is created in azure with multiple luns

arbulu89:
Fix how the LVM volumes are created in azure when multiple luns are involved.
This is needed for disk configuration such as:

```
hana_data_disks_configuration = {
  disks_type       = "Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS,Premium_LRS"
  disks_size       = "512,512,512,512,64,512"
  caching          = "ReadOnly,ReadOnly,ReadOnly,ReadOnly,ReadOnly,None"
  writeaccelerator = "false,false,false,false,false,false"
  luns             = "0,1,2#3#4#5"
  names            = "datalog#shared#usrsap#backup"
  lv_sizes         = "70,100#100#100#100"
  paths            = "/hana/data,/hana/log#/hana/shared#/usr/sap#/hana/backup"
}
```

Check how 2 LVMs are created for `datalog` (`/hana/data` which uses the 70% of disk size of luns `0,1,2` and `/hana/log` which uses the remaining empty space until 100% usage.

FYI: @petersatsuse

### DIFF
--- a/salt/hana_node/mount/azure.sls
+++ b/salt/hana_node/mount/azure.sls
@@ -18,13 +18,14 @@ hana_lvm_vgcreate_{{ vg }}_azure:
       {%- endfor %}
 
 # Configure the logical volumes
+{%- set lun_count = luns|length %}
 {%- for lv_size in grains['hana_data_disks_configuration']['lv_sizes'].split('#')[vg_index].split(',') %}
 hana_lvm_lvcreate_{{ vg }}_{{ loop.index0 }}_azure:
   lvm.lv_present:
     - name: lv_hana_{{ vg }}_{{ loop.index0 }}
     - vgname: vg_hana_{{ vg }}
     - extents: {{ lv_size }}%FREE
-    - stripes: {{ loop.length }}
+    - stripes: {{ lun_count }}
 
 hana_format_lv_{{ vg }}_{{ loop.index0 }}_azure:
   cmd.run:


stefanotorresi:
@arbulu89 I'm a bit lost with how `,` and `#` are used to compose a little DSL here. Do we have any documentation about this?

arbulu89:
> @arbulu89 I'm a bit lost with how `,` and `#` are used to compose a little DSL here. Do we have any documentation about this?

Here the small piece of documentation we have (https://github.com/SUSE/ha-sap-terraform-deployments/tree/master/azure#hana-configuration)

The code is based in the previous ARM (azure deployment templates) project:
https://github.com/SUSE/azure-resource-manager-sap-solution-templates

The project uses this nomenclature, and as we were asked to replicate the same disk mounting we just copied the idea:
https://github.com/SUSE/azure-resource-manager-sap-solution-templates/blob/master/2Tier/azuredeploy.json#L168

The `#` splits the created volume groups (`vg`), and inside this, depending on the elements separated by `,` we have in the paths variable, there are created the `lvm`s. 

So as example:
```
luns             = "0,1,2#3#4#5"
names         = "datalog#shared#usrsap#backup"
lv_sizes       = "70,100#100#100#100"
paths           = "/hana/data,/hana/log#/hana/shared#/usr/sap#/hana/backup
```

All the entries must have the same amount of elements separated by `#`. 4 in this case.This means 4 `vg` are created.
After that, the `,` says which `lun` is used to created the `lvm` and in which folder is mounted (the path and the extent used by each of the folders).

You must understand properly how `vg` and `lvm` work.

stefanotorresi:
I am indeed quite familiar with how LVM works, but still I couldn't immediately understand the semantics of those variables, so maybe we should add what you just explained to the documentation, unless Azure has a documentation page we can point users at. 😉
Feel free to convert this comment as a follow-up issue if you want to push this out of the door, though.

arbulu89:
> I am indeed quite familiar with how LVM works, but still I couldn't immediately understand the semantics of those variables, so maybe we should add what you just explained to the documentation, unless Azure has a documentation page we can point users at. wink
> Feel free to convert this comment as a follow-up issue if you want to push this out of the door, though.

Hi @stefanotorresi ,
Follow up issue created: https://github.com/SUSE/ha-sap-terraform-deployments/issues/622

We can take with this in the future if we think it's important

arbulu89:
@stefanotorresi @MalloZup @Simranpal Can anyone approve this? Or comment if you have any doubt

stefanotorresi (approved):

PR #623, opened 2021-02-03
install nginx only if monitoring enabled - move condition into bastion/init.sls

jamesongithub:
Currently if you don't have monitoring enabled the bastion role fails with:

```
local:
----------
          ID: states
    Function: no.None
      Result: False
     Comment: No Top file or master_tops data matches found. Please see master log for details.
     Changes:   

Summary for local
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:   0.000 ms
```


arbulu89:
Hi @jamesongithub !
Thanks for the patch.
Let me check on this, I thought that we had this fixed

jamesongithub:
@arbulu89 Could you assign a review? I don't have permission to request a review. thanks..

arbulu89 (approved):
Thanks @jamesongithub ,
You are right, we have this issue. It looks like we did something wrong in: https://github.com/SUSE/ha-sap-terraform-deployments/commit/0e06c76d53a8ef938fc14df9c16fe9fd864cf71d

Thank you for the contribution

PR #627, opened 2021-02-09
Fix aws data provider installation

arbulu89:
Fix AWS data provider installation as the process has been changed. Here the followed steps:
https://docs.aws.amazon.com/sap/latest/general/data-provider-install.html


stefanotorresi (approved):
way to go!

ldevulder:
Manually tested on 15-SP3: OK!

Issue #633, opened 2021-02-11
NW - NFS Mount Point Failure - Follow up


arbulu89:
Follow up of https://github.com/SUSE/ha-sap-terraform-deployments/issues/617

The error on NFS connection happens randomly:

```module.netweaver_node.module.netweaver_provision.null_resource.provision[0] (remote-exec): [ERROR   ] sapcontrol: Command not found.
module.netweaver_node.module.netweaver_provision.null_resource.provision[0] (remote-exec): [ERROR   ] Command 'ip a | grep 172.16.1.21/32' failed with return code: 1
module.netweaver_node.module.netweaver_provision.null_resource.provision[0] (remote-exec): [ERROR   ] retcode: 1
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): [ERROR   ] Command '['mount', '-o', 'defaults', '-t', 'nfs4', '172.16.1.31:/HA1/sapmnt', '/sapmnt']' failed with return code: 32
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): [ERROR   ] stderr: Created symlink /run/systemd/system/remote-fs.target.wants/rpc-statd.service → /usr/lib/systemd/system/rpc-statd.service.
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): mount.nfs4: access denied by server while mounting 172.16.1.31:/HA1/sapmnt
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): [ERROR   ] retcode: 32
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): [ERROR   ] Created symlink /run/systemd/system/remote-fs.target.wants/rpc-statd.service → /usr/lib/systemd/system/rpc-statd.service.
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): mount.nfs4: access denied by server while mounting 172.16.1.31:/HA1/sapmnt
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): [ERROR   ] Command '['mount', '-o', 'defaults', '-t', 'nfs4', '172.16.1.31:/HA1/usrsapsys', '/usr/sap/HA1/SYS']' failed with return code: 32
[...]

PR #634, opened 2021-02-12
Fix nfs connection error during nw installation

arbulu89:
Fix for: https://github.com/SUSE/ha-sap-terraform-deployments/issues/633

The error happens in the `nfs.sls` file. Time to time (rarely in fact), the mount/create folders/unmount process doesn't create the folder permanently. I cannot really say what happens.
The truth is that in failed deployments, the created folder are not there, even when the salt states passed. 
One of my assumptions is that the next removal takes place, as the unmount is not instant, but I cannot assure this.

That's why I have added some different delays. With this change I have not reproduced the error in 10 tries.

FYI: @ab-mohamed

### DIFF
--- a/salt/netweaver_node/nfs.sls
+++ b/salt/netweaver_node/nfs.sls
@@ -29,6 +29,14 @@ wait_until_nfs_is_ready:
 # Initialized NFS share folders, only with the first node
 # Executing these states in all the nodes might cause errors during deletion, as they try to delete the same files
 {% if grains['host_ip'] == grains['host_ips'][0] %}
+# Add a delay to the folder creation https://github.com/SUSE/ha-sap-terraform-deployments/issues/633
+wait_before_mount_sapmnt_temporary:
+  module.run:
+    - test.sleep:
+      - length: 3
+    - require:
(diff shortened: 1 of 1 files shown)


Simranpal (approved):
Thanks, looks okay to me. It is strange that the salt states still pass. 
Regarding the changes I have below suggestions for state name but its okay otherwise.

- `wait_mount_sapmnt_temporary / wait_until_mount_sapmnt_temporary`

- `check_folder_creation / check_sapmnt_creation`

stefanotorresi (approved):
Can't really say it looks good, because you know how much I hate sleep timeouts, but hey, I guess Salt is not giving us enough guarantees here, so we have to do what we have to do!

arbulu89:
> Can't really say it looks good, because you know how much I hate sleep timeouts, but hey, I guess Salt is not giving us enough guarantees here, so we have to do what we have to do!

Yes... 
My guessing is that the `unmount` process finishes before really doing the unmount sometimes (I don't know how the background process works), and as we remove the folder locally later, the folders are gone. 
Unfortunately, this is the best solution I found

PR #640, opened 2021-02-17
Add PackageHub usage to hana nodes to get saptune exporter

arbulu89:
Add PackageHub usage to the Hana nodes deployment. This is needed to get `prometheus-saptune_exporter` from official SUSE channels

### DIFF
--- a/salt/os_setup/registration.sls
+++ b/salt/os_setup/registration.sls
@@ -59,8 +59,11 @@ default_sle_module_public_cloud_registration:
 
 {% endif %}
 
-{% if grains['role'] == 'monitoring_srv' %}
-# Workaround to get the monitoring server packages (prometheus/grafana) from PackageHub
+{% if grains['role'] in ['hana_node', 'monitoring_srv'] %}
+# Workaround to get the monitoring packages from PackageHub
+# - prometheus/grafana family for the monitoring server
+# - prometheus-saptune_exporter for the hana nodes
+# If the packages are already preinstalled in the image, those will have preference
 # Works for BYOS and PAYG images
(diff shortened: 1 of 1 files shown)


salt/os_setup/registration.sls
```diff
@@ -59,8 +59,11 @@ default_sle_module_public_cloud_registration:
 
 {% endif %}
 
-{% if grains['role'] == 'monitoring_srv' %}
-# Workaround to get the monitoring server packages (prometheus/grafana) from PackageHub
+{% if grains['role'] in ['hana_node', 'monitoring_srv'] %}
```
stefanotorresi, line 62: I think we could safely add packagehub to all the nodes, just in case we want to quickly try out some unsupported package. What do you think?

stefanotorresi (commented):
I'm not sure I understand why only those roles are selected.

salt/os_setup/registration.sls
arbulu89: Well, this is the `blue-horizon` branch, where we only use these machines where PackageHub is needed. PackageHub is the only module that comes disabled in PAYG images, and I guess there is some reason behind this, it might has its drawbacks.
`saptune_exporter` is not even installed in Netweaver nodes by now. I guess once we enable this, we will need to add PackageHub there too

MalloZup (approved):

salt/os_setup/registration.sls
MalloZup: I'm ok that we start small not adding packagehub to all nodes but on the constraint we had for the blue-horizon project which is hana only

stefanotorresi: Fair enough! :)

PR #643, opened 2021-02-22
 libvirt: HANA scale-out

Simranpal:


yeoldegrove:
We need a complete rework of this based on the already merged work in the Azure and Openstack providers.
One challenge in the libvirt case will be "shared storage" (e.g. a NFS server) that can be used by each HANA scale-out site.
Here is some work on the Openstack provider that could be used:
https://github.com/SUSE/ha-sap-terraform-deployments/commit/12500dee77417905862238935253bc0928562b5f

yeoldegrove:
feature and implementation will be tracked here: https://github.com/SUSE/ha-sap-terraform-deployments/issues/706

PR #645, opened 2021-02-24
bsc#1179529 - Support use different backing device per node

nick-wang:
DRBD have to use different backing device in GCP to support multiple google disks attached.
Use "by-id: /dev/disk/by-id/google-*" as backing device.

In GCP, it is like /dev/disk/by-id/google--disk-drbd--part1
Need to align pillar file to use the template res_single_vol_v9.j2

Need to align with corresponding pull request of drbd-formula(0.4.1) https://github.com/SUSE/drbd-formula/pull/19


nick-wang:
My bad to select master branch to merge, close and recreate a new pull request

PR #646, opened 2021-02-24
bsc#1179529 - Support use different backing device per node

nick-wang:
DRBD have to use different backing device in GCP to support multiple google disks attached.
Use "by-id: /dev/disk/by-id/google-*" as backing device.

In GCP, it is like /dev/disk/by-id/google--disk-drbd--part1
Need to align pillar file to use the template res_single_vol_v9.j2

Need to align with corresponding pull request of drbd-formula(0.4.1) SUSE/drbd-formula#19

### DIFF
--- a/gcp/modules/drbd_node/salt_provisioner.tf
+++ b/gcp/modules/drbd_node/salt_provisioner.tf
@@ -27,6 +27,7 @@ host_ip: ${element(var.host_ips, count.index)}
 cluster_ssh_pub:  ${var.cluster_ssh_pub}
 cluster_ssh_key: ${var.cluster_ssh_key}
 drbd_disk_device: ${format("%s%s","/dev/disk/by-id/google-", element(google_compute_instance.drbd.*.attached_disk.0.device_name, count.index))}
+drbd_disk_device_list: [${join(", ", formatlist("'/dev/disk/by-id/google-%s'", google_compute_instance.drbd.*.attached_disk.0.device_name))}]
 drbd_cluster_vip: ${var.drbd_cluster_vip}
 fencing_mechanism: ${var.fencing_mechanism}
 sbd_storage_type: ${var.sbd_storage_type}
--- a/pillar_examples/automatic/drbd/drbd.sls
+++ b/pillar_examples/automatic/drbd/drbd.sls
@@ -64,7 +64,9 @@ drbd:
   resource:
     - name: "sapdata"
       device: "/dev/drbd1"
+      {% if grains['provider'] != 'gcp' %}
       disk: {{ drbd_disk_device }}1
+      {% endif %}
 
       file_system: "xfs"
       mount_point: "{{ grains['nfs_mounting_point'] }}/{{ grains['nfs_export_name'] }}"
@@ -73,9 +75,15 @@ drbd:
       nodes:
         - name: {{ grains['name_prefix'] }}01
           ip: {{ grains['host_ips'][0] }}
+          {% if grains['provider'] == 'gcp' %}
+          disk: {{ grains['drbd_disk_device_list'][0] }}-part1
+          {% endif %}
           port: 7990
           id: 1
         - name: {{ grains['name_prefix'] }}02
           ip: {{ grains['host_ips'][1] }}
+          {% if grains['provider'] == 'gcp' %}
+          disk: {{ grains['drbd_disk_device_list'][1] }}-part1
+          {% endif %}
           port: 7990
           id: 2


gcp/modules/drbd_node/salt_provisioner.tf
```diff
@@ -27,6 +27,7 @@ host_ip: ${element(var.host_ips, count.index)}
 cluster_ssh_pub:  ${var.cluster_ssh_pub}
 cluster_ssh_key: ${var.cluster_ssh_key}
 drbd_disk_device: ${format("%s%s","/dev/disk/by-id/google-", element(google_compute_instance.drbd.*.attached_disk.0.device_name, count.index))}
+drbd_disk_device_list: [${join(", ", formatlist("'/dev/disk/by-id/google-%s'", google_compute_instance.drbd.*.attached_disk.0.device_name))}]
```
arbulu89, line 30: I wonder if we need this join thing. What is the output of the `formatlist` itself. Doesn't it return a list already. It looks like we are doing created the list, creating a string, and creating a list again hehe

arbulu89: Ok, yes, it looks like it is the second. The data is provided as a list so drbd can find the disk names of each node.

arbulu89 (approved):
@nick-wang It looks good.
It is a bit confusing to pass the quite similar data twice (first in `drbd_disk_device` and second in `drbd_disk_device_list`, one of the components there is the same), but the solution looks ok.
Let's wait to merge the formula usage together with this one.

gcp/modules/drbd_node/salt_provisioner.tf
nick-wang: Good catch! 
(I did the most easy change here. The terraform/salt grammar are easily be forgot after days, sigh...)
Will do some test to enhance.

nick-wang:
> @nick-wang It looks good.
> It is a bit confusing to pass the quite similar data twice (first in `drbd_disk_device` and second in `drbd_disk_device_list`, one of the components there is the same), but the solution looks ok.
> Let's wait to merge the formula usage together with this one.

`drbd_disk_device` is useless in this case. I leave it here is because it is the only way to show the backing device attached to current node. For example, 
```
drbd_disk_device: /dev/disk/by-id/google-nicktest-disk-drbd-0
drbd_disk_device_list:
- /dev/disk/by-id/google-nicktest-disk-drbd-0
- /dev/disk/by-id/google-nicktest-disk-drbd-1
```
`drbd_disk_device` in other node will be `/dev/disk/by-id/google-nicktest-disk-drbd-1`. While both nodes have the same `drbd_disk_device_list`

gcp/modules/drbd_node/salt_provisioner.tf
nick-wang: Well... Due to the bad network (i am in office today). i replied when only 1 comment showed
Why i use the complex format is because i copied the usage of `host_ips: [${join(", ", formatlist("'%s'", var.host_ips))}]`
I will check if i can improve.

nick-wang: After more test, i think the current code is reasonable, cause it is not `list` type but a list like `string`. Otherwise, terraform will raise the issue:
```
Invalid template interpolation value: Cannot include the given value in a string template: string required.
```

arbulu89:
> > @nick-wang It looks good.
> > It is a bit confusing to pass the quite similar data twice (first in `drbd_disk_device` and second in `drbd_disk_device_list`, one of the components there is the same), but the solution looks ok.
> > Let's wait to merge the formula usage together with this one.
> 
> `drbd_disk_device` is useless in this case. I leave it here is because it is the only way to show the backing device attached to current node. For example,
> 
> ```
> drbd_disk_device: /dev/disk/by-id/google-nicktest-disk-drbd-0
> drbd_disk_device_list:
> - /dev/disk/by-id/google-nicktest-disk-drbd-0
> - /dev/disk/by-id/google-nicktest-disk-drbd-1
> ```
> 
> `drbd_disk_device` in other node will be `/dev/disk/by-id/google-nicktest-disk-drbd-1`. While both nodes have the same `drbd_disk_device_list`

It is fine. I was thinking that we could play maybe with the order (where the 1st element is your current disk), or something like that.
But it is fine as it is. Go with it

nick-wang:
@arbulu89 Thanks a lot for the review and idea

nick-wang:
> I believe it is probably because for some reason the provider variable wasn't set. @nick-wang is this something you face before? Could you take a look here?

I wonder if it is caused by packages version mismatch. Cause the terraform project(not rpm based) didn't force upgrade required related rpms. What we have is a required packages list in release note after bumped up the versions. Maybe a checker is helpful in this case.

PR #648, opened 2021-03-01
Develop

arbulu89:
Merge `develop` in `master` to create the new `7.0.0` version.

Some changes:
- Unify the OS selection to `os_image`
- Set PAYG as default images
- Improve the project to use S4/HANA
- Improve SSH keys management (now, plain SSH keys content can be used too)
- Improve bastion usage for Azure. Now, the monitoring server is bypassed with nginx
- Use the latest terraform azure/gcp/aws resources (matching the ones shipped in SLE)
- And as always, more bugfixing and variables improvement. 

I will detail all the new variables and changes in the release description.

FYI @juadk and @ldevulder There will be some required changes in your openQA tests if you want to adopt this new version (which looks the best as we won't be able to continue adding changes to `6.x.x` codestream)


ldevulder:
> FYI @juadk and @ldevulder There will be some required changes in your openQA tests if you want to adopt this new version (which looks the best as we won't be able to continue adding changes to `6.x.x` codestream)

We will but not until 15-SP4 OS version, as we are close to the RC version of 15-SP3 and there's too much changes in v7 :-)

MalloZup (approved):
:foot:

PR #651, opened 2021-03-08
Avoid loki installation and usage in SLE12 in client side

arbulu89:
As `loki` won't be supported nor used in SLE12 code streams we need to remove the installation and usage, to avoid false errors when `loki` installation fails.

**This only affects to the client side. The whole monitoring server is only supported in SLE15**


stefanotorresi (approved):

arbulu89:
What a mistake... I wanted to merge this in `develop` and I did in `master`

Issue #644, opened 2021-02-23
Prevent GCP environment to be exposed to the internet


ab-mohamed:
Hi, 

I assume that my request has been implemented in the Azure network implementation.

The current network implementation for the GCP HANA, NetWeaver, and DRBD/NFS HA clusters and Monitoring Servers provides access to them over the internet. 

Could you please adjust the current network implementation to:
1.  Deploy the environment instances in a protected private network(s).
2.  Create a bastion host that is exposed to the internet and used to access the whole environment?

Best regards,
Ab

arbulu89:
Hi @ab-mohamed ,
I would need some help here. Do you have some architectural details document where I can base my implementation?
I need to know the network setting (used vpc, network, firewall rules, etc) for a environment based in a bastion host.

arbulu89:
Adding link with information:
https://cloud.google.com/solutions/sap/docs/sap-hana-ha-dm-deployment-sles

arbulu89:
Implemented in: https://github.com/SUSE/ha-sap-terraform-deployments/pull/654

PR #654, opened 2021-03-09
Implement bastion usage for GCP

arbulu89:
Implementation of a bastion usage for GCP. Includes:
- Bastion machine implementation
- Use the correct firewall settings
- Create NAT resources to allow internet connectivity on VMs without public ip
- Change the other modules to use the bastion data

Based on: https://cloud.google.com/solutions/sap/docs/sap-hana-ha-dm-deployment-sles

https://github.com/SUSE/ha-sap-terraform-deployments/issues/644
FYI: @ab-mohamed

### DIFF
--- a/gcp/infrastructure.tf
+++ b/gcp/infrastructure.tf
@@ -29,6 +29,8 @@ locals {
   vpc_name             = var.vpc_name == "" ? google_compute_network.ha_network.0.name : var.vpc_name
   subnet_name          = var.subnet_name == "" ? google_compute_subnetwork.ha_subnet.0.name : var.subnet_name
   subnet_address_range = var.subnet_name == "" ? var.ip_cidr_range : (var.ip_cidr_range == "" ? data.google_compute_subnetwork.current-subnet.0.ip_cidr_range : var.ip_cidr_range)
+
+  create_firewall = ! var.bastion_enabled && var.create_firewall_rules ? 1 : 0
 }
 
 # Network resources: Network, Subnet
@@ -68,7 +70,7 @@ resource "google_compute_firewall" "ha_firewall_allow_internal" {
 }
 
 resource "google_compute_firewall" "ha_firewall_allow_icmp" {
-  count   = var.create_firewall_rules ? 1 : 0
+  count   = local.create_firewall
   name    = "${local.deployment_name}-fw-icmp"
   network = local.vpc_name
 
@@ -78,7 +80,7 @@ resource "google_compute_firewall" "ha_firewall_allow_icmp" {
 }
 
 resource "google_compute_firewall" "ha_firewall_allow_tcp" {
-  count   = var.create_firewall_rules ? 1 : 0
+  count   = local.create_firewall
   name    = "${local.deployment_name}-fw-tcp"
   network = local.vpc_name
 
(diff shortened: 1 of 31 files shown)


gcp/infrastructure.tf
```diff
@@ -29,6 +29,8 @@ locals {
   vpc_name             = var.vpc_name == "" ? google_compute_network.ha_network.0.name : var.vpc_name
   subnet_name          = var.subnet_name == "" ? google_compute_subnetwork.ha_subnet.0.name : var.subnet_name
   subnet_address_range = var.subnet_name == "" ? var.ip_cidr_range : (var.ip_cidr_range == "" ? data.google_compute_subnetwork.current-subnet.0.ip_cidr_range : var.ip_cidr_range)
+
+  create_firewall = ! var.bastion_enabled && var.create_firewall_rules ? 1 : 0
```
Simranpal, line 33: Is the` ! `condition for whole expression or just the `bastion_enabled`

gcp/terraform.tfvars.example
```diff
@@ -124,6 +124,21 @@ pre_deployment = true
 # true or false (default)
 #hwcct = false
 
+##########################
+# Bastion (jumpbox) machine variables
+##########################
+
+# Enable bastion usage. This option configures the deployment to only create a unique public ip address that is attached to the bastion machine
```
Simranpal, line 131: It is bit unclear to me when its says `only create a unique public ip address ..`

Do you mean to say `If this option is enabled, it would create a unique public ip address that ..`

gcp/modules/iscsi_server/main.tf
```diff
@@ -21,8 +26,12 @@ resource "google_compute_instance" "iscsisrv" {
     subnetwork = var.network_subnet_name
     network_ip = element(var.host_ips, count.index)
 
-    access_config {
-      nat_ip = ""
+    # Set public IP address. Only if the bastion is not used
+    dynamic "access_config" {
+      for_each = local.bastion_enabled ? [] : [1]
```
Simranpal, line 31: Should be not `local.bastion_enabled ? "" :  [1]`? in iscsi and monitoring case?

I did not understand the usage of dynamic vs before

Simranpal (approved):
@arbulu89 thanks, I added some comments below

gcp/modules/bastion/main.tf
```diff
@@ -0,0 +1,92 @@
...
+
+# Bastion subnet
+resource "google_compute_subnetwork" "bastion_subnet" {
+  count         = local.bastion_count
+  name          = "${local.deployment_name}-bastion-subnet"
+  network       = var.network_link
+  region        = var.region
+  ip_cidr_range = var.snet_address_range
+}
+
```
Simranpal, line 16: Maybe also add the comment with reference to th GCP guide

gcp/infrastructure.tf
arbulu89: It only applies to the `var.bastion_enabled`. Basically `if not bastion and create firewall set count to 1`

gcp/modules/iscsi_server/main.tf
arbulu89: `for_each` must be a list.
The `dynamic` entry creates the `access_config` content in this case, for each of the elements in the `for_each` list. So if the list is empty (bastion enabled) we won't have the `access_config` entry, which is what we want to avoid having public ip address.
Otherwise, we will have one `access_config` entry with the provided content `nat_ip = ""`. The entry would be:

PR #657, opened 2021-03-10
Develop

arbulu89:
Merge `develop` in `master` to get the `Issues` new template.
It brings other small correction about `loki` usage which is not bad.

I won't create a new new release for this, but it will help with the issues template at least


Simranpal (approved):
Looks good, I guess we don't need release version update after the loki fix

PR #658, opened 2021-03-11
Fix errors in non HA scenarios

arbulu89:
I have found some errors running some deployments in non HA scenarios.

- In Azure and GCP some non created resources are referenced
- In Libvirt there was an incorrect usage of one variable for Netweaver


MalloZup (approved):

Simranpal (approved):
Thanks

PR #660, opened 2021-03-15
Fix hardcoded architecture

mfriesenegger:
There are two lines where x86_64 is hardcoded.  The hardcoded architecture is replaced with a command to determine the architecture at runtime.

### DIFF
--- a/salt/provision.sh
+++ b/salt/provision.sh
@@ -63,9 +63,11 @@ install_salt_minion () {
 
       # Register the modules accordingly with the SLE version.
       if [[ $VERSION_ID =~ ^12\.? ]]; then
-        SUSEConnect -p sle-module-adv-systems-management/12/x86_64
+        # shellcheck disable=SC2046
+        SUSEConnect -p sle-module-adv-systems-management/12/$(uname -m)
       elif [[ $VERSION_ID =~ ^15\.? ]]; then
-        SUSEConnect -p sle-module-basesystem/${VERSION_ID:+"$VERSION_ID"}/x86_64
+        # shellcheck disable=SC2046
+        SUSEConnect -p sle-module-basesystem/${VERSION_ID:+"$VERSION_ID"}/$(uname -m)
       else
         log_error "SLE Product version not supported by this script. Please, use version 12 or higher."
       fi


arbulu89 (approved):
Great @mfriesenegger !

Could you add a short description in the PR description box? This helps to understand the why of this change (even though it is pretty obvious).

We can merge afterwards

**Edit**: Now I see the travis validation is failing. We will need to fix it. You can amend the fix to the same commit if you want

mfriesenegger:
> Great @mfriesenegger !
> 
> Could you add a short description in the PR description box? This helps to understand the why of this change (even though it is pretty obvious).
> 
> We can merge afterwards
> 
> **Edit**: Now I see the travis validation is failing. We will need to fix it. You can amend the fix to the same commit if you want.
> You could simply disable this checker (which I don't like much personally) using `# shellcheck disable=SC2046`. Check in the current code, there are already some examples

Done.  Thank you @arbulu89!

salt/provision.sh
```diff
@@ -63,9 +63,9 @@ install_salt_minion () {
 
       # Register the modules accordingly with the SLE version.
       if [[ $VERSION_ID =~ ^12\.? ]]; then
-        SUSEConnect -p sle-module-adv-systems-management/12/x86_64
+        SUSEConnect -p sle-module-adv-systems-management/12/"$(uname -m)"
```
arbulu89, line 66: Does this work? I would be surprised, but who knows.
I would just add the shellcheck disable line which I commented in the PR on top of the used line, and let as it was before

mfriesenegger: > Does this work? 

I believe it does based on this simple test.

~> cat test.sh 
#!/bin/bash

echo Hello world! I am a "$(uname -m)".

~> bash ./test.sh 
Hello world! I am a x86_64.

> I would be surprised, but who knows. It looks ugly at least.
> I would just add the shellcheck disable line which I commented in the PR on top of the used line, and let as it was before

I will remove the quotes and add shellcheck before each line.

arbulu89: Yep, it turns out that it works fine. Well, anyway, let's disable the check to follow the same convention than in other pieces of code

arbulu89:
Thanks @mfriesenegger 
Merged

Issue #649, opened 2021-03-03
v7: terraform plan asks for NW password even if NW is not enabled in tfvar file


petersatsuse:
v7: terraform plan asks for NW password even if NW is not enabled in tfvar file

arbulu89:
Yes, this is a needed limitation. We don't want to put a default password, because the users wouldn't change it

petersatsuse:
ok, understand, but its from the usage point of view not nice. 
I want to deploy a HANA DB and get asked about the Netweaver password.

arbulu89:
@stefanotorresi @MalloZup 
What do you think of this? I didn't want to put default passwords because it can be a security hole but it creates this bad behaviour. It is a pity that terraform doesn't have conditional variables usage...

stefanotorresi:
Yeah, this is a tricky one. The only thing I can think of right now is to introduce an additional layer to configure the terraform variables via CLI, like you did demonstrate in the past, but it's not an trivial thing to do.

petersatsuse:
what if the default is set to "null" 

It looks like newer version have some improvements here
https://www.hashicorp.com/blog/terraform-0-12-conditional-operator-improvements

ab-mohamed:
Btw, I confirm the same issue when I deploy HANA HA cluster on GCP.

arbulu89:
Hey @ab-mohamed and @petersatsuse ,
I have implemented a solution to avoid this behaviour. It is not really a terraform way, because we are converting the code more and more an imperative style, but well, it is what it is. Terraform is limited in many ways.

arbulu89:
Fixed in: https://github.com/SUSE/ha-sap-terraform-deployments/pull/659

PR #662, opened 2021-03-16
Terraform variables validation

arbulu89:
Some terraform variables validation. This helps to avoid incorrect usage of the variables.
The variables that goes to `common_variables` are validated there to avoid duplication of code.

Mostly it valides:
- IP address and range formats
- Variables with defined options (like multi select options)
- HANA and Netweaver passwords
- Some HANA and Netweaver parameters (sid and instance number)

Besides that, I have implemented a way to avoid asking for the Netweaver password if this is not enabled: https://github.com/SUSE/ha-sap-terraform-deployments/issues/649
It is not a implementation that I really like, because we are using terraform in a quite strange way, but it makes the work.

**Discalimer:** Unfortunately there is not any way to create macros or custom function in terraform, so I have been forced to duplicate a lot of code.


petersatsuse:
A remark: SAP does not allow lowercase SID's, they need to be uppercase, but the validation check allows lowercase.
Reference:
https://help.sap.com/viewer/2c1988d620e04368aa4103bf26f17727/2.0.05/en-US/7672838aa9c44c7ab9d10188531669f3.html

arbulu89:
> A remark: SAP does not allow lowercase SID's, they need to be uppercase, but the validation check allows lowercase.
> Reference:
> https://help.sap.com/viewer/2c1988d620e04368aa4103bf26f17727/2.0.05/en-US/7672838aa9c44c7ab9d10188531669f3.html

Ok, I could change that, but it doesn't really matter. They will be transformed to the correct format afterwards.

Simranpal (approved):
Thanks @arbulu89 I like the validations. We will also need to update the validation text/options as more options are added, something to keep in mind for future.

Regarding the uppercase of SAP sid variables, I think that should be fine as we do the check in lower layer code.  I hope there is no usecase where we required lowercase in salt code

PR #663, opened 2021-03-17
Move DRBD habootstrap template and notify script of exporter to drbd-formula

nick-wang:
Move the DRBD habootstrap formula template and notify script
of ha exporter to drbd-formula.

Should use together with https://github.com/SUSE/drbd-formula/pull/21, drbd-formula >= 0.4.2

As long as this pull request merged. All changes in habootstrap cluster template should 
change in drbd-formula:templates/habootstrap-formula/cluster_resources_nfs_cloud.j2


pillar_examples/automatic/drbd/cluster.sls
```diff
@@ -47,7 +47,7 @@ cluster:
       {% endif %}
     method: 'update'
     template:
-      source: salt://drbd_node/files/templates/drbd_cluster.j2
+      source: /usr/share/salt-formulas/states/drbd/templates/habootstrap-formula/cluster_resources_nfs_cloud.j2
```
arbulu89, line 50: Let's put `salt://drbd/templates/habootstrap-formula/cluster_resources_nfs_cloud.j2` here.
Here `salt` refers to the root_files we have in the salt configuration file, and `/usr/share/salt-formulas/states` is already there

salt/drbd_node/custom_handlers.sls
```diff
@@ -1,5 +1,7 @@
 # install the custom handler for splitbrain
 /usr/lib/drbd/notify-split-brain-haclusterexporter-suse-metric.sh:
   file.managed:
-    - source: salt://drbd_node/files/notify-split-brain-haclusterexporter-suse-metric.sh
+    - source: /usr/share/salt-formulas/states/drbd/templates/ha_cluster_exporter/notify-split-brain-haclusterexporter-suse-metric.sh
```
arbulu89, line 4: The same for `salt` thing: `salt://drbd/templates/ha_cluster_exporter/notify-split-brain-haclusterexporter-suse-metric.sh`

arbulu89 (changes requested):
Thanks @nick-wang Some comments below.
Besides, is the PR removing the whole `files` folder in `drbd_node` folder?

pillar_examples/automatic/drbd/cluster.sls
nick-wang: Thanks, i totally forget this tips.
It looks better to remove the absolute path.

arbulu89 (approved):
Thanks @nick-wang ,
If you tested (at least in 1 provider) and works go ahead and merge

nick-wang:
> Thanks @nick-wang ,
> If you tested (at least in 1 provider) and works go ahead and merge

Yes, i tested this with new drbd-formula package on libvirt (against the latest develop) already.
It works as expected, so i will merge them.

Thanks a lot for the review.

PR #668, opened 2021-04-08
Freeze deployment automation package versions

arbulu89:
Freeze the packages version involved in the deployment automation. This is needed to avoid errors if non backward compatible changes are introduced, specially in the formulas (which is quite common by now).

Some context:
As we don't expect to continue adding new features to this 1st version of the console project, having frozen packages is the best option. This will avoid unexpected errors in the future.
I have noticed that all the versions are stored in our SCC channels, so it gives the option to specify versions.
It is a pretty straightforward change and doesn't imply major consequences as we want to always stay in these versions.

The used versions are the current latest versions in the SLES4SAP product (all the testing has been done with them). When new versions are submitted, the same older versions are used.

### DIFF
--- a/salt/os_setup/packages.sls
+++ b/salt/os_setup/packages.sls
@@ -2,17 +2,20 @@
 install-salt-formulas:
   pkg.installed:
     - pkgs:
+      - salt-shaptools: 0.3.11+git.1605797958.ae2f08a-3.6.1
       {%- if grains['role'] == 'iscsi_srv' %}
-      - iscsi-formula
+      - iscsi-formula: 1.1.1-1.6.1
       {%- elif grains['role'] == 'drbd_node' %}
-      - drbd-formula
-      - habootstrap-formula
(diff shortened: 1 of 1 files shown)


stefanotorresi:
I wasn't aware that this was possible! 😱 
I remember asking about how to install a previous version of some package via `zypper`, and I was told that we don't actually store old binaries: once a new one is built, that's what you get. What am I missing?

arbulu89:
> I wasn't aware this was possible! scream
> I remember asking about how to install a previous version of some package via `zypper`, and I was told that we don't actually store old binaries: once a new one is built, that's what you get. What am I missing?

This might be our SCC repositories specific thing. It stores all the versions. A normal OBS repository only contains the latest one, but our SCC system stores all the submitted versions, I guess for some reason that we unknown (audits, required by customers, etc)

Simranpal (approved):
If this version of project does not need much changes in future, it looks fine. Perhaps, also useful to document in in readme or somewhere for reference

PR #672, opened 2021-04-14
Add cloud provider "OpenStack"

yeoldegrove:
This adds "OpenStack" as a cloud provider.
The setup is comparable to the azure/gcp deployment.

The deployment of all components (bastion, monitoring, iSCSI, DRBD, HANA, Netweaver/S4HANA), was tested and verified by me via the "bastion host" option.

additional new feature (only for OpenStack so far):
- use bastion host as NFS server for SAP install sources.

prerequisites:
- fix for terraform-proivder-libvirt in #673 / e206f43 to pass build


arbulu89:
Thanks a lot @yeoldegrove 
This looks really promising. Some initial things:
- I need to see if this can go directly to `master` or it must go to `develop` first
- I will do the review the next days.
I will come to you back! And I repeat, really appreciated

salt/drbd_node/files/templates/drbd_cluster.j2
```diff
@@ -38,7 +38,7 @@ primitive fs_{{ res.name }} ocf:heartbeat:Filesystem \
   options=noatime,nodiratime \
   op monitor interval="20" timeout="40s"
 
-{%- if cloud_provider not in ["amazon-web-services", "google-cloud-platform"] %}
+{%- if cloud_provider not in ["amazon-web-services", "google-cloud-platform", "openstack"] %}
```
arbulu89, line 41: This files have been removed in our `develop` branch, so this will create some conflict:
https://github.com/SUSE/ha-sap-terraform-deployments/pull/663/files and https://github.com/SUSE/drbd-formula/pull/21

Besides that, I think this change doesn't make sense. With this, you are saying that you won't have any virtual ip address in the cluster (even though you have added the value in the pillar file).
This `if` condition won't do anything for openstack by now (as it doesn't have any `elif` at the end)

salt/hana_node/hana_inst_media.sls
```diff
@@ -9,7 +9,7 @@ hana_inst_directory:
     - name: {{ grains['hana_inst_folder'] }}
     - mode: "0755"
     - makedirs: True
-  {% if grains['provider'] == 'libvirt' %}
+  {% if grains['provider'] == 'libvirt' or grains['provider'] == 'openstack' %}
```
arbulu89, line 12: I would prefer to have something like:
```{% if grains.get('provider') in ['libvirt', 'openstack'] %}```
But both work. We can change in the future if we add more things

File: salt/os_setup/packages.sls [hunk 3]
```diff
@@ -3,3 +3,19 @@ iscsi-formula:
     - retry:
         attempts: 3
         interval: 15
+
+# iscsi kernel modules are not available in kernel-default-base
```
arbulu89, line 7: Can I ask why you have added this? We never had this issues. Do you think if this is compatible with the other cloud providers?

arbulu89 (commented):
Hi @yeoldegrove ,

Fantastic work! It perfectly follows all the structure we created without almost any invasive change.
I have raised some minimal comments that I would prefer to discuss before merging.

My unique concern is if we will find any issue moving this back to develop to avoid conflicts in the future.

salt/os_setup/packages.sls
yeoldegrove: IMHO `kernel-default-base` is never a good idea unless you're really not wanting a lot of kernel modules to be available.
I had the issue that my OpenStack base image (no idea if it was an "SUSE official" image) had this installed. It bugged me so hard that it went in here.
A cloud-init `userdata` file would be an alternative to mess with the image before running salt but this is the more general approach.
It should not interfere with other cloud providers. The `require` section in `kernel-default` should make sure that it is only installed if `kernel-default-bae` was removed.

salt/drbd_node/files/templates/drbd_cluster.j2
yeoldegrove: > Besides that, I think this change doesn't make sense.

Your are right. Is is a development artefact.
fixed in 797539b

yeoldegrove:
@arbulu89 is a rebase against your `develop` needed and will you handle that?

salt/hana_node/hana_inst_media.sls
yeoldegrove: fixed in c0ec2c5

arbulu89:
> @arbulu89 is a rebase against your `develop` needed and will you handle that?

Yes, we will need to rebase this later to develop, but I will take care. I think we can merge the currently opened 2 PRs, and create a new version `7.1.0`. I will do the next

arbulu89 (approved):
@yeoldegrove Great job as commented. 
I will merge the PR if you don't plan to add anything else

yeoldegrove:
> @yeoldegrove Great job as commented.
> I will merge the PR if you don't plan to add anything else

You're welcome... finally I can easily test all the SLES SAP/HA-Products and showcase them ;)
Nothing to add from my side for now.

PR #674, opened 2021-04-15
libvirt: Draft/shared storage architecture for Hana scaleout and Netweaver 

Simranpal:
The motivation behind this architecture change is to have one terraform module for providing shared storage to all type of deployments. Depending on the type of deployed application(SAP Hana scaleout, NW etc) user can choose the type of needed shared storage, and at the same time configure the storage parameters like disk, cpu resources etc. This configuration is needed as resources will be different on type of deployed application. 
Currently the project supports NFS and Drbd provided storage options.
The changes are not complete and I create this draft for time being while testing new additional changes

### DIFF
--- a/libvirt/main.tf
+++ b/libvirt/main.tf
@@ -105,7 +105,7 @@ module "common_variables" {
   netweaver_swpm_sar                  = var.netweaver_swpm_sar
   netweaver_sapexe_folder             = var.netweaver_sapexe_folder
   netweaver_additional_dvds           = var.netweaver_additional_dvds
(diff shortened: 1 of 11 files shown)


yeoldegrove:
We need a complete rework of this based on the already merged work in the Azure and Openstack providers.
One challenge in the libvirt case will be "shared storage" (e.g. a NFS server) that can be used by each HANA scale-out site.
Here is some work on the Openstack provider that could be used:
https://github.com/SUSE/ha-sap-terraform-deployments/commit/12500dee77417905862238935253bc0928562b5f

yeoldegrove:
feature and implementation will be tracked here: https://github.com/SUSE/ha-sap-terraform-deployments/issues/706

Issue #669, opened 2021-04-09
SAP deployment failure on SLES12 - Incorrect python version


ab-mohamed:
**Used cloud platform**
GCP

**Used SLES4SAP version**
SLES4SAP 12 SP5 

**Used client machine OS**
SLES15 SP2

**Expected behaviour vs observed behaviour**
This is a fresh  HANA HA cluster deployment on SLSES4SAP 12 SP5.
During the deployment phase using the `terraform apply` command, copying HANA installation media to the HANA nodes using `gsutil` command was failed and it showed the following messages:

```
module.hana_node.module.hana_provision.null_resource.provision[0] (remote-exec): ----------
module.hana_node.module.hana_provision.null_resource.provision[0] (remote-exec):           ID: download_files_from_gcp
module.hana_node.module.hana_provision.null_resource.provision[0] (remote-exec):     Function: cmd.run
module.hana_node.module.hana_provision.null_resource.provision[0] (remote-exec):         Name: /root/google-cloud-sdk/bin/gsutil -m cp -r gs://sap-automation-media/HANA/2.0/SPS04/51054413/* /sapmedia/HANA
module.hana_node.module.hana_provision.null_resource.provision[0] (remote-exec):       Result: False
module.hana_node.module.hana_provision.null_resource.provision[0] (remote-exec):      Comment: Command "/root/google-cloud-sdk/bin/gsutil -m cp -r gs://sap-automation-media/HANA/2.0/SPS04/51054413/* /sapmedia/HANA" run
module.hana_node.module.hana_provision.null_resource.provision[0] (remote-exec):      Started: 08:21:56.473951
module.hana_node.module.hana_provision.null_resource.provision[0] (remote-exec):     Duration: 530.766 ms
module.hana_node.module.hana_provision.null_resource.provision[0] (remote-exec):      Changes:
```

**How to reproduce**
After the deployment was failed, I executed the command manually from one of the HANA nodes:
```
demo1-hana01:~ # /root/google-cloud-sdk/bin/gsutil -m cp -r gs://sap-automation-media/HANA/2.0/SPS04/51054413/* /sapmedia/HANA
WARNING:  Python 3.4.x is no longer officially supported by the Google Cloud SDK
and may not function correctly.  Please use Python version 2.7.x or 3.5 and up.

If you have a compatible Python interpreter installed, you can use it by setting
the CLOUDSDK_PYTHON environment variable to point to it.

[...]

PR #676, opened 2021-04-20
Correct python version for gcloud and gsutil utilities

arbulu89:
Use correct python version for gcloud and gsutil utilities

Fix for: https://github.com/SUSE/ha-sap-terraform-deployments/issues/669


diegoakechi (approved):
LGTM

Issue #671, opened 2021-04-14
Enhancement: required package list for the project


nick-wang:
The terraform project is not RPM based. However, it did required the packages set with the minimized version. We only mark this in release note after bumped up the tags. It is better to show this in projects as well.

Two options can be used for this:
1) Implemented in salt code, like https://github.com/SUSE/ha-sap-terraform-deployments/commit/dbb6fcdd60d1c3f21b7cc23581f0c5578f630708
However, the side effect is user/develper may don't know this file and forget to update when needed.
2) Add a `requirements.txt` in top level, without force install/check. Further more, could implement a script after installation to check the environment with the `requirements.txt`

nick-wang:
Some ideas for the solution:
Create a plain text listed packages/versions like `requirements.txt` in top level.

Then three methods to use this file:
1. Compare the packages after predeploy stage
Mark ERROR in logs if installed packages too old, WARN if list packages not installed. 
In this case, doesn't need to check the node role of var flag with corresponding packages. (eg. drbd-formula not in hana node)
No matter the result, only record it in log file. So that user still can add repo/install new versions during deploy stage.

2. WIthout calling the package check after predeploy stage. Only implement a checker to get logs and environment when needed to ease the debug. 
In this case, we can add more things into the checker than just check package versions.

3. Find a way to get the version from the requirements.txt into all salt state of pkg.installed in salt of ha-sap-terraform project

nick-wang:
After discussed with @arbulu89 we agree to use solution like this:
+ Create a `requirements.txt` to list all the required packages with `exactly` version needed
+ If `ha_sap_deployment_repo` not provided in `terraform.tfvars`. Install the exactly version packages based on the file before `predeply` stage.
   Just install the latest version during `prodeploy` stage if `ha_sap_deployment_repo` provided.

Notes:
+ Need to update `requirements.txt` when required packages changed, before SCC released. Use `ha_sap_deployment_repo` with corresponding OBS project `network:ha-clustering:sap-deployments:xxx` is needed. Version specifc.
+ Need to mentioned the usage/change of `requirements.txt` to doc file for all contributors.

arbulu89:
Implemented at: https://github.com/SUSE/ha-sap-terraform-deployments/pull/678

PR #678, opened 2021-04-21
Requirements file usage implementation

arbulu89:
Implementation of the usage of a `requirements.txt` file to get specific versions for the specified packages.

It works with the following way:
1. If `ha_sap_deployments_repo` is specified it doesn't make any difference from the current usage
2. Otherwise, it will take the content of the `requirements.txt` file in the root folder of the project, and it tries to install them
3. If something goes wrong it displays a notification state to help fixing the issue.

The worst thing is that this will affect to all the modules and install all the packages in all of them even though they are not used.
Besides that, the `requirements.txt` file must be always in the root folder of the project. It doesn't work otherwise. If the file doesn't exist or it is empty the installation is skipped.

Implements: https://github.com/SUSE/ha-sap-terraform-deployments/issues/671

Pending: Create a short bash script to get the packages from a OSC repository and format them to a requirements file format.


requirements.txt
```diff
@@ -0,0 +1 @@
+python3-shaptools: 0.3.12+git.1619007514.1951d23
```
nick-wang, line 1: we need more packages than just `python3-shaptools`. 
I think we need the package list of [Breaking changes
](https://github.com/SUSE/ha-sap-terraform-deployments/releases/tag/7.1.0)

salt/os_setup/requirements.sls
```diff
@@ -0,0 +1,20 @@
+{% if grains['pkg_requirements'] and not grains['ha_sap_deployment_repo'] %}
```
nick-wang, line 1: I think `{% if not grains['ha_sap_deployment_repo'] %}` is enough. We won't get rid of the `grains['pkg_requirements']` anymore if we decide to use this method.

salt/os_setup/requirements.sls
```diff
@@ -0,0 +1,20 @@
...
+install_package_requirements:
+  pkg.installed:
+    - pkgs:
+{% for package in grains['pkg_requirements'] %}
```
nick-wang, line 5: Likely we only able to add the most basic packages for all nodes here. For example, `grafana-xxx` and `sapnwbootstrap-*` may not suitable for DRBD nodes, vice versa.

I was thinking different method to for this state file:
1. Create `requirements.txt` as yaml file, which can be imported to salt directly.
```
pkgs:
  python3-shaptools: 0.3.12+git.1619007514.1951d23
  drbd-formula: 0.4.1+git.1614576184.d8e78f1-1.1
  ...
```
2. Check `grains['ha_sap_deployment_repo']` to decide include/run this `requirements.sls` or not.
3, `requirements.sls` import the yaml file to get versions. And `requirements.sls` is grains['role'] specific, each role installed its specific packages. And the package like `python3-shaptools` could existed in multiple nodes

How do you think of it?

nick-wang (changes requested):
Idea about the implementation to discuss.

requirements.txt
arbulu89: Yes, this was just an example. The PR is in draft mode. I pretended to add all the required packages later on

salt/os_setup/requirements.sls
arbulu89: This just prevents the code going inside if `pkg_requirements` comes with an empty list. If this happens it would render an incorrect state making the execution fail

salt/os_setup/requirements.sls
arbulu89: I don't really like the idea of importing the `pkgs` entry as it is to salt.  We would need to implement the copy of the file from local to the remote machines (this is not complex in any case) and a way to include the file content as it is in the `requirements.sls` state file, which I don't this is possible. That's why we use grains, because they are natively imported in the salt execution. And if they are in the grains file, they must follow a correct structure.

I will definitely check how to do the role based installation (I already had this plan)

salt/os_setup/requirements.sls
nick-wang: yes, i mean the package list should always not empty, unless one day there is no dependencies, which seems impossible.
so it is likely development bug if the package is empty.

salt/os_setup/requirements.sls
nick-wang: > I don't really like the idea of importing the `pkgs` entry as it is to salt. We would need to implement the copy of the file from local to the remote machines (this is not complex in any case) and a way to include the file content as it is in the `requirements.sls` state file, which I don't this is possible. That's why we use grains, because they are natively imported in the salt execution. And if they are in the grains file, they must follow a correct structure.

hmm... i still think format the plain list in terraform to grains, then get them in `salt` is not perfect. And the grains is long enough already. If the `package: version` pair only used in one state, maybe we can avoid this to grains file. So i prefer to use `salt` supported `yaml` file to import directly.

The yaml format isn't hard to maintain/edit. The `pkgs` entry (for example) need to be imported only in the `requirements.sls`. 
About how to copy the yaml file to host. I was thinking put the yaml file in `salt` folder `salt/os_setup` with `requirements.sls`, then create a link named `requirements.txt` or `requirements.yaml` in top level of project. 
 

> 
> I will definitely check how to do the role based installation (I already had this plan)

arbulu89:
Hey @nick-wang ,
I have reimplemented the requirements.yml file usage. Now it has roles included.
As you suggested, this file is a yml file. In the terraform common this, yaml data is included in the grains (I know, you don't really like it, but it is super simple). After that, the data is used in the `requirements.sls` to get the packages per role.
It looks pretty clean. Have a look.

The versions are based on: https://build.opensuse.org/project/show/network:ha-clustering:sap-deployments:v7 (that are the same than in SLE, or at least the MU are on going)

nick-wang:
> I have reimplemented the requirements.yml file usage. Now it has roles included.
> As you suggested, this file is a yml file. In the terraform common this, yaml data is included in the grains (I know, you don't really like it, but it is super simple). After that, the data is used in the `requirements.sls` to get the packages per role.
> It looks pretty clean. Have a look.
> 

@arbulu89 Haha, it doesn't matter whether i like it or not, as long as we find a suitable solution. I am not a guy stick to one way. It looks clean in a glance.
One idea in the glance, maybe we can get the common packages necessary for all roles to shrink the long list, like `python3-shaptools`) It was me asked a role specific list, now i am asking a common list :( 

Anyway, i will check it tomorrow.

arbulu89:
> > I have reimplemented the requirements.yml file usage. Now it has roles included.
> > As you suggested, this file is a yml file. In the terraform common this, yaml data is included in the grains (I know, you don't really like it, but it is super simple). After that, the data is used in the `requirements.sls` to get the packages per role.
> > It looks pretty clean. Have a look.
> 
> @arbulu89 Haha, it doesn't matter whether i like it or not, as long as we find a suitable solution. I am not a guy stick to one way. It looks clean in a glance.
> One idea in the glance, maybe we can get the common packages necessary for all roles to shrink the long list, like `python3-shaptools`) It was me asked a role specific list, now i am asking a common list :(
> 
> Anyway, i will check it tomorrow.

Well, not all the nodes need `python-shaptools`. The iscsi server, monitoring server and the bastion server don't need it.
In any case, this file shouldn't be touched by the users and it doesn't require that much maintenance. Only update the package versions when new MU are done to the involved packages. I have included only the formulas by now, as I don't see the need to have the exporters and other packages here. But we will continue checking this.

nick-wang:
> Well, not all the nodes need `python-shaptools`. The iscsi server, monitoring server and the bastion server don't need it.
OK.

> In any case, this file shouldn't be touched by the users and it doesn't require that much maintenance. Only update the package versions when new MU are done to the involved packages. I have included only the formulas by now, as I don't see the need to have the exporters and other packages here. But we will continue checking this.

The change looks good to me now. We just need record this to document as well, for "formula maintainers" acknowledge the change in this file after bump up the version and submit package to `ha_sap_deployment_repo`.

nick-wang (approved):

PR #679, opened 2021-04-22
fix(prometheus): corosync service not checked

faust64:
The pacemaker unit is checked twice. Corosync however, is not checked.


arbulu89:
@faust64 Thanks for the fix! It looks good
Let's move the PR to `develop` branch. I plan to create a new release that will hit `master` this or the next week. Just to avoid unnecessary rebasing and merging

faust64:
Allright, I've cherry-picked that commit into another branch, based on the develop one.
Let's close that one then.

PR #680, opened 2021-04-22
Fix prometheus rules develop

faust64:
cherry-picking commit originally submitted in https://github.com/SUSE/ha-sap-terraform-deployments/pull/679

The pacemaker unit is checked twice. Corosync however, is not checked.


faust64:
pebkac / wrong target branch.

PR #682, opened 2021-04-22
add AlertManagerWebUI inbound tcp port 9093 to security group

kevinayres:
Alertmanager can be viewed/managed via UI. adding allowed inbound TCP port 9093


arbulu89 (changes requested):
Hi @kevinayres ,
Thank you for the change.
1. Could you move the PR to `develop` branch? (you don't need to open a new one, in the main PR view, you can change the destination branch).
2. Rename the `alertmanagerWebUI` to `alertmanager_server` to keep consistency

Issue #683, opened 2021-04-22
GCP S/4HANA deployment failure


ab-mohamed:
**Used cloud platform**
GCP

**Used SLES4SAP version**
SLES15SP2 for SAP Applications 

**Used client machine OS**
SLES15SP2

**Expected behaviour vs observed behaviour**
The expected delivery is a successful S/4HANA deployment but it was failed

**How to reproduce**
1. Use S/4HANA 1909.
2. Use the default S/4HANA Terraform configurations
3. Deploy the environment, as usual, using `terrafrom init`, `terraform plan`, and `terraform apply --auto-approve` commands.
4. I had the following deployment error messages:
```
module.netweaver_node.module.netweaver_provision.null_resource.provision[2] (remote-exec): Thu Apr 22 13:54:44 UTC 2021::demo1-netweaver03::[ERROR] deployment failed 
Error: error executing "/tmp/terraform_1501732074.sh": Process exited with status 1
```

**Troubleshooting**
1. SAP HANA HA cluster deployment was completed successfully.
2. The DRBD HA cluster deployment was completed successfully. 
3. The S/4HANA deployment was failed.
4.  Even if the NFS shares were exported to the ASCS and ERS nodes, `sapcontrol` command was not found. 
```
demo1-netweaver01:~ # df -ht nfs4
Filesystem                Size  Used Avail Use% Mounted on
10.0.1.22:/HA1/sapmnt      10G   42M   10G   1% /sapmnt
10.0.1.22:/HA1/usrsapsys   10G   42M   10G   1% /usr/sap/HA1/SYS
10.0.1.22:/HA1/ASCS        10G   42M   10G   1% /usr/sap/HA1/ASCS00
```
```
demo1-netweaver01:~ # su - ha1adm
demo1-netweaver01 /home/ha1adm% sapcontrol -nr 00 -function GetProcessList
sapcontrol: Command not found.
```
5. The NFS shares have no files:
```
demo1-netweaver01:~ # ls /sapmnt/ /usr/sap/HA1/SYS/ /usr/sap/HA1/ASCS00/
/sapmnt/:

/usr/sap/HA1/ASCS00/:

/usr/sap/HA1/SYS/:
``` 
6. The ASCS and ERS resource agents were not running due to the `sapcontrol` commands issue:
```
demo1-netweaver01:~ # crm resource status rsc_sap_HA1_ASCS00 rsc_sap_HA1_ERS10
[...]

PR #689, opened 2021-04-30
S/4HANA Master Password

ab-mohamed:
S/4HANA master password must be 10 characters at least. 
Refer to `https://github.com/SUSE/ha-sap-terraform-deployments/issues/683#issuecomment-826701299`


arbulu89 (changes requested):
Thanks @ab-mohamed 
This looks a generic change and it should be extended to all providers, not only `gcp`.

Could you do that?

yeoldegrove:
fixed in #698

PR #699, opened 2021-05-06
hana: allow to set ignore min mem check

rtorrero:
This PR adds a terraform variable to allow to ignore the min memory limitations in hana. This is specially useful for development environments which rely on the `libvirt` terraform provider.

The issue was reported in https://bugzilla.suse.com/show_bug.cgi?id=1185643


pillar_examples/automatic/hana/hana.sls
```diff
@@ -80,6 +84,8 @@ hana:
         {% endif %}
         system_user_password: {{ grains['hana_master_password'] }}
         sapadm_password: {{ grains['hana_master_password'] }}
+        extra_parameters:
```
arbulu89, line 87: We should have the `if` condition even here before the `extra_parameters` entry

arbulu89 (changes requested):
@rtorrero Nice!
Some few details:
- Change the PR from `master` to `develop`
- We need to expand this to the other providers (gcp, azure, aws, openstack). You simply need to copy/paste what you did for them
- Add the new variable example in the `terraform.tfvars.example`, under the HANA chapter. You can put more or less the same description ot the variable. It turns out that the users don't really check the variables.tf content, so we need to double the content in the examples files. THis should be done for all the providers too

pillar_examples/automatic/hana/hana.sls
rtorrero: Totally right, I forgot about the second appearance of the `extra_parameters`. I'll fix it.

rtorrero: Fixed.

rtorrero:
I've updated the PR based on your input, thanks!

aws/terraform.tfvars.example
```diff
@@ -154,6 +154,9 @@ cluster_ssh_key = "salt://sshkeys/cluster.id_rsa"
 # Enable system replication and HA cluster
 #hana_ha_enabled = true
 
+# Disable minimal memory checks for hana. Useful to deploy development clusters.
```
arbulu89, line 157: I would put this `hana` on the comment in capital letters, to keep the same syntax -> `HANA`

arbulu89 (changes requested):
Hey @rtorrero ,
I noticed that one part is missing here. We are not passing the variable to the grains. For that we need to the next:
1. Add the variable in the outputs. Here: https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/generic_modules/common_variables/outputs.tf#L43
2. Add the variable in the output string. Here: https://github.com/SUSE/ha-sap-terraform-deployments/blob/master/generic_modules/common_variables/outputs.tf#L110

Finally, add the same `if` condition for the cost-optimized installation too:
https://github.com/SUSE/ha-sap-terraform-deployments/blob/34a3d5f06238f00ea9a278d7577366fe4845ae37/pillar_examples/automatic/hana/hana.sls#L121

aws/terraform.tfvars.example
rtorrero: Done!

rtorrero:
I've updated the PR again based on your input @arbulu89 , thanks!

arbulu89 (approved):
@rtorrero Great! Nice job.
If you have tested and works it is ready to be merged.

rtorrero:
Did one last change to warn about possible errors when deploying on a host with less than 16 GiB RAM. Merging as the tests have show it works as expected.

PR #701, opened 2021-05-12
Deploy shared storage, HANA HSR, SLE HA

mfriesenegger:
This PR enables the deployment of multiple IBM PowerVS instances where HANA will be installed.  A shared SCSI LUN is also created so that HANA System Replication and SLE HA can be setup with with the SAPHanaSR resource agent.


arbulu89 (approved):
Hi @mfriesenegger ,

Add a description to the PR. It helps to understand the changes before going to the code changes.

PR #705, opened 2021-05-25
Update promtail installation

stefanotorresi:
Update promtail installation according to the latest changes to the RPM package (https://build.opensuse.org/request/show/893918).


arbulu89 (approved):
Thanks @stefanotorresi ,
This was causing the failure or our jenkins CI too

Issue #611, opened 2020-12-04
Grafana: Netweaver Dashboard do not work


picoroma:
I'm testing the deployment of SAP NW750 + Monitoring (Grafana & Prometheus)
Deployment finishes well. No errors.
Everything works fine, but the GRAFANA Dashboard related to SAP Netweaver does not work.
Show No-data
HANA Dashboard its Ok

Checking into Prometheus page I have ALL the Netweaver endpoint DOWN (see attach)
And the error is:
Get "http://10.0.1.10:9664/metrics": dial tcp 10.0.1.10:9664: connect: connection refused
and
Get "http://192.168.1.30:9100/metrics": context deadline exceeded
![Screenshot_100](https://user-images.githubusercontent.com/23618664/101201080-7953d280-3667-11eb-9329-1a50307bf796.jpg)

MalloZup:
@picoroma  thx for issue! in which cloud are you running this? AWS?

It could be that the ports are not open or some security policy on cloud is causing this. ( on azure for example we do have security rules for this)
let me know tia

picoroma:
Yes, I'm on AWS.
I checked the SG. THe 1st think I verified...
The AllTraffic Rule is in place for host involved (all 10.0.0.0/16 subnet) and also from MY IP
I will perform some more test..
maybe the 192.168.xxx.yyy was missing...

P.

picoroma:
I repeated the test.
The problem seems to be related to traffic from "Monitoring host" to netweaver host.
But, the SG is quite OPEN...
But if I telnet from monitoring host to netweaver host using 192.168.xxx.yyy IP (virtual IP) - telnet fails. See attached images.
If i Use 10.0.2.30 IP (real IP) - Telnet works!

![Screenshot_101](https://user-images.githubusercontent.com/23618664/101261251-11b88880-3736-11eb-8885-ff622f89bb27.jpg)
![Screenshot_102](https://user-images.githubusercontent.com/23618664/101261254-1bda8700-3736-11eb-9aff-c612742df9f7.jpg)
![Screenshot_103](https://user-images.githubusercontent.com/23618664/101261255-2006a480-3736-11eb-9188-78297348becc.jpg)

MalloZup:
thx @picoroma . I will note this , so we can investigate further.

stefanotorresi:
This seems to also affects Azure, pending confirmation.

yeoldegrove:
[...]

PR #710, opened 2021-06-15
configure prometheus targets based on host, HA, VIP addresses

yeoldegrove:
This PR introduces:
 - different grains for
   - `netweaver_targets`       --> node addresses
   - `netweaver_targets_ha` --> HA/corosync addresses
   - `netweaver_targets_vip` --> Virtual Addresses of Applications (might be cluster services)
- This corrects the behaviour of `netweaver_targets` which included only the virtual address before.
- It is now easier to distinguish between the different targets for the different prometheus exporters

This PR fixes:
- azure: netweaver''s application VIPs are only reachable via a load-balancer #611 
  - add port 9680 for `sap_host_expoter` to lb
- aws: netweaver's application VIPs are not reachable #611 
- security group additions for aws/azure/gcp


yeoldegrove:
I did successful test runs on AWS + Azure.
It should work on GCP, libvirt, openstack too as the changes are basically the same everywhere.

salt/monitoring_srv/prometheus/prometheus.yml.j2
```diff
@@ -47,9 +53,11 @@ scrape_configs:
...
         - "{{ ip }}:9100" # node_exporter
-        - "{{ ip }}:9680" # sap_host_exporter
         {%- endfor %}
-        {%- for ip in grains['netweaver_targets'][0:2] %}
+        {%- for ip in grains['netweaver_targets_ha'] %}
         - "{{ ip }}:9664" # ha_cluster_exporter
         {%- endfor %}
+        {%- for ip in grains['netweaver_targets_vip'] %}
+        - "{{ ip }}:9680" # sap_host_exporter
+        {%- endfor %}
```
stefanotorresi, line 62: Hmm, I'm not sure about this: unlike `hanadb_exporter`, `sap_host_exporter` is intended to be targeted on all the nodes, regardless of their role in the HA setup.
With this change, we would effectively stop collecting metrics for the ERS, PAS and AAS, which is not desirable.
We can keep the `netweaver_targets_vip` grain for future use, but I guess we don't actually need to use it, unless I'm missing something.

stefanotorresi (changes requested):
Good changes, but there is something I'm not sure about with the `sap_host_exporter` targets.

salt/monitoring_srv/prometheus/prometheus.yml.j2
yeoldegrove: Here is an example config from a provisioned monitoring node:
```
  - job_name: netweaver
    static_configs:
      - targets:
        - "10.0.3.30:9100" # node_exporter
        - "10.0.4.31:9100" # node_exporter
        - "10.0.3.32:9100" # node_exporter
        - "10.0.4.33:9100" # node_exporter
        - "10.0.3.30:9664" # ha_cluster_exporter
        - "10.0.4.31:9664" # ha_cluster_exporter
        - "192.168.1.30:9680" # sap_host_exporter
        - "192.168.1.31:9680" # sap_host_exporter
        - "192.168.1.32:9680" # sap_host_exporter
        - "192.168.1.33:9680" # sap_host_exporter
```

The grain`netweaver_targets_vip` basically has the old values from `netweaver_targets` ... so nothing really changed in regards to the `sap_host_exporter`.
It has still the 4 VIPs of the actual SAP instances.

aws/main.tf
```diff
@@ -254,22 +254,28 @@ module "hana_node" {
...
+  aws_region            = var.aws_region
+  availability_zones    = data.aws_availability_zones.available.names
+  os_image              = local.monitoring_os_image
+  os_owner              = local.monitoring_os_owner
+  subnet_ids            = aws_subnet.infra-subnet.*.id
+  timezone              = var.timezone
+  hana_targets          = local.hana_ips
+  hana_targets_ha       = var.hana_ha_enabled ? local.hana_ips : []
+  hana_targets_vip      = var.hana_ha_enabled ? [local.hana_cluster_vip] : [local.hana_ips[0]] # we use the vip for HA scenario and 1st hana machine for non HA to target the active hana instance
+  drbd_targets          = var.drbd_enabled ? local.drbd_ips : []
```
arbulu89, line 273: Do we need these `drbd_targets` and `drbd_targets_ha`? Both have the same values

aws/main.tf
```diff
@@ -254,22 +254,28 @@ module "hana_node" {
...
+  monitoring_srv_ip     = local.monitoring_ip
+  aws_region            = var.aws_region
+  availability_zones    = data.aws_availability_zones.available.names
+  os_image              = local.monitoring_os_image
+  os_owner              = local.monitoring_os_owner
+  subnet_ids            = aws_subnet.infra-subnet.*.id
+  timezone              = var.timezone
+  hana_targets          = local.hana_ips
+  hana_targets_ha       = var.hana_ha_enabled ? local.hana_ips : []
+  hana_targets_vip      = var.hana_ha_enabled ? [local.hana_cluster_vip] : [local.hana_ips[0]] # we use the vip for HA scenario and 1st hana machine for non HA to target the active hana instance
```
arbulu89, line 272: Even though this will work, the naming will be a bit confusing, because in non HA scenarios the vip shouldn't be used.
Anyway, I guess we can live with this as the comment clarifies what is going on

salt/monitoring_srv/prometheus/prometheus.yml.j2
arbulu89: Yes, the netweaver VIPs are not only attached to clustered nodes, they are set in all nodes

arbulu89 (changes requested):
Hi @yeoldegrove ,
The functionality looks good.
However, I have a change request. Instead of creating all of these variables in each of the individual provides variables, what about adding these target variables to `common_variables`? Eventually, the `common_variables` are passed to the monitoring module, so we could get rid off many duplicated code in the variables files.

salt/monitoring_srv/prometheus/prometheus.yml.j2
stefanotorresi: uh okay, all good then!

aws/main.tf
yeoldegrove: I could not think of a better naming scheme. So it will stay this way until this changes ;)

aws/main.tf
yeoldegrove: We do not need it at the moment. It is this way to be consistent in all 3 setups we currently deploy.
Maybe a "DRDB exporter" will come in the future... Who knows.

yeoldegrove:
> However, I have a change request. Instead of creating all of these variables in each of the individual provides variables, what about adding these target variables to `common_variables`? Eventually, the `common_variables` are passed to the monitoring module, so we could get rid off many duplicated code in the variables files.

@arbulu89 Good idea. This saves a lot of code. I added the variables to `common_variables` now and did a successfull test buildup on azure.

arbulu89 (approved):
Great @yeoldegrove

stefanotorresi (approved):

PR #714, opened 2021-06-18
add optional native fencing for microsoft-azure

yeoldegrove:
This adds the possibility to deploy HANA, DRBD, Netweaver Clusters with the native "Azure Fence Agent".

- add documentation
- add new variables/grains and pillars to be used by formulas

Code from PR https://github.com/ClusterLabs/fence-agents/pull/415 is needed to make `stonith:fence_azure_arm` work with the latest Microsoft API changes. Without this code, the resources will not start.

The actual resources are defined in formulas and this PRs:
https://github.com/SUSE/saphanabootstrap-formula/pull/130
https://github.com/SUSE/drbd-formula/pull/26
https://github.com/SUSE/sapnwbootstrap-formula/pull/87

### DIFF
--- a/azure/main.tf
+++ b/azure/main.tf
@@ -130,26 +130,30 @@ module "common_variables" {
 }
 
 module "drbd_node" {
-  source              = "./modules/drbd_node"
-  common_variables    = module.common_variables.configuration
-  bastion_host        = module.bastion.public_ip
-  az_region           = var.az_region
-  drbd_count          = var.drbd_enabled == true ? 2 : 0
-  vm_size             = var.drbd_vm_size
-  drbd_image_uri      = var.drbd_image_uri
-  os_image            = local.drbd_os_image
-  resource_group_name = local.resource_group_name
-  network_subnet_id   = local.subnet_id
-  storage_account     = azurerm_storage_account.mytfstorageacc.primary_blob_endpoint
-  cluster_ssh_pub     = var.cluster_ssh_pub
-  cluster_ssh_key     = var.cluster_ssh_key
-  host_ips            = local.drbd_ips
-  fencing_mechanism   = var.drbd_cluster_fencing_mechanism
-  sbd_storage_type    = var.sbd_storage_type
-  iscsi_srv_ip        = join("", module.iscsi_server.iscsisrv_ip)
-  nfs_mounting_point  = var.drbd_nfs_mounting_point
-  nfs_export_name     = var.netweaver_sid
-  drbd_cluster_vip    = local.drbd_cluster_vip
+  source                    = "./modules/drbd_node"
+  common_variables          = module.common_variables.configuration
+  bastion_host              = module.bastion.public_ip
+  az_region                 = var.az_region
+  drbd_count                = var.drbd_enabled == true ? 2 : 0
+  vm_size                   = var.drbd_vm_size
+  drbd_image_uri            = var.drbd_image_uri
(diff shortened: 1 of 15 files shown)


yeoldegrove:
https://github.com/ClusterLabs/fence-agents/pull/415 was created to make `stonith:azure_fence_arm` to work with the latest Microsoft API changes.

azure/main.tf
```diff
@@ -130,26 +130,30 @@ module "common_variables" {
...
+  cluster_ssh_pub           = var.cluster_ssh_pub
+  cluster_ssh_key           = var.cluster_ssh_key
+  host_ips                  = local.drbd_ips
+  fencing_mechanism         = var.drbd_cluster_fencing_mechanism
+  sbd_storage_type          = var.sbd_storage_type
+  iscsi_srv_ip              = join("", module.iscsi_server.iscsisrv_ip)
+  nfs_mounting_point        = var.drbd_nfs_mounting_point
+  nfs_export_name           = var.netweaver_sid
+  drbd_cluster_vip          = local.drbd_cluster_vip
+  subscription_id           = data.azurerm_subscription.current.subscription_id
```
arbulu89, line 153: Could we put a comment on top of these new variables? Saying that they are used only by the azure fence agent

salt/cluster_node/ha/native_fencing.sls
```diff
@@ -0,0 +1,7 @@
+{% if grains['provider'] == 'azure' %}
+python3-azure-identity:
```
arbulu89, line 2: I'm not sure if this pkg installation should go here, or maybe better in the `habootstrap-formula`. We already have some similar package installation for gcp in the formula itself

arbulu89 (commented):
Thanks @yeoldegrove ,
Just a question whether the package installation should go here or in the `habootstrap-formula`

azure/main.tf
yeoldegrove: fixed in 80ed2bb

salt/cluster_node/ha/native_fencing.sls
yeoldegrove: You are absolutely right.
A better place is here: https://github.com/SUSE/habootstrap-formula/pull/83

arbulu89: I guess we will need to remove completely this file, right?

arbulu89 (changes requested):
Hi @yeoldegrove ,
As commented in https://github.com/SUSE/sapnwbootstrap-formula/pull/87, we will need to use the grains values to modify the pillar content of the `pillar_examples/automatic` to use the data from the pillar files, and not the grains, as the grains are only used in this project, but the formulas should work standalone

salt/cluster_node/ha/native_fencing.sls
yeoldegrove: removed in 8c7c7ec

yeoldegrove:
> As commented in [SUSE/sapnwbootstrap-formula#87](https://github.com/SUSE/sapnwbootstrap-formula/pull/87), we will need to use the grains values to modify the pillar content of the `pillar_examples/automatic` to use the data from the pillar files, and not the grains, as the grains are only used in this project, but the formulas should work standalone

@arbulu89 pillars are now used instead of grains

arbulu89 (approved):
Thanks @yeoldegrove

PR #715, opened 2021-06-23
"unicast: True" in libvirt pillars

yeoldegrove:
This sets `unicast: True` in the libvirt pillars.
This is the same value as in any other cloud provider.

### DIFF
--- a/pillar_examples/automatic/hana/cluster.sls
+++ b/pillar_examples/automatic/hana/cluster.sls
@@ -10,8 +10,8 @@ cluster:
   interface: eth1
   {% else %}
   interface: eth0
-  unicast: True
   {% endif %}
+  unicast: True
   join_timeout: 500
   {% if grains['fencing_mechanism'] == 'sbd' %}
   sbd:
--- a/pillar_examples/automatic/netweaver/cluster.sls
+++ b/pillar_examples/automatic/netweaver/cluster.sls
@@ -8,8 +8,8 @@ cluster:
   interface: eth1
   {%- else %}
   interface: eth0
-  unicast: True
   {%- endif %}
+  unicast: True
   {% if grains['fencing_mechanism'] == 'sbd' %}
   sbd:
     device: {{ grains['sbd_disk_device']|default('') }}
--- a/pillar_examples/libvirt/cost_optimized/cluster.sls
+++ b/pillar_examples/libvirt/cost_optimized/cluster.sls
@@ -2,6 +2,7 @@ cluster:
   name: 'hana_cluster'
   init: 'hana01'
   interface: 'eth1'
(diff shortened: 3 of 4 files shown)


arbulu89 (commented):
Hi @yeoldegrove ,
This won't change much, as almost none of the users use these pillar.
We would need to change the pillar in the `pillar_examples/automatic`.
Besides that, we would need to see if this only applies to the HANA cluster, or the DRBD and NETWEAVER clusters should have the same configuration.

@pirat013 as you reported this issues, do you know what we should do?
Maybe @fmherschel or @lpinne could help on this too

yeoldegrove:
> We would need to change the pillar in the `pillar_examples/automatic`.

This is already the case ...
https://github.com/SUSE/ha-sap-terraform-deployments/blob/14b543b6e55bc69e5a012fdad7c3a8bf4675f84c/pillar_examples/automatic/drbd/cluster.sls#L10
https://github.com/SUSE/ha-sap-terraform-deployments/blob/14b543b6e55bc69e5a012fdad7c3a8bf4675f84c/pillar_examples/automatic/hana/cluster.sls#L13
https://github.com/SUSE/ha-sap-terraform-deployments/blob/14b543b6e55bc69e5a012fdad7c3a8bf4675f84c/pillar_examples/automatic/netweaver/cluster.sls#L11

> Besides that, we would need to see if this only applies to the HANA cluster, or the DRBD and NETWEAVER clusters should have the same configuration.

IMHO it is a good idea to alway use unicast. It does not matter what type of cluster it is. There are no drawbacks.
It is e.g. also referenced in the SUSE azure guide: https://documentation.suse.com/sbp/all/html/SLES4SAP-hana-sr-guide-PerfOpt-12_AWS/index.html#id-1.10.6.5.6
And in the SUSE HANA HA guide: https://documentation.suse.com/sbp/all/html/SLES4SAP-hana-sr-guide-PerfOpt-15/index.html#_corosync_configuration

From this point of view there is no point in referencing non unicast setups in our pillars.
This PR is more or less cosmetic and does not change any automation.

arbulu89:
Hi @yeoldegrove ,
I still see that unicast is not set for `hana.sls` and `netweaver.sls` for libvirt, which is the reported error.
```
{% if grains['provider'] == 'libvirt' %}
interface: eth1
{% else %}
interface: eth0
unicast: True
{% endif %}
```

We would need to extract the `unicast` entry from that `if` statement for hana/netweaver

yeoldegrove:
> We would need to extract the `unicast` entry from that `if` statement for hana/netweaver

Thanks for pointing this out.
Should be fixed now.

arbulu89 (approved):
Thanks!

PR #720, opened 2021-06-30
add support for rootdisk as data_disk_type and change default drbd fencing to sbd

melzer-b1:
Deployment failed on using rootdisk as data source.

### DIFF
--- a/salt/bastion/init.sls
+++ b/salt/bastion/init.sls
@@ -1,3 +1,4 @@
+{# 'rootdisk' is used for nfs server on openstack #}
 {% if grains.get('monitoring_enabled', False) or grains.get('data_disk_type') in ['rootdisk', 'ephemeral', 'volume'] %}
 include:
 {% if grains.get('monitoring_enabled', False) %}
--- a/salt/bastion/init.sls
+++ b/salt/bastion/init.sls
@@ -1,9 +1,9 @@
-{% if grains.get('monitoring_enabled', False) or grains.get('bastion_data_disk_type') in ['ephemeral', 'volume'] %}
+{% if grains.get('monitoring_enabled', False) or grains.get('data_disk_type') in ['rootdisk', 'ephemeral', 'volume'] %}
 include:
 {% if grains.get('monitoring_enabled', False) %}
   - bastion.nginx
 {% endif %}
-{% if grains.get('data_disk_type') in ['ephemeral', 'volume'] %}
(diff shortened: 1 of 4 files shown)


yeoldegrove (changes requested):
@melzer-b1 Thanks for adding this!
Could you please add an example line for `rootdisk` in `terraform.tfvars.example`?

melzer-b1:
added rootdisk example in terraform.tfvars.example

arbulu89 (commented):
Hi guys,

I found out now that we create a NFS service in the bastion host for openstack. I don't like much this idea, but well, I guess it is fine.

I would suggest to put some comments in the salt code saying that this only applies to openstack

salt/bastion/init.sls
```diff
@@ -1,9 +1,10 @@
-{% if grains.get('monitoring_enabled', False) or grains.get('bastion_data_disk_type') in ['ephemeral', 'volume'] %}
+{# 'rootdisk' is used for nfs server on openstack #}
```
yeoldegrove, line 1: `rootdisk` AND `ephemeral` AND `volume` are only used on openstack.
Maybe rephrase it like "NFS server for `/sapinst` on bastion is only available on openstack." ?

yeoldegrove (approved):

yeoldegrove:
@arbulu89 good to merge now from your side?

salt/bastion/sapinst.sls
```diff
@@ -10,6 +10,9 @@ nfs-kernel-server:
         attempts: 3
         interval: 15
 
+{# NFS server used for '/sapinst' on bastion is only available on openstack. #}
```
arbulu89, line 13: Do we really need this `jinja` kind of comments?
Using simple `# my comment` should work too

arbulu89:
> @arbulu89 good to merge now from your side?

Yes. I don't personally like this `jinja` kind of comments (mostly, because the rest of the code uses normal comments).
But it is acceptable of course.
Up to you if you want to change them or not

melzer-b1:
My thoughts by using jinja2 comments are: First we can distinguish comments for jinja2 blocks from ordinary salt comments.
Secondary there is a jinja2 feature where you can write a expresion "{% if … }" as "# if …" so i wanted to avoid collisions.
We found some other cases of jinja2 comments in the code base so we decided to use them.

If you think this conflicts to your coding style I can change it to the other type of comment.

PR #721, opened 2021-07-06
Adding more details for terraform.tfvars.example file

pirat013:
To avoid the wrong usage of the variables like lower case SID and passwords with special characters the examples files has been adapted.


pirat013:
If the changes will be accepted and if they are valid for the cloud deployments too, I'm happy to implement the changes there as well.

PR #722, opened 2021-07-08
azure: configure VIPs on NIC based on different scenarios for netweaver

yeoldegrove:
Even with #710 merged there is still a problem on azure to reach the VIPs of PAS/AAS NetWeaver instances from the monitoring host.
The ASCS/ERS VIPs are on a load-balancer and are well reachable. They are also configured inside the OS.
The PAS/AAS VIPs are only configured inside the OS at the moment and are not routed in azure.
This is azure specific and looks completely different on AWS.
For GCP there is a parameter [can_ip_forward](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#can_ip_forward) that solves the issue of "only configured VIP on OS".
For Azure there is a similar parameter [enable_ip_forwarding](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/network_interface#enable_ip_forwarding) which does a different thing in the background and does not work for this use case.

The only way I see is to actually assign these VIPs via terraform.
For this a quite complex logic has to be added, based on the different NetWeaver deployment scenarios.

If merged, this approach could maybe also be ported to AWS/Azure.

### DIFF
--- a/azure/modules/netweaver_node/main.tf
+++ b/azure/modules/netweaver_node/main.tf
@@ -210,10 +210,70 @@ resource "azurerm_network_interface" "netweaver" {
 
   ip_configuration {
     name                          = "ipconf-primary"
+    primary                       = true
     subnet_id                     = var.network_subnet_id
     private_ip_address_allocation = "static"
     private_ip_address            = element(var.host_ips, count.index)
     public_ip_address_id          = local.bastion_enabled ? null : element(azurerm_public_ip.netweaver.*.id, count.index)
+  }
+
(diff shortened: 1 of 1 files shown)


arbulu89 (commented):
@yeoldegrove The usage of virtual ip addreses was a technical decision in the past to configure the monitoring connfiguration, but maybe we could use the actual ip addresses.

Have you thought of this? Seeing that the usage of virtual addresses is inconvenient in the cloud environments...

Maybe, we could even use virtual addresses for clustered nodes and real addresses for the other scenarios

yeoldegrove:
closed in favor of #746

PR #723, opened 2021-07-08
changing SID example, extending SID comment, modifying password comme…

pirat013:
To avoid the wrong usage of the variables like lower case SID and passwords with special characters the examples files has been adapted.
If this change would make sense for the CSP's too, I'm happy to change the example file there as well.


yeoldegrove (approved):
I added the remaining cloud providers and rebased it.

Issue #724, opened 2021-07-08
Deployment on azure fails possibly due to missing `netweaver_ips` var


rtorrero:
**Used cloud platform**
Azure

**Used SLES4SAP version**
SLES15SP2

**Used client machine OS**
Linux

**Expected behaviour vs observed behaviour**
Attempting to deploy to azure results in a immediate failure

```
│   on infrastructure.tf line 3, in provider "azurerm":
│    3:   version = "~> 2.32.0"
│ 
│ Terraform 0.13 and earlier allowed provider version constraints inside the provider configuration block, but that is now
│ deprecated and will be removed in a future version of Terraform. To silence this warning, move the provider version
│ constraint into the required_providers block.
╵
╷
│ Error: Invalid index
│ 
│   on main.tf line 137, in module "common_variables":
│  137:   monitoring_netweaver_targets_ha     = var.netweaver_ha_enabled ? [local.netweaver_ips[0], local.netweaver_ips[1]] : []
│     ├────────────────
│     │ local.netweaver_ips is empty list of string
│ 
│ The given key does not identify an element in this collection value.
╵
╷
│ Error: Invalid index
│ 
│   on main.tf line 137, in module "common_variables":
│  137:   monitoring_netweaver_targets_ha     = var.netweaver_ha_enabled ? [local.netweaver_ips[0], local.netweaver_ips[1]] : []
│     ├────────────────
│     │ local.netweaver_ips is empty list of string
│ 
│ The given key does not identify an element in this collection value.
```
**How to reproduce**
Attempt to deploy to azure without specifying the optional variable `netweaver_ips`

**Used terraform.tfvars**
```
pre_deployment         = true
provisioning_log_level = "info"

# Region where to deploy the configuration
az_region = "westeurope"
admin_user = "cloudadmin"


# Optional SUSE Customer Center Registration parameters
reg_code = "***"
reg_email = "***"

# Public sles4sap image
os_image = "SUSE:sles-sap-15-sp2-byos:gen2:latest"

public_key      = "~/.ssh/id_rsa.pub"
private_key     = "~/.ssh/id_rsa"

# Azure storage account name
storage_account_name = "***"

# Azure storage account secret key (key1 or key2)
storage_account_key = "***"

# Number of nodes in the cluster
hana_master_password               = "****"
hana_count                         = 2
hana_ha_enabled                    = true
hana_vm_size                       = "Standard_E4s_v3"
hana_enable_accelerated_networking = false
[...]

PR #726, opened 2021-07-09
fix monitoring and azure native fence default values

yeoldegrove:
This fixes:
 - a logic error in defining `monitoring_netweaver_targets_ha` #724.
 - set default values for azure native fencing variables

### DIFF
--- a/aws/main.tf
+++ b/aws/main.tf
@@ -136,7 +136,7 @@ module "common_variables" {
   monitoring_drbd_targets_ha          = var.drbd_enabled ? local.drbd_ips : []
   monitoring_drbd_targets_vip         = var.drbd_enabled ? [local.drbd_cluster_vip] : []
   monitoring_netweaver_targets        = var.netweaver_enabled ? local.netweaver_ips : []
-  monitoring_netweaver_targets_ha     = var.netweaver_ha_enabled ? [local.netweaver_ips[0], local.netweaver_ips[1]] : []
+  monitoring_netweaver_targets_ha     = var.netweaver_enabled ? (var.netweaver_ha_enabled ? [local.netweaver_ips[0], local.netweaver_ips[1]] : []) : []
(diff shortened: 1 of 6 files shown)


aws/main.tf
```diff
@@ -136,7 +136,7 @@ module "common_variables" {
   monitoring_drbd_targets_ha          = var.drbd_enabled ? local.drbd_ips : []
   monitoring_drbd_targets_vip         = var.drbd_enabled ? [local.drbd_cluster_vip] : []
   monitoring_netweaver_targets        = var.netweaver_enabled ? local.netweaver_ips : []
-  monitoring_netweaver_targets_ha     = var.netweaver_ha_enabled ? [local.netweaver_ips[0], local.netweaver_ips[1]] : []
+  monitoring_netweaver_targets_ha     = var.netweaver_enabled ? (var.netweaver_ha_enabled ? [local.netweaver_ips[0], local.netweaver_ips[1]] : []) : []
```
arbulu89, line 139: I would say the next looks more logic:

```
var.netweaver_enabled && var.netweaver_ha_enabled ? [local.netweaver_ips[0], local.netweaver_ips[1]]: []
```

arbulu89 (approved):
@yeoldegrove It looks good.
I proposed an alternative solution though ;)

aws/main.tf
yeoldegrove: @arbulu89 your approach is now in 7dbfec8
Good to merge?

arbulu89: Yes. Test is just in case, maybe I wrote some syntax error hehe

rtorrero:
I can confirm that the issue #724 is solved by this. Thanks!

aws/main.tf
yeoldegrove: As a matter of fact, you had one ;)

PR #731, opened 2021-07-22
Melzer/iscsi

melzer-b1:
Check if iscsi-formula should be installed on node. It is only needed on nodes which use sbd and on the iscsi server.

### DIFF
--- a/salt/os_setup/packages.sls
+++ b/salt/os_setup/packages.sls
@@ -1,5 +1,5 @@
 # check if iscsi-formula should be installed
-{%- if grains.get('sbd_storage_type') == "iscsi" or grains.get('role') == "iscsi_srv" %}
+{%- if grains.get('role') == "iscsi_srv" %}
 iscsi-formula:
   pkg.installed:
     - retry:
--- a/salt/os_setup/packages.sls
+++ b/salt/os_setup/packages.sls
@@ -1,3 +1,5 @@
+# check if iscsi-formula should be installed
+{%- if grains.get('sbd_storage_type') == "iscsi" or grains.get('role') == "iscsi_srv" %}
 iscsi-formula:
   pkg.installed:
     - retry:
@@ -19,3 +21,4 @@ kernel-default:
   # install kernel-default if kernel-default-base is installed, do not touch otherwise
   - require:
     - pkg: kernel-default-base
+{%- endif %}
--- a/salt/cluster_node/ha/iscsi_initiator.sls
+++ b/salt/cluster_node/ha/iscsi_initiator.sls
@@ -10,6 +10,12 @@ lsscsi:
       attempts: 3
       interval: 15
 
+iscsi-formula:
+  pkg.installed:
+  - retry:
+      attempts: 3
+      interval: 15
+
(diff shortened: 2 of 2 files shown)


yeoldegrove (changes requested):
I find it quite tricky to remove iscsi from `os_setup` as e.g. `kernel-default` can only be installed here (`cluster_node` state will not trigger a reboot).
Nevertheless, does it make sense to remove `iscsi-formula` completely from `os_setup`?
@arbulu89 What do you think?

arbulu89:
Hey @melzer-b1 ,
I don't think we use the `iscsi-formula` in the iscsi initiator in the nodes.
Have you tried the current changes without that part?

Anyway, as @yeoldegrove commented, maybe it makes more sense to remove this from `os_setup` and put it in the `iscsi_srv` folder, as the role based filtering would be already done there

melzer-b1:
I've tested the changes without this part and removed it.

yeoldegrove:
@melzer-b1 As @arbulu89 pointed out... Is `iscsi-formula` really needed on the the iscsi initiators (clients)?
Please double check.

salt/os_setup/packages.sls
```diff
@@ -1,3 +1,5 @@
+# check if iscsi-formula should be installed
```
arbulu89, line 1: As commented previously, to make the code better and use the roles properly, we should just move the `iscsi-formula` installation from `os_setup/packages` to `iscsi_srv` folder (we could create there a `packages.sls` file if needed). This way, the role filtering would be already done and we wouldn't need any `if` statement, as `iscsi_srv` is only executed in iscsi servers which, and the user already has selected to use it

melzer-b1:
@arbulu89 
I removed the first part of the condition. So iscsi-formula is only  installed on node with role iscsi_srv. It works.

arbulu89:
> @arbulu89
> I removed the first part of the condition. So iscsi-formula is only installed on node with role iscsi_srv. It works.

Hey @melzer-b1 ,
I'm sure it works, but I still think the solution proposed in the next comment is the way to go:https://github.com/SUSE/ha-sap-terraform-deployments/pull/731#discussion_r677156854
Unless it has some major contraindication, that I cannot think of

melzer-b1:
@arbulu89 the iscsi-formula is now installed from within iscsi_srv. It is removed from os_setup.
We had to solve a chicken egg problem here. Because the installed package iscsi-formula requires further salt run we
introduced role:iscsi_srv in deployment to be run after predeployment. We also had to modify the provision.sh script.

arbulu89:
> @arbulu89 the iscsi-formula is now installed from within iscsi_srv. It is removed from os_setup.
> We had to solve a chicken egg problem here. Because the installed package iscsi-formula requires further salt run we
> introduced role:iscsi_srv in deployment to be run after predeployment. We also had to modify the provision.sh script.

Ok...
Now I understand why we had the `iscsi-formula` installation there. I remember that we had this exact problem.
Knowing this, I would prefer to go back to the https://github.com/SUSE/ha-sap-terraform-deployments/pull/731/commits/fb910b628640adfeba865a8af7378978e1332d3b
The code was simpler.
Sorry, for all the inconveniences...
Once you have it, we will merge it, as this commit was tested. Just remove the last code.
Thanks @melzer-b1

melzer-b1:
Rolled git branch back to first solution.

arbulu89 (approved):
Thanks @melzer-b1

PR #732, opened 2021-07-28
add first steps to enable HANA scale-out scenario

yeoldegrove:
* shared storage via DRBD/NFS
  * HANA scale-out provisioning via "extra_parameters.addhosts"

### DIFF
--- a/azure/README.md
+++ b/azure/README.md
@@ -139,6 +139,7 @@ For example, the disks configuration for the HANA database is a crucial step in 
 #### Demo
 
 ```
+hana_count = "2"
 hana_vm_size = "Standard_E4s_v3"
 hana_enable_accelerated_networking = false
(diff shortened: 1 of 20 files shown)


arbulu89 (commented):
Hi @yeoldegrove ,
I have just done a superficial check, and I have to confess that it looks pretty heavy.

I think that we need to split things up.
I would have different implementations for scale out and drbd. So I would appreciate if the PR is splitted. I know that one depends on the other, so let's make them in the correct order
In fact, I don't know if I would even start with drbd, as it is not the most preferred option by SUSE. Maybe @diegoakechi or @stefanotorresi can put some light on this.

yeoldegrove:
After a few discussions I won't implement drbd for HANA scale-out as an option.
It might come later, or not.
I will open a new PR with upcoming features.

PR #733, opened 2021-07-30
azure: add initial ANF (Azure Netapp Files) shared storage support

yeoldegrove:
This is a preview of initial ANF (Azure Netapp Files) shared storage support for azure.
ANF implementation is based on https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-scale-out-standby-netapp-files-suse#mount-the-azure-netapp-files-volumes

The variables *_anf_quota_* could be put into a Dictionary to have fewer variables.

HANA deployment is tested with this code. Netweaver is outstanding.


nick-wang:
Are the commits of this pull request also included in https://github.com/SUSE/ha-sap-terraform-deployments/pull/734?
If so, could focus on the later one only.

yeoldegrove:
> Are the commits of this pull request also included in #734?
> If so, could focus on the later one only.

Yes they are...as this shared storage implementation is a prerequisite to enable scale-out in #734.

nick-wang:
> Yes they are...as this shared storage implementation is a prerequisite to enable scale-out in #734.
> The ANF feature should be merged separately though.

Well, it is slightly wired we have everything(discussion/code) in #734 but merge this pull request first.
Even if we do so, the scale-out pull request (#734) need rebase on this. Then the code belong here won't be shown in #734 anymore.

azure/infrastructure.tf
```diff
@@ -105,6 +136,29 @@ resource "azurerm_route_table" "myroutes" {
   }
 }
 
+# Azure Netapp Files resources
+resource "azurerm_netapp_account" "mynetapp-acc" {
+  count = local.shared_storage_anf
+  # name                = local.anf_account_name
```
arbulu89, line 142: Are these comments some leftovers?

azure/infrastructure.tf
```diff
@@ -105,6 +136,29 @@ resource "azurerm_route_table" "myroutes" {
   }
 }
 
+# Azure Netapp Files resources
```
arbulu89, line 139: What are these resources for? If we had some document links it would be helpful

azure/modules/hana_node/main.tf
```diff
@@ -201,6 +202,148 @@ resource "azurerm_image" "sles4sap" {
   }
 }
 
+# ANF volumes
+resource "azurerm_netapp_volume" "hana-netapp-volume-data" {
```
arbulu89, line 206: Would it make sense to move this to a new terraform module? It would't reduce much the amount of code, but it might help a bit reducing the number of constants that are duplicated all over the code

arbulu89 (commented):
Hi @yeoldegrove ,
In general the code looks great. I have just added some few comments.
Let me know when you remove the draft tag, I don't think we are far away of having a definitive code

azure/infrastructure.tf
yeoldegrove: Restructured the resources and added some more input to the README.
A lot more information on how to use ANF will follow in #734.

azure/infrastructure.tf
yeoldegrove: removed

azure/terraform.tfvars.example
```diff
@@ -148,6 +153,29 @@ cluster_ssh_key = "salt://sshkeys/cluster.id_rsa"
 # BYOS example
 # bastion_os_image = "SUSE:sles-sap-15-sp2-byos:gen2:latest"
 
+#########################
+# shared storage variables
```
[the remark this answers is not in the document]
yeoldegrove: fixed

azure/main.tf
```diff
@@ -226,10 +234,17 @@ module "hana_node" {
...
   # only used by azure fence agent (native fencing)
-  subscription_id           = data.azurerm_subscription.current.subscription_id
-  tenant_id                 = data.azurerm_subscription.current.tenant_id
-  fence_agent_app_id        = var.fence_agent_app_id
-  fence_agent_client_secret = var.fence_agent_client_secret
+  subscription_id                 = data.azurerm_subscription.current.subscription_id
+  tenant_id                       = data.azurerm_subscription.current.tenant_id
+  fence_agent_app_id              = var.fence_agent_app_id
+  fence_agent_client_secret       = var.fence_agent_client_secret
+  hana_scale_out_anf_quota_data   = var.hana_scale_out_anf_quota_data
```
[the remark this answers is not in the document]
yeoldegrove: fixed

azure/modules/hana_node/main.tf
```diff
@@ -201,6 +202,148 @@ resource "azurerm_image" "sles4sap" {
...
+  count               = local.shared_storage_anf * 2
```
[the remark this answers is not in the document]
yeoldegrove: This is the number of sites, which is hard-coded to 2.
You are right, a non-HA HANA scale-out would only need 1 site.
I adapted it now to use the new local variable `site` which is multiplied by `shared_storage_anf` (0 or 1) now.

azure/modules/hana_node/main.tf
yeoldegrove: I would vote against putting this into a new module. There are not that many resources added (only 4 for HANA and 1 for Netweaver) and initializing 2 new modules would mean defining many of the variables from `{hana,netweaver}_node/variables.tf` again.
Initializing only 1 new module for both HANA/Netweaver would mean abstracting this even more and adding additional variables for e.g. the name of the volume (counts in HANA case, does not count in Netweaver case).

salt/hana_node/mount/azure.sls
```diff
@@ -47,5 +47,94 @@ mount_{{ vg }}_{{ loop.index0 }}:
...
+## scale-out on ANF NFS storage
+
+install_nfs_client:
+  pkg.installed:
+    - name: nfs-client
+    - retry:
+       attempts: 3
+       interval: 15
+
+/etc/idmapd.conf:
```
arbulu89, line 63: The next 3 states looks the same for netweaver.
Could we move them to a state so we can reuse them without duplicating the code?

salt/hana_node/mount/azure.sls
```diff
@@ -47,5 +47,94 @@ mount_{{ vg }}_{{ loop.index0 }}:
...
+      - pkg: nfs-client
+      - file: /etc/idmapd.conf
+
+clear_idmap_cache:
+  cmd.run:
+    - name: nfsidmap -c
+    - onchanges:
+      - file: /etc/idmapd.conf
+
+# We cannot use showmount as some of the required ports are not always available
```
arbulu89, line 88: I would bet this piece of code exists somewhere else.
Could we create a new state/macro to have it only in 1 place?
Even though it is not exactly the same, using macros we can pass variables to populate some of the parameters.
Just give a thought

arbulu89 (commented):
Hi @yeoldegrove ,
The code looks nice! I just added a couple of comments to see if we can remove some duplicated (or almost duplicated) code

salt/hana_node/mount/azure.sls
yeoldegrove: @arbulu89 added dedicated states in 9db80e1. Please have a look.

yeoldegrove:
@arbulu89 please review again

arbulu89 (approved):
Hey @yeoldegrove ,
It looks good (I didn't have a deep look. I guess that you at least gave a try in other cloud providers to see that it doesn't break anything, as it looks a quite invasive change)

stefanotorresi:
@yeoldegrove I didn't quite understand one thing: you say this implementation depends on the scale-out, and there are references to that in the code, but that's not merged in yet. What gives?

yeoldegrove:
@stefanotorresi It is the other way around... scale-out depends on this ANF code, as shared storage is needed for scale-out.
Nevertheless this ANF implementation is based on the "Azure HANA scale-out reference implementation" (this is what the link is supposed th say).

The plan is to merge this PR first and after that the scale-out PR #734.
I am just testing out a few last scenarios.

stefanotorresi:
Thank you for the clarification!

PR #734, opened 2021-07-30
azure: HANA scale-out

yeoldegrove:
prerequisites:
- #733 shared storage via ANF is needed for this.
- an updated `saphanabootstrap-formula` is needed https://github.com/SUSE/saphanabootstrap-formula/pull/133

This is a work in progress version of a HANA scale-out implementation.
It is still missing features and is not fully tested.

features implemented:

- corosync
  - pcmk_action_limit=-1
  - wait_for_all=1

- pacemaker
  - migration-threshold=50

- stonith settings
  - concurrent-fencing=true

- HANA scale-out resources
  - add resources
  - SAPHanaSR hook
    - sudoers

- majority maker
  - initial steps done
  - actual implementation missing

- shared storage for /hana/{data/log/shared}
  - via Azure Netapp Files (ANF)

TODOs (incomplete):
- The variables `hana_scale_out_addhosts` and `hana_scale_out_site_*` can be consolidated.
- Look into putting majority maker (mm) into own module and make it work at all.


azure/README.md
```diff
@@ -136,9 +136,72 @@ The database configuration may vary depending on the expected performance. In or
...
+Please read [Deploy a SAP HANA scale-out system with standby node on Azure VMs by using Azure NetApp Files on SUSE Linux Enterprise Server](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/sap/sap-hana-scale-out-standby-netapp-files-suse) to get an idea how this setup will look like.
+
+To enable a scale-out deployment on 4,6,8,... HANA nodes you have to raise the `hana_count` variable accordingly.
+
+HANA sites and roles must be provided via these parameters in `terraform.tfvars`. Please make sure your hostnames match with your defined values.
+
+```
+hana_count = 6
+hana_scale_out_enabled = true
+hana_scale_out_site_01 = ["vmhana01", "vmhana03", "vmhana05"]
```
arbulu89, line 150: This will be difficult to identify. Maybe we could start just assuming that the odd host number are in site 1 and even numbers hosts in site 2

azure/README.md
```diff
@@ -136,9 +136,72 @@ The database configuration may vary depending on the expected performance. In or
...
+hana_scale_out_site_02 = ["vmhana02", "vmhana04", "vmhana06"]
+hana_scale_out_addhosts = {
```
arbulu89, line 152: Based in the last comment, I don't know if we can find a way to make this map easier. I'm sure that the users will find it difficult.
Maybe we just start asking simple question like. `worker_count = 2`. This would mean that the 1st 2 nodes act as worker and the remaining ones as standby.

I want to make it simple

pillar/top.sls
```diff
@@ -11,6 +11,11 @@ base:
     - match: compound
     - hana.cluster
 
+  'role:mm_node':
+    - match: grain
+    - hana.hana
```
arbulu89, line 16: Why would we want to run `hana` in the majority maker?

pillar_examples/automatic/hana/cluster.sls
```diff
@@ -37,31 +48,47 @@ cluster:
     password: linux
   {% endif %}
   resource_agents:
+  {% if grains['hana_scale_out_enabled']|default(False) %}
```
arbulu89, line 51: This looks good. I have doubts about if we need to do it here and the `saphanabootstrap-formula` together.
Initially we thought to do it here, because some nodes (like the majority maker) won't have tha hana formula execution (as example)

arbulu89 (commented):
Hi @yeoldegrove ,
I see many things I like (the majority maker implementation for example), and other that I don't like that much (how we ask the user to define the scale out machines setup).

I understand why you did it this way, because it offers a high amount of flexibility, but it worries me that it could be a little bit challenging to configure it right.
We already had bad experiences asking the users to put variables in the terraform.tfvars file, and if we start adding dictionaries and things like that, we will lose them

azure/README.md
yeoldegrove: I also already thought about using the even/odd approach.
You would have to configure less but would be less flexible.
Maybe a hybrid approach is also possible... I will look into it.

azure/README.md
yeoldegrove: > Maybe we just start asking simple question like. worker_count = 2. This would mean that the 1st 2 nodes act as worker and the remaining ones as standby.

Good idea. Will look into it.

pillar/top.sls
yeoldegrove: work in progress ;)

pillar_examples/automatic/hana/cluster.sls
yeoldegrove: I actually had issues in my test setup because this is defined in 2 places.
I will remove it here than and move (un)installation completely to `saphanabootstrap-formula`.

yeoldegrove:
> I understand why you did it this way, because it offers a high amount of flexibility, but it worries me that it could be a little bit challenging to configure it right.
> We already had bad experiences asking the users to put variables in the terraform.tfvars file, and if we start adding dictionaries and things like that, we will lose them

I totally get your point here... This should be as convenient as possible.
It will be improved to have less parameters and/or sane defaults so that you "have to think about less" when running it.

azure/README.md
```diff
@@ -266,6 +373,13 @@ netweaver_data_disk_caching = "ReadWrite"
 netweaver_xscs_accelerated_networking = false
 netweaver_app_accelerated_networking = true
 netweaver_app_server_count = 5
+
+netweaver_shared_storage_type = "drbd" # only drbd supported at the moment
+or
+hana_scale_out_shared_storage_type = "anf"
+anf_pool_size = 4
+anf_pool_service_level = "Premium"
+netweaver_anf_quota_data = "400"
```
nick-wang, line 382: This may not perfect to be comment in this pull request. But after more entries added into `small, medium, large` scenario. I see more and more duplicated entries list. Maybe only list the different items and show `*and*` ones in an additional section is much easier.

> For reference:
Exactly the same content between `Small` and `Medium` (besides extra quotation mark in small/Demo).
Only `netweaver_app_server_count = 10` changed in `Large`

salt/mm_node/packages.sls
```diff
@@ -0,0 +1,12 @@
...
+  pkg.removed:
+    - pkgs:
+      - SAPHanaSR
+      - SAPHanaSR-doc
+
+install_SAPHanaSR_ScaleOut:
+  pkg.installed:
+    - pkgs:
+      - SAPHanaSR-ScaleOut
+      - SAPHanaSR-ScaleOut-doc
```
nick-wang, line 12: Is this necessary? since it is added in https://github.com/SUSE/saphanabootstrap-formula/pull/133 as well.

salt/hana_node/mount/azure.sls
```diff
@@ -47,5 +47,83 @@ mount_{{ vg }}_{{ loop.index0 }}:
     - require:
       - hana_format_lv_{{ vg }}_{{ loop.index0 }}_azure
 
-{% endfor %}
 {%- endfor %}
+{%- endfor %}
+
+{%- if grains['hana_scale_out_enabled'] and grains['hana_scale_out_shared_storage_type'] == "anf" %}
+## scale-out on DRBD/NFS storage
```
nick-wang, line 54: should be `anf/NFS` storage now.

pillar_examples/automatic/hana/cluster.sls
```diff
@@ -36,35 +47,46 @@ cluster:
...
       token: 20000
       consensus: 24000
   {% endif %}
+  {% if grains['hana_scale_out_enabled']|default(False) %}
+    quorum:
+      # expected_votes: {{ grains['node_count'] }}
+      wait_for_all: 1
+  {% else %}
+    # quorum:
+      # expected_votes: {{ grains['node_count'] }}
```
nick-wang, line 70: It looks wired only comments left in `else`, can remove the `else` condition.

azure/README.md
yeoldegrove: I boiled this down so that these are the only HANA related variables that need to be set to enable the scale-out scenario.
```
hana_count = "8"
hana_scale_out_enabled = true
```
In additon to that one might want to tweak the storage variables... but these have some sane defaults too.

azure/README.md
yeoldegrove: This is now implemented via:

```
hana_scale_out_standby_count = 1 # default:1 - Deploy X standby nodes per site. The rest of the nodes will be worker nodes.
```

yeoldegrove: No need to define the sites anymore... This is done by odd/even hostnames now.

pillar/top.sls
yeoldegrove: removed

pillar_examples/automatic/hana/cluster.sls
yeoldegrove: this is handled in https://github.com/SUSE/saphanabootstrap-formula/pull/133

azure/README.md
yeoldegrove: I now added separate sections in `azure/README.md` for scale-out and ANF.

azure/README.md
```diff
@@ -279,6 +393,13 @@ netweaver_data_disk_caching = "ReadWrite"
 netweaver_xscs_accelerated_networking = false
 netweaver_app_accelerated_networking = true
 netweaver_app_server_count = 10
+
+netweaver_shared_storage_type = "drbd" # only drbd supported at the moment
```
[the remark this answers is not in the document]
yeoldegrove: should be fixed

salt/mm_node/packages.sls
yeoldegrove: `saphanabootstrap-formula` is not run on the mm node and the cluster pillar `pillar_examples/automatic/hana/cluster.sls` does not include the resource-agents anymore (in favor of `saphanabootstrap-formula`). This is why we need it.

salt/hana_node/mount/azure.sls
yeoldegrove: fixed

pillar_examples/automatic/hana/cluster.sls
yeoldegrove: Removed the commented out else section.

aws/main.tf
```diff
@@ -229,6 +234,8 @@ module "netweaver_node" {
     aws_security_group_rule.ssh,
     aws_security_group_rule.outall
   ]
+  # prevent timeouts and race conditions on waiting for DB
+  depends_on = [module.hana_node]
```
arbulu89, line 238: I'm not really sure about this. 
The netweaver module was initially designed to be independent of the hana execution, that's why it has its own synchronization parts in salt.

azure/README.md
```diff
@@ -360,6 +368,73 @@ netweaver_app_accelerated_networking = true
...
+  luns        = "0"
+  names       = "usrsap"
+  lv_sizes    = "100"
+  mount_paths = "/usr/sap"
+}
+```
+
+#### terraform parallelism
+
+When deploying many scale-out nodes, e.g. 8 or 10, you should must pass the [`-nparallelism=n`](https://www.terraform.io/docs/cli/commands/apply.html#parallelism-n) parameter to `terraform apply` operations.
```
arbulu89, line 431: Is there anyway to update the default value?
As a fallback is good option, but I'm sure the people won't read this...

azure/main.tf
```diff
@@ -8,31 +8,33 @@ module "local_execution" {
 # Iscsi server: 10.74.0.4
 # Monitoring: 10.74.0.5
 # Hana ips: 10.74.0.10, 10.74.0.11
+# MM ip: 10.74.0.9
```
arbulu89, line 11: Let's be more precies with `Majority maker`

azure/main.tf
```diff
@@ -8,31 +8,33 @@ module "local_execution" {
...
+# Netweaver ips: 10.74.0.60, 10.74.0.61, 10.74.0.62, 10.74.0.63
+# Netweaver virtual ips: 10.74.0.64, 10.74.0.65, 10.74.0.66, 10.74.0.67
 # If the addresses are provided by the user will always have preference
 locals {
   iscsi_ip      = var.iscsi_srv_ip != "" ? var.iscsi_srv_ip : cidrhost(local.subnet_address_range, 4)
   monitoring_ip = var.monitoring_srv_ip != "" ? var.monitoring_srv_ip : cidrhost(local.subnet_address_range, 5)
 
   hana_ip_start              = 10
   hana_ips                   = length(var.hana_ips) != 0 ? var.hana_ips : [for ip_index in range(local.hana_ip_start, var.hana_count + local.hana_ip_start) : cidrhost(local.subnet_address_range, ip_index)]
+  mm_ip                      = var.mm_ip != "" ? var.mm_ip : cidrhost(local.subnet_address_range, local.hana_ip_start - 1)
```
arbulu89, line 25: I would rename to `majority_maker_ip`

azure/modules/hana_node/main.tf
```diff
@@ -256,7 +434,56 @@ resource "azurerm_virtual_machine" "hana" {
...
+    enabled     = "true"
+    storage_uri = var.storage_account
+  }
+
+  tags = {
+    workspace = var.common_variables["deployment_name"]
+  }
+}
+
+resource "azurerm_virtual_machine" "mm" {
```
arbulu89, line 460: I would prefer to have the majority maker in a new module honestly.
I think it might be other scenarios where we might be using it in the future (there are some new Netweaver scenarios in draft mode with this mechanism for example)

arbulu89 (commented):
Hi @yeoldegrove ,
I tried to review this PR, but it is difficult right now.
Let's try to merge first https://github.com/SUSE/ha-sap-terraform-deployments/pull/733 this one, to have a cleaner changes set

aws/main.tf
yeoldegrove: It was just an idea to really have a reliable an direct dependency at terraform level.
I might have to tune the "synchronization parts in salt" than. Let me retest this without the `depends_on`.

azure/README.md
yeoldegrove: As far as my research goes it would be possible to put it into `$HOME/.terraformrc` or do `export TF_CLI_ARGS_apply="-parallelism=1"` ...
Configuring it on a per project level seems not to be possible.

azure/main.tf
yeoldegrove: Renamed everything `mm` to `majority_maker`. One exception is the hostname where `mm` is just shorter and sufficient.

azure/modules/hana_node/main.tf
yeoldegrove: @arbulu89 5e2d77a adds the majority maker as an own module.
The module is than called by the hana module, so there is the possibility to inject variables from hana to majority maker.
The new module could later also be called from e.g. netweaver nodes.

PR #736, opened 2021-08-05
Switch to using IBM Cloud Object Storage

mfriesenegger:
Installing SAP HANA from an NFS server was used for testing prior to this pull request.  This pull request uses the more cost effective method which downloads SAP media from IBM Cloud Object Storage.

### DIFF
--- a/salt/hana_node/download_hana_inst.sls
+++ b/salt/hana_node/download_hana_inst.sls
@@ -44,6 +44,36 @@ hana_inst_directory:
   grains['hana_inst_master'],
   grains['hana_inst_folder']) }}
 
+{% elif grains['provider'] == 'powervs' %}
+
+{% set headers = {"Accept":"application/json","Content-Type":"application/x-www-form-urlencoded"} %}
+{% set data = {"response_type":"cloud_iam","grant_type":"urn:ibm:params:oauth:grant-type:apikey", "apikey": grains["ibmcloud_api_key"]} %}
+{% set cos_access_response = salt['http.query'](url="https://iam.cloud.ibm.com/oidc/token", method="POST", header_dict=headers, data=data).body | load_json %}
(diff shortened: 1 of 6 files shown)


salt/hana_node/download_hana_inst.sls
```diff
@@ -44,6 +44,37 @@ hana_inst_directory:
   grains['hana_inst_master'],
   grains['hana_inst_folder']) }}
 
+{% elif grains['provider'] == 'powervs' %}
```
arbulu89, line 47: I guess that we will want to download other software components (like NW) in the future, so it would be nice to have this in a macro, as we do for `gcp`.
Anyway, let's keep it this way by now. When we have this scenario we will make the change

salt/hana_node/download_hana_inst.sls
```diff
@@ -44,6 +44,37 @@ hana_inst_directory:
...
+
+{% set headers = {"Accept":"application/json","Content-Type":"application/x-www-form-urlencoded"} %}
+{% set data = {"response_type":"cloud_iam","grant_type":"urn:ibm:params:oauth:grant-type:apikey", "apikey": grains["ibmcloud_api_key"]} %}
+{% set cos_access_response = salt['http.query'](url="https://iam.cloud.ibm.com/oidc/token", method="POST", header_dict=headers, data=data).body | load_json %}
+{# json_query requires python3-jmespath package to be installed #}
```
arbulu89, line 52: We could maybe remove this comment, right? It doesn't add much value

arbulu89 (approved):
@mfriesenegger ,
Great job. I did some small comments.
Ready to be merged in any case!

mfriesenegger:
@arbulu89 I forced pushed the update to powervs/modules/hana_node/salt_provisioner.tf which was mistakenly part of #737.  Please review and merge.

PR #737, opened 2021-08-05
Fix download speed when using bastion

mfriesenegger:
Identified download speed issues while testing #736 with bastion deployments.  This PR adapts the bastion setup based on IBM recommendations in https://test.cloud.ibm.com/docs/power-iaas?topic=power-iaas-using-rhel-within-the-power-systems-virtual-server-service


powervs/modules/bastion/main.tf
```diff
@@ -14,6 +14,7 @@ locals {
 #   The need for an extra volume can be removed once https://github.com/IBM-Cloud/terraform-provider-ibm/pull/2797
 #   is released.
 resource "ibm_pi_volume" "bastion_volume"{
+  count                 = local.bastion_count
```
arbulu89, line 17: It looks like we have an extra space (the indentation(

powervs/modules/hana_node/outputs.tf
```diff
@@ -24,7 +24,7 @@ depends_on            = [ibm_pi_instance.ibm_pi_hana]
 }
 
 data "ibm_pi_instance_ip" "ibm_pi_hana_public" {
-count                 = var.hana_count
+count                 = local.bastion_enabled == "false" ? var.hana_count : 0
```
arbulu89, line 27: Does this work? I thought that `bastion_enabled` if a boolean, and here you are comparing with a string.
It looks dangerous at least

powervs/modules/hana_node/salt_provisioner.tf
```diff
@@ -1,3 +1,45 @@
...
+  bastion_user        = var.common_variables["authorized_user"]
+  bastion_private_key = var.common_variables["bastion_private_key"]
+  }
+
+  # Sets up instance to use bastion SNAT router - https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-using-linux#linux-networking
+  # and https://test.cloud.ibm.com/docs/power-iaas?topic=power-iaas-using-rhel-within-the-power-systems-virtual-server-service
+  # This assume eth0 is on the private subnet
+  provisioner "file" {
+    content     = <<EOF
+#!/bin/bash
```
arbulu89, line 24: The same question than before. Could this go in `salt`?

powervs/modules/hana_node/salt_provisioner.tf
```diff
@@ -30,18 +72,10 @@ hana_data_disks_wwn: {${join(", ", [for i in range(0, var.hana_count): format("'
 sbd_disk_device: "${var.common_variables["hana"]["sbd_storage_type"] == "shared-disk" ? format("/dev/disk/by-id/wwn-0x%s", lower(var.sbd_disk_wwn)) : ""}"
 cluster_ssh_pub:  ${var.cluster_ssh_pub}
 cluster_ssh_key: ${var.cluster_ssh_key}
+ibmcloud_api_key: ${var.ibmcloud_api_key}
```
arbulu89, line 75: Is this something forgotten from other changes?
It doesn't look like it has any relation with the rest of the PR

arbulu89 (commented):
Hi @mfriesenegger ,
Added some doubts I have about the the usage of this shell scripts out of salt

powervs/modules/hana_node/outputs.tf
mfriesenegger: Yes this worked but I see your point.  I will change to local.bastion_enabled == false and retest.

powervs/modules/bastion/main.tf
mfriesenegger: FIxed.

powervs/modules/hana_node/salt_provisioner.tf
mfriesenegger: I was having difficulty with quote handling using salt inline.  The second note in https://www.terraform.io/docs/language/resources/provisioners/remote-exec.html states "Since inline is implemented by concatenating commands into a script" which gave me the idea to do this.  This is the easiest solution to the quote handling headaches.

powervs/modules/hana_node/salt_provisioner.tf
mfriesenegger: This is a mistake and should be in PR #736.  I will remove from this PR and add to #736.

powervs/modules/hana_node/outputs.tf
mfriesenegger: This change worked.

powervs/modules/hana_node/salt_provisioner.tf
arbulu89: I didn't actually mean to run it in `remote-exec`. 
I wrote a better comment before, it looks like it got lost...

I was referring to the possibility to run these operation within the salt provisioning, as other salt states.
My unique concern was if the network reset could affect somehow to the terraform connection and if you had look into it.
It looks pretty ugly to have this code here if we could have it in salt as everything else

mfriesenegger: Now I understand your comment and good question.  The bastion needs to be working properly before provision.sh is called.  Bootstapping salt requires a properly functioning network connection through the bastion.  This is why I added this to the terraform code and not in salt.

arbulu89: Why is that? I thought this was only to improve the performance of the connections and not a mandatory thing.
I just want to avoid at any cost to have shell code outside of salt. We should have it there only if it is 100% required.
I didn't have time to look for other solutions, and I think we should invest some time there.

We already do network operations in salt (we even reboot the machines)

mfriesenegger: This PR fixes the download speed when going through a bastion and I believe is mandatory.  The networking does work but I have seen where the salt bootstrapping of a HANA node will fail due to network slowness which causes zypper timeouts.

Issue #740, opened 2021-08-11
Monitoring VM has no route for the Netweaver PAS and AAS monitoring exporter IP (sap_host_exporter)


diegoakechi:
**Used cloud platform**
Azure

**Used SLES4SAP version**
SLES for SAP 15.2

**Used client machine OS**
openSUSE TW.

**Expected behaviour vs observed behaviour**

The PAS and AAS sap_host_exporter targets should be available on Prometheus, but currently, they appear configured by Down.

Currently, there is a secondary IP and vHostname `sapha1pas`  used by the sap_host_exporter monitoring. On `etc/hosts` we see:

```
10.74.1.30		vmnetweaver01
10.74.1.31		vmnetweaver02
10.74.1.32		vmnetweaver03
10.74.1.33		vmnetweaver04
10.74.1.34		sapha1as
10.74.1.35		sapha1er
10.74.1.36		sapha1pas
10.74.1.37		sapha1aas1
```

But there is no Static IP associated with the NIC used by the machine. Adding it manually solves the problem (IP 10.74.1.36 was added manually):

![image](https://user-images.githubusercontent.com/8420718/129006929-92c1959d-0a88-4581-b8b0-43d9caa8e98a.png)

**How to reproduce**

The default deployment of HANA + Netweaver + Monitoring using the Develop branch produces this end-result. 

**Logs**
This was spot on the Trento Demo deployment. The logs can be collected there.

yeoldegrove:
> This is apparently a regression of #710

my bad ... I will look into it.
#710 was merged to fix #611

diegoakechi:
@yeoldegrove No worries ;)

yeoldegrove:
#722 introduces some logic to add VIPs to PAS/AAS instance NIC and also to ASCS (in non-HA use case). The logic can get quite complex.

A decision has to be made:
  - 1. use logic from #722 to add VIPs to NIC
    - only in azure, as other cloud providers work
  - 2. do not use VIPs at all for PAS/AAS
    - remove code from all cloud providers
    - same logic as in i. has to be implemented to pass "real IP" as VIP to pillar
  - 3. use "real IP" to monitor NW
    - might not be that easy, as virtual host name is set in `/etc/sap_host_exporter/${SID}_PAS${INSTANCE}.yaml` and this comes directly from the pillar
    - logic in pillar/tf code has to be changed which might be as complex as i.

There might be other solutions that I cannot think of at the moment.

[...]

PR #746, opened 2021-09-08
do not use virtual IPs on netweaver PAS/AAS and use node IP instead

yeoldegrove:
This is a proposed fix for #740.
It removes the virtual IPs for the PAS/AAS nodes for all providers. PAS/AAS will be rachable via the node IPs instead.


arbulu89 (changes requested):
Hi @yeoldegrove ,
I don't think this is correct. We need virtual IP address for Netweaver installation, this is what SAP recommeneds.
I just thought that we were talking on how to use the exporter itself.
I would just do the changes for the exported and not the whole installation

yeoldegrove:
Alright, seemed also a little odd to me, removing the vIPs.
I will test the a different solution where only the exporter is touched.

PR #750, opened 2021-09-16
make Netweaver PAS/AAS virtual IPs reachable on public cloud setups

yeoldegrove:
The Netweaver virtual IPs for AS/ERS are reachable via different mechanism for public clouds:
- azure -> load-balancer
- gcp -> route to VM (changed by cluster)
- aws -> route to VM (changed by cluster)

For the Netweaver virtual IPs for PAS/AAS there is currently no mechanism implemented to make them reachable to other nodes.
Therefore the current monitoring setup #740 is also broken.

This PR makes these PAS/AAS virtual IPs reachable via these mechanisms:
- azure -> vIP on network interface on VMs
  - load-balancer setup would also be possible in theory but is way more complex (e.g. create rules per node)
- gcp -> route to VM (static)
- aws -> route to VM (static)

In addition to that there are fixes included to not only deploy AS routes in HA scenarios.
```
count = local.create_ha_infra
vs.
count = local.vm_count > 0 ? 1 : 0
```

### DIFF
--- a/aws/modules/netweaver_node/main.tf
+++ b/aws/modules/netweaver_node/main.tf
@@ -37,7 +37,7 @@ resource "aws_route" "nw-ers-route" {
   instance_id            = aws_instance.netweaver.1.id
 }
 
-# deploy if only PAS
+# deploy if PAS on same machine as ASCS
 resource "aws_route" "nw-pas-route" {
   count                  = var.app_server_count == 0 ? 1 : 0
   route_table_id         = var.route_table_id
@@ -45,7 +45,7 @@ resource "aws_route" "nw-pas-route" {
   instance_id            = aws_instance.netweaver.0.id
 }
 
-# deploy if PAS and AAS
+# deploy if PAS and AAS on seperate hosts
 resource "aws_route" "nw-app-route" {
   count                  = var.app_server_count
   route_table_id         = var.route_table_id
--- a/aws/modules/netweaver_node/main.tf
+++ b/aws/modules/netweaver_node/main.tf
@@ -1,6 +1,7 @@
 locals {
   vm_count        = var.xscs_server_count + var.app_server_count
   create_ha_infra = local.vm_count > 1 && var.common_variables["netweaver"]["ha_enabled"] ? 1 : 0
+  app_start_index = local.create_ha_infra == 1 ? 2 : 1
 }
 
 # Network resources: subnets, routes, etc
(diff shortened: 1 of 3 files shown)


aws/modules/netweaver_node/main.tf
```diff
@@ -22,20 +23,36 @@ resource "aws_route_table_association" "netweaver-subnet-route-association" {
...
-resource "aws_route" "ers-cluster-vip" {
+resource "aws_route" "nw-ers-route" {
   count                  = local.create_ha_infra
   route_table_id         = var.route_table_id
   destination_cidr_block = "${element(var.virtual_host_ips, 1)}/32"
   instance_id            = aws_instance.netweaver.1.id
 }
 
+# deploy if only PAS
+resource "aws_route" "nw-pas-route" {
```
arbulu89, line 41: I would put a comment to say that this scenario is when the PAS is installed in the same machine than the ASCS

azure/modules/netweaver_node/main.tf
```diff
@@ -210,12 +211,48 @@ resource "azurerm_network_interface" "netweaver" {
...
   ip_configuration {
     name                          = "ipconf-primary"
+    primary                       = true
     subnet_id                     = var.network_subnet_id
     private_ip_address_allocation = "static"
     private_ip_address            = element(var.host_ips, count.index)
     public_ip_address_id          = local.bastion_enabled ? null : element(azurerm_public_ip.netweaver.*.id, count.index)
   }
 
+  # deploy if non-HA AS
```
arbulu89, line 221: Is this the ASCS?

gcp/modules/netweaver_node/main.tf
```diff
@@ -37,6 +38,28 @@ resource "google_compute_route" "nw-ers-route" {
...
+  dest_range             = "${element(var.virtual_host_ips, local.app_start_index)}/32"
+  network                = var.network_name
+  next_hop_instance      = google_compute_instance.netweaver.0.name
+  next_hop_instance_zone = element(var.compute_zones, 0)
+  priority               = 1000
+}
+
+# deploy if PAS and AAS
+resource "google_compute_route" "nw-app-route" {
+  name                   = "${var.common_variables["deployment_name"]}-nw-appoute-${format("%02d", local.app_start_index + count.index + 1)}"
```
arbulu89, line 54: Typo here? `appoute`

arbulu89 (approved):
@eik
The changes look correct so far, but I have added some small comments.

aws/modules/netweaver_node/main.tf
yeoldegrove: I made the comments more clear.

gcp/modules/netweaver_node/main.tf
yeoldegrove: Yes, named it that way now.

azure/modules/netweaver_node/main.tf
yeoldegrove: Yes, named it that way now.

PR #751, opened 2021-09-21
add minion.d files for each environment (base/predeployment)

yeoldegrove:
This fixes an issue which comes with the updated rpm `salt-standalone-formulas-configuration-3002.2-46.1` which introduces a new file `/etc/salt/minion.d/standalone-formulas-configuration.conf` that overwrites our `base/predeployment` environments in `/etc/salt/minion`.

This should be a permanent fix as long as there is no tampering with the environments by additional rpms.

All other minion settings are now in separate config files, too.

### DIFF
--- a/salt/os_setup/minion_configuration.sls
+++ b/salt/os_setup/minion_configuration.sls
@@ -14,6 +14,12 @@
             - /srv/salt
             - /usr/share/salt-formulas/states
 
+# prevent "[WARNING ] top_file_merging_strategy is set to 'merge' and multiple top files were found."
+/etc/salt/minion.d/top_file_merging_strategy.conf:
+  file.managed:
+    - contents: |
+        top_file_merging_strategy: same
+
 backup_salt_configuration:
   file.copy:
     - name: /etc/salt/minion.backup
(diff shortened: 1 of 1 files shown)


salt/os_setup/minion_configuration.sls
```diff
@@ -1,27 +1,30 @@
-backup_salt_configuration:
-  file.copy:
-    - name: /etc/salt/minion.backup
-    - source: /etc/salt/minion
-
-configure_file_roots:
-  file.append:
-    - name: /etc/salt/minion
-    - text: |
+/etc/salt/minion.d/environment_base.conf:
```
arbulu89, line 1: Do we really need this 1st entry?
As far as I understood, this configuration is already coming in the salt configuration package

salt/os_setup/minion_configuration.sls
```diff
@@ -1,27 +1,30 @@
...
-    - name: /etc/salt/minion
-    - text:
-      - 'use_superseded:'
-      - '- module.run'
+# prevent "[WARNING ] top_file_merging_strategy is set to 'merge' and multiple top files were found."
+/etc/salt/minion.d/top_file_merging_strategy.conf:
+  file.managed:
+    - contents: |
+        top_file_merging_strategy: same
+
```
arbulu89, line 22: I would still put a comment on top of this like `Override the usage of module.run to the new version.`

arbulu89 (approved):

PR #754, opened 2021-09-21
ci: ignore www.linux-ha.org links as site is down since days

yeoldegrove:
This fixes the failing "Ci tests / markup link checker" tests.


arbulu89 (changes requested):
@yeoldegrove I'm not sure if this is the best solution.
The url by now is not working, and I don't know if this is something temporal.
I would vote to remove the url from `doc/fencing.md`

yeoldegrove:
@arbulu89 You are right. The documentation in the clusterlabs wiki is good enough as a starting point.

arbulu89 (approved):
Thanks!

PR #756, opened 2021-09-23
master: Fix 'pkg_requirements' empty variable

ldevulder:
Variable `pkg_requirements` can be uninitialized if `requirements.yml` file is delete. As this file is optional, the variable needs to be initialized with an empty value.


yeoldegrove (approved):
If the use case is "no package requirements", than this fix is fine.

PR #759, opened 2021-09-25
powervs: update README.md files

mfriesenegger:


powervs/README.md
```diff
@@ -0,0 +1,211 @@
...
+Error while installing hashicorp/ibm: provider registry registry.terraform.io
+does not have a provider named registry.terraform.io/hashicorp/ibm
+```
+
+Use the following commands as a workaround then rerun terraform init.
+
+```
+mkdir .terraform/plugins/registry.terraform.io/hashicorp/ibm
+cp -a .terraform/plugins/registry.terraform.io/ibm-cloud/ibm/* .terraform/plugins/registry.terraform.io/hashicorp/ibm
+```
```
stefanotorresi, line 154: hmm this is very strange, as it shouldn't happen at all if the provider is correctly configured.
The correct provider name is `IBM-Cloud/ibm` now. As long as that is used, then `ŧerraform init` shouldn't have any problems in fetching the provider remotely from the hashicorp registry.
I think in general we should fix these kind of problems at the root, rather than providing a workaround in the official documentation itself!

mfriesenegger: Thank you @stefanotorresi for the comment.  I was able to figure out what I need to change so the provider error does not happen.  I will update this PR that removes the lines you have highlighted and create an additional PR that resolves the terraform error.

mfriesenegger: @stefanotorresi The PR has been updated and the lines you highlighted have been removed.  PR #764 contains the updates to terraform that remove the `terraform init` error.

stefanotorresi: Thank you Mike!!

yeoldegrove (approved):
Looks good to me.

PR #762, opened 2021-09-28
Change azure default iscsi vm size

stefanotorresi:
SBD on ISCSI shouldn't need that big of a vm.


arbulu89 (approved):
Could we have a follow up ticket to expand this to the other cloud providers?

yeoldegrove:
The follow up ticket should include all VM types.
The default size should be cost effective but supported/certified for HANA/Netweaver (where needed).

PR #765, opened 2021-09-30
openstack: use existing volume on bastion NFS server and smaller fixes

yeoldegrove:
- openstack: possibility to use existing volume on bastion NFS server
be45329
- openstack: fix cloud-init and provide template file
54d5891
- openstack: smaller fixes
29e4201
- openstack: fix iscsi provisioning

### DIFF
--- a/openstack/main.tf
+++ b/openstack/main.tf
@@ -63,33 +63,10 @@ resource "openstack_compute_keypair_v2" "key_terraform" {
 }
 
 data "template_file" "userdata" {
-  template = <<CLOUDCONFIG
-#cloud-config
-
-cloud_config_modules:
-  - resolv_conf
-
-manage_resolv_conf: true
-
-resolv_conf:
-  nameservers: ['8.8.4.4', '8.8.8.8']
-
-users:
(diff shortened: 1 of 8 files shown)


openstack/main.tf
```diff
@@ -63,33 +63,10 @@ resource "openstack_compute_keypair_v2" "key_terraform" {
...
-
-runcmd:
-- |
-  # add any command here
-  # echo "any command"
-
-CLOUDCONFIG
+  template = file("${path.root}/cloud-config.tpl")
+  vars = {
+    public_key = file(var.public_key)
```
arbulu89, line 68: I don't see the `public_key` variable anywhere in the new template. How does this work?

arbulu89 (approved):
Thanks @yeoldegrove

openstack/main.tf
yeoldegrove: You are correct. This is a leftover from trying this out. Will remove it.
The actual ssh public_key is passed via `key_pair` to the `openstack_compute_instance_v2`.

Issue #757, opened 2021-09-24
Make the HANA and NW machines hostname configurable


arbulu89:
In order to make multiple different deployents, that are not exactly a replica, it would be good to have the option to customize the name if the machines, specially the HANA and NW machine names.

So, for example in the Azure case, instead of having `vmhana01` always, it would be good to have the option to configure the name to `myname01`, etc, keeping the current value as default. The suffix should still be used to number the multiple machines.

yeoldegrove:
The new feature is added with #767.

PR #767, opened 2021-09-30
make hostnames configurable

yeoldegrove:
This PR makes it possible to configure the hostnames of each VM type via variables in `terraform.tfvars`. It addresses #757 .
Additionally there is the possibility to add the `deployment_name` as a prefix to the actual hostname (in salt and OS).
The `deployment_name` will be added to terraform resources as usual (no change in behavior).
``` 
  # Add the "deployment_name" as a prefix to the hostname.
  #deployment_name_in_hostname = false

  #hana_name = "vmhana"
  #iscsi_name = "vmiscsi"
  #monitoring_name = "vmmonitoring"
  #drbd_name = "vmdrbd"
  #netweaver_name = "vmnetweaver"
``` 

- run salt default state on all hosts
  * needed to actually change hostnames via state `default.hostname`

- ignore terraform lockfiles

- ignore gcp credentials

- azure: make hostnames configurable

- aws: make hostnames configurable

- gcp: make hostnames configurable

- libvirt: make hostnames configurable

### DIFF
--- a/aws/modules/hana_node/variables.tf
+++ b/aws/modules/hana_node/variables.tf
@@ -2,6 +2,16 @@ variable "common_variables" {
   description = "Output of the common_variables module"
 }
 
+variable "name" {
+  description = "hostname, without the domain part"
+  type        = string
+}
+
+variable "network_domain" {
+  description = "hostname's network domain"
+  type        = string
+}
+
 variable "hana_count" {
   type    = string
   default = "2"
@@ -10,12 +20,6 @@ variable "hana_count" {
 variable "instance_type" {
   type    = string
   default = "r3.8xlarge"
-}
-
-variable "name" {
-  description = "prefix of the machines names"
-  type        = string
-  default     = "hana"
 }
 
 variable "aws_region" {
--- a/aws/variables.tf
+++ b/aws/variables.tf
@@ -89,6 +89,18 @@ variable "deployment_name" {
   default     = ""
 }
 
+variable "deployment_name_in_hostname" {
+  description = "Add deployment_name as a prefix to all hostnames."
+  type        = bool
+  default     = true
+}
+
+variable "network_domain" {
+  description = "hostname's network domain for all hosts. Can be overwritten by modules."
+  type        = string
+  default     = "tf.local"
+}
+
 variable "os_image" {
   description = "Default OS image for all the machines. This value is not used if the specific nodes os_image is set (e.g. hana_os_image)"
   type        = string
@@ -182,9 +194,16 @@ variable "provisioning_output_colored" {
 
 # Hana related variables
 
-variable "name" {
+variable "hana_name" {
   description = "hostname, without the domain part"
   type        = string
+  default     = "vmhana"
+}
+
+variable "hana_network_domain" {
+  description = "hostname's network domain"
+  type        = string
+  default     = ""
 }
 
 variable "hana_count" {
@@ -408,6 +427,18 @@ variable "scenario_type" {
 
 # DRBD related variables
 
+variable "drbd_name" {
+  description = "hostname, without the domain part"
+  type        = string
+  default     = "vmdrbd"
+}
+
+variable "drbd_network_domain" {
+  description = "hostname's network domain"
+  type        = string
+  default     = ""
+}
+
 variable "drbd_enabled" {
   description = "Enable the DRBD cluster for nfs"
   type        = bool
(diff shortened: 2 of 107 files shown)


aws/variables.tf
```diff
@@ -182,7 +188,7 @@ variable "provisioning_output_colored" {
 
 # Hana related variables
 
-variable "name" {
+variable "hana_name" {
   description = "hostname, without the domain part"
   type        = string
```
arbulu89, line 193: Shouldn't we have a default value here? `vmhana`

aws/modules/hana_node/variables.tf
```diff
@@ -2,6 +2,17 @@ variable "common_variables" {
   description = "Output of the common_variables module"
 }
 
+variable "name" {
+  description = "hostname, without the domain part"
+  type        = string
+}
+
+variable "network_domain" {
```
arbulu89, line 10: I'm sure now, but is the `network_domain` used in the cloud providers? Can you check if this variable executes something on salt in these scenarios?
Maybe we can even remove it

azure/modules/drbd_node/variables.tf
```diff
@@ -52,12 +46,17 @@ variable "os_image" {
   type        = string
 }
 
+variable "name" {
+  type    = string
```
arbulu89, line 50: Let's add the description

azure/modules/hana_node/variables.tf
```diff
@@ -74,11 +69,21 @@ variable "os_image" {
   type        = string
 }
 
+variable "name" {
+  type    = string
```
arbulu89, line 73: Description

arbulu89 (changes requested):
Hey @yeoldegrove,

One important consideration.
I have seen that you have remove the workspace part from gcp, aws, libvirt resources name.
I'm not sure if we can do that.
The thing if we do that, we most probably are not going to be able to have multiple deployments in one subscription/user, and as the name would conflict.

That's why we added the workspace, to have the option to easily prefix the resources on these places. In azure, we have the advantage of the resource groups, what doesn't exist in the others.

I'm afraid you will need to revert this back.

With this, what we could do is to make the `deployment_name` a mandatory value, and remove any default option. We should always know who is the owner of the deployment

aws/modules/hana_node/variables.tf
yeoldegrove: As far as I can tell `network_domain` is only used here to set the FQDN hostname: https://github.com/SUSE/ha-sap-terraform-deployments/blob/619c50f66436282e479823ddb8526f52700cf3dd/salt/default/hostname.sls#L31

This might be an important option in some customer cases. Even different domains per VM type might be a use case. This is why I built it to be set on a per module level.
I forgot to extend it so that one can set it from `terraform.tfvars` on a global and VM type level.
@arbulu89  Are you ok with this?

arbulu89: If it is useful for the users and we still have the default value, it is totally fine for me

aws/variables.tf
yeoldegrove: Will be fixed.

azure/modules/drbd_node/variables.tf
yeoldegrove: Will be added everywhere.

azure/modules/hana_node/variables.tf
yeoldegrove: Will be added everywhere.

yeoldegrove:
@arbulu89 I get the point about the workspace and deployment_name prefixing in the resources...
The intention was not to change this behavior, will rework it.

I plan to built it in a way that the resource names stay the same as now and you can only decide if you want to have the deployment_name as a prefix to the actual hostname (within salt/OS, not terraform).

arbulu89:
> @arbulu89 I get the point about the workspace and deployment_name prefixing in the resources... The intention was not to change this behavior, will rework it.
> 
> I plan to built it in a way that the resource names stay the same as now and you can only decide if you want to have the deployment_name as a prefix to the actual hostname (within salt/OS, not terraform).

I don't know if there is a way to use resources with the same name. Maybe using tags or similar.

yeoldegrove:
@arbulu89 please re-review.

arbulu89 (changes requested):
@yeoldegrove Just a small request.
I would put the `deployment_name_in_hostname` variable default value to true in `aws`, `gcp` and `libvirt`, and leave it to false in `azure`.
Not having the deployment name should be the edge case.

aws/variables.tf
```diff
@@ -89,6 +89,18 @@ variable "deployment_name" {
   default     = ""
 }
 
+variable "deployment_name_in_hostname" {
+  description = "Add deployment_name as a prefix to all hostnames."
+  type        = bool
+  default     = false
```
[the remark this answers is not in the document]
yeoldegrove: @arbulu89 fixed

PR #784, opened 2021-10-27
Fix wait_until_ssh_is_ready_xxx failure in libvirt and correct requirements for hana_node

nick-wang:
Bug fixes for hana_node/drbd_node:

1. Fix wait_until_ssh_is_ready_xxx failure after timeout when no cluster_ssh_pub/cluster_ssh_key defined
Salt state 'wait_until_ssh_is_ready_xxx' in 'wait.sls' is introduced in scale out feature.
It depends on the ssh keys (cluster_ssh_pub/cluster_ssh_key) of clusters and only deployed via 'cluster_node/ha/ssh.sls' (cluster_node/ha/init.sls)

Without the two keys defined in tfvars file, 'wait.sls' will wait until
failure. And wait.sls is not a must in non scale out hana scenario.
Only run the state when ssh key available.

2. drbd_node: allow create parent directory of notify-split-brain-haclusterexporter-suse-metric.sh

3. fix the requirements.yml for saphanabootstrap-formula since the template name is changed.

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -4,7 +4,7 @@ hana_node:
   python3-shaptools: 0.3.12+git.1619007514.1951d23
   salt-shaptools: 0.3.11+git.1605797958.ae2f08a
   habootstrap-formula: 0.4.1+git.1611775401.451718e
-  saphanabootstrap-formula: 0.7.1+git.1619008686.8600866
+  saphanabootstrap-formula: 0.8.0+git.1634566980.9dcfa9f
 drbd_node:
   python-shaptools: 0.3.12+git.1619007514.1951d23
   python3-shaptools: 0.3.12+git.1619007514.1951d23
--- a/salt/hana_node/init.sls
+++ b/salt/hana_node/init.sls
@@ -6,4 +6,6 @@ include:
   {% endif %}
   - hana_node.mount
   - hana_node.hana_packages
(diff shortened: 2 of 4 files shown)


requirements.yml
```diff
@@ -4,7 +4,7 @@ hana_node:
   python3-shaptools: 0.3.12+git.1619007514.1951d23
   salt-shaptools: 0.3.11+git.1605797958.ae2f08a
   habootstrap-formula: 0.4.1+git.1611775401.451718e
-  saphanabootstrap-formula: 0.7.1+git.1619008686.8600866
+  saphanabootstrap-formula: 0.8.0+git.1634566980.9dcfa9f
```
yeoldegrove, line 7: We usually bump this only in the `master` branch, but it is fine I guess.

salt/hana_node/init.sls
```diff
@@ -6,4 +6,6 @@ include:
   {% endif %}
   - hana_node.mount
   - hana_node.hana_packages
+  {% if grains['cluster_ssh_pub'] is defined and grains['cluster_ssh_key'] is defined %}
```
yeoldegrove, line 9: I was not aware that this is not defined for `libvirt` when implementing this.
Good finding :+1:

yeoldegrove (approved):

PR #789, opened 2021-11-11
hana scale-out: fix ssh test and make it more resilient

yeoldegrove:
- ssh commands will not hang anymore because of password promt
- make sure `saphanabootstrap-formula` is installed in majority maker


melzer-b1 (commented):
Looks good for me. Thanks Eike

melzer-b1 (approved):

PR #790, opened 2021-11-12
hana scale-out: reuse hana code for majority maker

yeoldegrove:
This PR removes most of the majority maker related salt code.

The majority maker will be built up by reusing the code from `saphanabootstrap-formula` and `salt/hana_node`.
To enable this, the majority maker has to be a node in the hana pillars to access (or the `saphanabootstrap-formula` code will not run).


melzer-b1 (approved):
Reviewed. Thanks Eike

PR #792, opened 2021-11-17
Disable nw-pas-route creation if nw is not enabled

arbulu89:
The deployment in AWS is failing if Netweaver is not enabled:

```
Error: Invalid index

  on modules/netweaver_node/main.tf line 46, in resource "aws_route" "nw-pas-route":
  46:   instance_id            = aws_instance.netweaver.0.id
    |----------------
    | aws_instance.netweaver is empty tuple

The given key does not identify an element in this collection value.
```

This fix basically checks if there is any NW vm before checking the current condition and create the route

### DIFF
--- a/aws/modules/netweaver_node/main.tf
+++ b/aws/modules/netweaver_node/main.tf
@@ -40,7 +40,7 @@ resource "aws_route" "nw-ers-route" {
 
 # deploy if PAS on same machine as ASCS
 resource "aws_route" "nw-pas-route" {
-  count                  = var.app_server_count == 0 ? 1 : 0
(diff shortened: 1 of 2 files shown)


yeoldegrove (changes requested):
Thanks for finding/fixing this.
Could you please also fix this for GCP? It has the exact same logic in it ;)

arbulu89:
> Thanks for finding/fixing this. Could you please also fix this for GCP? It has the exact same logic in it ;)

Good catch. Thanks!

arbulu89:
@yeoldegrove Requested change force pushed

yeoldegrove (approved):
Good to merge.

PR #794, opened 2021-11-18
openstack: fix netweaver virtual IPs on ports

yeoldegrove:
This will allow all virtual IPs to be bindable/reachable on on the netweaver cluster.


melzer-b1 (approved):
Thanks Eike

PR #795, opened 2021-11-18
add minimal NFS server to be used on libvirt/openstack (instead of DRBD) - not for production workloads

yeoldegrove:
This is a prerequisite to deploy HANA scale-out on libvirt/openstack. It should not be used for production workloads.


melzer-b1 (approved):
Reviewed. Thanks Eike

PR #797, opened 2021-11-22
set default VIP mechanisms to support latest DRBD, HANA, Netweaver formulas

yeoldegrove:
These formula changes make it mandatory to set a valid `*_vip_mechanism` for each cluster setup.

* https://github.com/SUSE/saphanabootstrap-formula/pull/137
* https://github.com/SUSE/sapnwbootstrap-formula/pull/91
* https://github.com/SUSE/drbd-formula/pull/28

Currently we mostly set `""` which will result in the grains beeing `null` which is not interpretable by the formulas.


melzer-b1 (approved):
Looks OK. Thanks Eike.

PR #799, opened 2021-12-02
master ci: enable salt-lint again (was missing)

yeoldegrove:
Somehow `salt-lint` is not executed in the CI.


melzer-b1 (approved):
Looks OK, thanks Eike

PR #807, opened 2022-01-11
remove loki user from promtail setup

yeoldegrove:
The rpm `loki-2.4.1+git.1636374074.f61a4d261` updated in https://build.opensuse.org/request/show/940653 removed the `loki` user and now runs as root.

We need to adapt this in our promtail setup.


arbulu89 (commented):
Hey @yeoldegrove ,
Could we comment somewhere here that this removal is due to an incorrect change in the `loki` package?
Maybe, instead of removing these states, just comment them and put some link to the failed obs package

yeoldegrove:
@arbulu89 There is now a bit of an explanation in the code.

arbulu89 (approved):
Thanks!

PR #808, opened 2022-01-19
add S/4HANA 2020 and 2021 as supported versions

yeoldegrove:
Latest `saphana-bootstrap-formula` is also needed with code from https://github.com/SUSE/sapnwbootstrap-formula/pull/92.


arbulu89 (approved):
Nice!

PR #868, opened 2022-06-21
Enchance os registration

mr-stringer:
A small change to better detect systems suitable for registration with SUSEConnect.


yeoldegrove (approved):
LGTM

Issue #910, opened 2023-10-23
NEW Feature: Implement susTkOver, susChkSrv, & susCostOpt srHooks


tuttipazzo:
**Used cloud platform**
AWS, GCP, Azure, libvirt

**Used SLES4SAP version**
SLES15.SPx

**Used client machine OS**
Leap/TW

**Expected behaviour vs observed behaviour**
Nothing is failing.  These new hooks should be included.

**How to reproduce**
N/A

**Logs**
N/A

yeoldegrove:
The hook should be integrated in https://github.com/SUSE/saphanabootstrap-formula.
Maybe one could think about an interface/switch to enable/disable certain hooks.

tuttipazzo:
I see susTkOver from 2-node cluster I deployed back in late September.  I had to manually add susChksrv so that's missing.  But I agree with yeoldegrove that there needs ot be a switch to turn on/off some of the hooks.

yeoldegrove:
@tuttipazzo https://github.com/SUSE/ha-sap-terraform-deployments/releases/tag/9.2.0 lets you define the following in `terraform.tfvars`:
```
# HANA HA/DR provider configuration
# See https://documentation.suse.com/sbp/all/single-html/SLES4SAP-hana-sr-guide-PerfOpt-15/#cha.s4s.hana-hook for details.
# The SAPHanaSR hook is always enabled.
# enable susTkOver hook (disabled by default)
# hana_ha_dr_sustkover_enabled = true
# enable susChkSrv hook (disabled by default)
# hana_ha_dr_suschksrv_enabled = true
# susChkSrv action on lost, see `man 7 susChkSrv.py` (Options: stop [default], fence)
# hana_ha_dr_suschksrv_action_on_lost = "fence"
```

This solves your use case.

PR #911, opened 2023-11-07
add susChkSrv HA/DR provider and make providers configurable

yeoldegrove:
* add susChkSrv HA/DR provider
* make providers configurable

This will be a part of fixing https://github.com/SUSE/ha-sap-terraform-deployments/issues/910

Needs https://github.com/SUSE/saphanabootstrap-formula/pull/153 to be merged.

### DIFF
--- a/aws/variables.tf
+++ b/aws/variables.tf
@@ -590,6 +590,24 @@ variable "hana_efs_performance_mode" {
   default     = "generalPurpose"
 }
 
+variable "hana_ha_dr_sustkover_enabled" {
+  description = "enable susTkOver hook"
+  type        = bool
+  default     = false
+}
+
+variable "hana_ha_dr_suschksrv_enabled" {
+  description = "enable susChkSrv hook"
+  type        = bool
+  default     = false
+}
+
+variable "hana_ha_dr_suschksrv_action_on_lost" {
+  description = "define action on lost for susChkSrv, see `man 7 susChkSrv.py`"
+  type        = string
+  default     = "stop"
+}
+
 # DRBD related variables
 
 variable "drbd_name" {
--- a/generic_modules/common_variables/hana_variables.tf
+++ b/generic_modules/common_variables/hana_variables.tf
@@ -217,3 +217,24 @@ variable "hana_basepath_shared" {
   type        = bool
   default     = true
 }
+
+variable "hana_ha_dr_sustkover_enabled" {
+  description = "enable susTkOver hook"
+  type        = bool
+}
+
+variable "hana_ha_dr_suschksrv_enabled" {
+  description = "enable susChkSrv hook"
+  type        = bool
+}
+
+variable "hana_ha_dr_suschksrv_action_on_lost" {
+  description = "define action on lost for susChkSrv, see `man 7 susChkSrv.py`"
+  type        = string
+  validation {
+    condition = (
+      can(regex("^(stop|fence)$", var.hana_ha_dr_suschksrv_action_on_lost))
+    )
+    error_message = "Invalid action on lost for susChkrSrv. Options: stop, fence."
+  }
+}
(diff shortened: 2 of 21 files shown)


yeoldegrove:
`requirements.yml` will be bumped as soon as new `saphanabootstrap-formula` rpm is build.

generic_modules/common_variables/hana_variables.tf
```diff
@@ -217,3 +217,24 @@ variable "hana_basepath_shared" {
...
+  type        = bool
+}
+
+variable "hana_ha_dr_suschksrv_action_on_lost" {
+  description = "define action on lost for susChkSrv, see `man 7 susChkSrv.py`"
+  type        = string
+  validation {
+    condition = (
+      can(regex("^(stop|fence)$", var.hana_ha_dr_suschksrv_action_on_lost))
+    )
```
stefanotorresi, line 237: nitpick: you could simplify this by using `contains` instead of the regex

aws/variables.tf
```diff
@@ -590,6 +590,30 @@ variable "hana_efs_performance_mode" {
...
+  description = "define action on lost for susChkSrv, see `man 7 susChkSrv.py`"
+  type        = string
+  validation {
+    condition = (
+      can(regex("^(stop|fence)$", var.hana_ha_dr_suschksrv_action_on_lost))
+    )
+    error_message = "Invalid action on lost for susChkrSrv. Options: stop, fence."
+  }
+  default = "stop"
+}
```
stefanotorresi, line 615: since we are feeding this variable into the generic HANA module, I guess repeating the validation is not really needed in each provider?

generic_modules/common_variables/hana_variables.tf
yeoldegrove: It is copypasta and looks like everywhere else ;)

aws/variables.tf
yeoldegrove: Of course...again copypasta... I could correct this everywhere else, too.
Let me have a look how much changes this would be.

yeoldegrove: Fixed it for `hana_ha_dr_suschksrv_action_on_lost` the other validations in the cloud specific `variables.tf` files are fine as the are related to IPs, storage, fencing which tends to differ between clouds.

yeoldegrove:
@stefanotorresi @angelabriel Is it fine to enable both hooks here by default?

aws/variables.tf
```diff
@@ -590,6 +590,24 @@ variable "hana_efs_performance_mode" {
   default     = "generalPurpose"
 }
 
+variable "hana_ha_dr_sustkover_enabled" {
+  description = "enable susTkOver hook"
+  type        = bool
+  default     = true
```
arbulu89, line 596: The default value should be `false`, right?
Or do we want to have it enabled by default.
Coming to the discussion in the formula, as it is not mandatory, we could set it disabled by default.
If you change this, remember to change the `tfvars` example file

pillar_examples/automatic/hana/hana.sls
```diff
@@ -62,6 +62,9 @@ hana:
   monitoring_enabled: {{ grains['monitoring_enabled']|default(False) }}
   ha_enabled: {{ grains['ha_enabled'] }}
   basepath_shared: {{ grains['hana_basepath_shared']|default(True) }}
+  ha_dr_sustkover_enabled: {{ grains['hana_ha_dr_sustkover_enabled']|default(True) }}
```
arbulu89, line 65: The same with the default value

arbulu89 (approved):
Hey @yeoldegrove ,
All good codewise. I'm wondering if we should put them disabled by default, as we did in the formula.
No hard feelings.

angelabriel:
> @stefanotorresi @angelabriel Is it fine to enable both hooks here by default?

I would prefer to have it consistent, which means disabled by default. For the not mandatory Hooks the customer should actively enable them to get them work. He should be aware of these Hooks and their behavior as they might interfere with other tools he is using for daily work.

yeoldegrove:
@arbulu89 @angelabriel Changed the defaults to disabled. Please approve.

arbulu89 (approved):

